PR #459, opened 2024-01-24
Pascal/mar 459 internal metabase dashboard

Pascal-Delange:
The design is the following:
- a list of sql definitions of views for analytics (metabase) are stored in `repositories/analytics_views`. 
- every time we run the migrations, we recreate all the views based on the current definitions (this is a fast operation because overwriting views is a metadata only operation)
- we do this so that we can expose a read only user _without access to fields that we don't want to share_ for metabase (or similar tools). For instance, we don't expose the decision's trigger payload or the rules' AST definition. Feedback welcome if you think some fields that are exposed as of this PR should actually stay hidden
- I took the occasion to refactor the migration function a little bit

Next steps:
- deploy staging
- connect our metabase app to our staging db
- deploy prod, create a read only replica (we'll need this anyway for fail over in case of an incident)
- connect our metabase app to our prod replica db
- start building dashboards on metabase / expose customer data by building filtered views (by organization_id) in metabase

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,7 @@
+{
+      "recommendations": [
+            "golang.go",
+            "inferrinizzard.prettier-sql-vscode"
+      ]
+    }
+    
--- a/repositories/analytics_views/apikeys.sql
+++ b/repositories/analytics_views/apikeys.sql
@@ -1,0 +1,13 @@
+DROP VIEW IF EXISTS analytics.apikeys;
+
+CREATE VIEW
+      analytics.apikeys
+WITH
+      (security_invoker = false) AS
+SELECT
+      id,
+      org_id AS organization_id,
+      deleted_at,
+      role
+FROM
+      marble.apikeys;
--- a/repositories/analytics_views/data_model_fields.sql
+++ b/repositories/analytics_views/data_model_fields.sql
@@ -1,0 +1,16 @@
+DROP VIEW IF EXISTS analytics.data_model_fields;
+
+CREATE VIEW
+      analytics.data_model_fields
+WITH
+      (security_invoker = false) AS
+SELECT
+      id,
+      table_id,
+      name,
+type,
+nullable,
+description,
+is_enum
+FROM
+      marble.data_model_fields
--- a/repositories/analytics_views/users.sql
+++ b/repositories/analytics_views/users.sql
@@ -1,0 +1,15 @@
+DROP VIEW IF EXISTS analytics.users;
+
+CREATE VIEW
+      analytics.users
+WITH
+      (security_invoker = false) AS
+SELECT
+      id,
+      organization_id,
+      email,
+      firebase_uid,
+      role,
+      deleted_at
+FROM
+      marble.users
--- a/repositories/migrations.go
+++ b/repositories/migrations.go
@@ -1,11 +1,11 @@
 package repositories
 
 import (
+	"context"
 	"database/sql"
 	"embed"
 	"fmt"
 	"io/fs"
-	"log/slog"
 
 	"github.com/checkmarble/marble-backend/utils"
 	"github.com/cockroachdb/errors"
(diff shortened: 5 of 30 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+      "recommendations": [
+            "golang.go",
+            "adpyke.vscode-sql-formatter"
```
Pascal-Delange, line 4: let's use this to have uniformly formatted sql queries in the analytics_views folder

repositories/analytics_views/apikeys.sql
```diff
@@ -0,0 +1,10 @@
+DROP VIEW IF EXISTS analytics.apikeys;
+
+CREATE VIEW analytics.apikeys WITH (security_invoker = false) AS
```
Pascal-Delange, line 3: I do "drop" then "create", because "CREATE OR REPLACE" does not allow for schema updates (e.g. you can't rename a column on an existing view doing so)

repositories/analytics_views/apikeys.sql
```diff
@@ -0,0 +1,10 @@
...
+SELECT
+      id,
+      org_id AS organization_id,
```
Pascal-Delange, line 6: Only level of normalization I did here was to use organization_id instead of org_id everywhere, for clarity

repositories/analytics_views/apikeys.sql
```diff
@@ -0,0 +1,10 @@
+DROP VIEW IF EXISTS analytics.apikeys;
+
+CREATE VIEW analytics.apikeys WITH (security_invoker = false) AS
+SELECT
```
Pascal-Delange, line 4: note that private fields that should never leave our prod environment, such as in this case the api key value, are not exposed in the view

repositories/migrations/20240124111516_create_analytics_schema.sql
```diff
@@ -0,0 +1,35 @@
+-- +goose Up
+-- +goose StatementBegin
+-- create and make default the marble schema
+CREATE SCHEMA IF NOT EXISTS analytics;
+
+GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA analytics TO postgres;
+
+DO $ $ BEGIN CREATE USER analytics WITH PASSWORD 'default';
```
Pascal-Delange, line 8: This routine will create the user analytics if it doesn't already exists. The expectation is that, in our cloud environment at least, we will have created the analytics user in cloudsql, and stored the password in secret manager/1password. But this should still work in local dev mode or for open source clients.

.vscode/extensions.json
Pascal-Delange: Ok the problem with this, we need to use pl/sql format or else it breaks some queries (at least it breaks the double $$ sign into two in my latest migration)

Pascal-Delange: P.S. in the end I propose to use `inferrinizzard.prettier-sql-vscode` with the postgresql formatted, that handles $$ ok

repositories/migrations.go
```diff
@@ -17,63 +19,101 @@ import (
...
+				return m.createViewFromFile(ctx, folder, path)
+			}
+
+			return nil
+		},
+	); err != nil {
+		return errors.Wrap(err, "error while walking embedded analytics_views folder")
+	}
+	return nil
+}
```
balzdur, line 104: What about cleaning ? IIUC, we need to let at least a file with `DROP VIEW IF EXISTS ...` to ensure suppression of deprecated views is that right ?

repositories/migrations/20240124111516_create_analytics_schema.sql
```diff
@@ -0,0 +1,37 @@
+-- +goose Up
+-- +goose StatementBegin
+-- create and make default the marble schema
```
balzdur, line 3:
```suggestion
-- create the analytics schema and an analytics user
```

repositories/migrations/20240124111516_create_analytics_schema.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE SCHEMA IF NOT EXISTS analytics;
+
+GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA analytics TO postgres;
+
+DO $$ BEGIN
+      CREATE USER analytics WITH PASSWORD 'default';
```
balzdur, line 9: Is it ok to let an hardcoded password like this ? The access to the db is handled in another way ?

repositories/migrations/20240124111516_create_analytics_schema.sql
```diff
@@ -0,0 +1,37 @@
...
+
+ALTER ROLE analytics
+SET
+      search_path TO analytics;
+
+-- +goose StatementEnd
+-- +goose Down
+-- +goose StatementBegin
+DROP SCHEMA IF EXISTS analytics CASCADE;
+
```
balzdur, line 36: You may need to remove the user too

repositories/analytics_views/users.sql
```diff
@@ -0,0 +1,15 @@
+DROP VIEW IF EXISTS analytics.users;
+
+CREATE VIEW
+      analytics.users
+WITH
+      (security_invoker = false) AS
```
balzdur, line 6: For my curiosity, why do we need this ?

repositories/analytics_views/users.sql
```diff
@@ -0,0 +1,15 @@
...
+SELECT
+      id,
+      organization_id,
+      email,
+      firebase_uid,
+      role,
+      deleted_at
```
balzdur, line 13: Do we want to expose all user data ?

repositories/analytics_views/data_model_fields.sql
```diff
@@ -0,0 +1,16 @@
...
+WITH
+      (security_invoker = false) AS
+SELECT
+      id,
+      table_id,
+      name,
+type,
+nullable,
+description,
+is_enum
```
balzdur, line 14: format issue

balzdur (approved):
Ok for me, still I think I miss some context on all the things we want to expose to Metabase (like FirebaseUID for example)

repositories/migrations/20240124111516_create_analytics_schema.sql
Pascal-Delange: So the idea is that the default solution would be to create an analytics user before the migrations run, and expose an env variable for this.
I guess the best solution here is to create a user without password, which IIUC is possible (it just won't be possible to login as it)

repositories/migrations/20240124111516_create_analytics_schema.sql
Pascal-Delange: no that's kinda tricky, because you don't know what the state before running the "up" migration is. So in this case it's hard to make it so that up=>down = identity

repositories/analytics_views/users.sql
Pascal-Delange: So there are two ways of running a view:
- with the permissions of the user running the query
- with the permissions of the user that created the query

think of it like this:
- I have two tables A and B, and a user analytics that has permissions to read table A only.
- You define a view that reads from tables A and B
- if the view is defined with security_invoker=true, you won't be able to read (cuz the user can't read B) and if you set security_invoker=false, you will be able to read because the view is "its own black box" as long as the user that created it has access to A and B (assuming your user has select permissions on that view)

repositories/analytics_views/users.sql
Pascal-Delange: IMO anything that goes to segment is fair game to see on metabase. In particular I'd not be too stressed about the email. The firebase uid, I don't know, I don't think it's much more confidential but then again it's kinda useless to know it so fine to remove it from here

repositories/analytics_views/data_model_fields.sql
Pascal-Delange: oopsies

Pascal-Delange: ok actually it's not "oopsies", it's just that "type" is a postgres keyword so the sql formatter tries to apply its own format (it's not a reserved keyword though mind you, so it's fine to use it as a field name https://www.postgresql.org/docs/current/sql-keywords-appendix.html) 
the alternative is to use
```
SELECT
      id,
      table_id,
      name,
      nullable,
      description,
      is_enum,
      "type" AS "type"
```
I'm just not sure if it's worth it

repositories/migrations.go
Pascal-Delange: Hmm, you mean if we remove a table or such, right 🤔 

Actually, most likely we need to drop the view before we drop the corresponding table, so we'd be kinda forced to do this in the migration file anyway

PR #464, opened 2024-01-30
feat(ingestion): improve error messages

balzdur:
By doing ingestion locally, I found that :
- there is an error on the line number computation
- we can leverage some usefull CSV parser error (to help the user understand why there is a problem)

As we display raw error messages in the frontend, I assume we can live with that for now

### DIFF
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cockroachdb/errors"
+
 	"github.com/google/uuid"
 
 	"github.com/checkmarble/marble-backend/models"
@@ -102,8 +104,8 @@ func (usecase *IngestionUseCase) ValidateAndUploadIngestionCsv(ctx context.Conte
 			break
 		}
 		if err != nil {
-			if parseError, ok := err.(*csv.ParseError); ok {
-				return models.UploadLog{}, fmt.Errorf("%w (%w)", parseError, models.BadParameterError)
+			if errors.As(err, &csv.ParseError{}) {
+				return models.UploadLog{}, fmt.Errorf("%w (%w)", err, models.BadParameterError)
 			}
 			return models.UploadLog{}, fmt.Errorf("error found at line %d in CSV (%w)", lineNumber, models.BadParameterError)
 		}
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -95,17 +95,22 @@ func (usecase *IngestionUseCase) ValidateAndUploadIngestionCsv(ctx context.Conte
 
 	var processedLinesCount int
 	for processedLinesCount = 0; ; processedLinesCount++ {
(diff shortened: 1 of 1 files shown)


usecases/ingestion_usecase.go
```diff
@@ -95,17 +95,22 @@ func (usecase *IngestionUseCase) ValidateAndUploadIngestionCsv(ctx context.Conte
...
 	for processedLinesCount = 0; ; processedLinesCount++ {
+		// line number starts at 1, and we already read the first line as headers
+		lineNumber := processedLinesCount + 2
 		row, err := fileReader.Read()
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
-			return models.UploadLog{}, fmt.Errorf("error found at line %d in CSV (%w)", processedLinesCount+1, models.BadParameterError)
+			if parseError, ok := err.(*csv.ParseError); ok {
```
Pascal-Delange, line 105: excellent initiative ! 
Just one formal thing: here it's probably equivalent since we're directly consuming the library's error, but we try to use errors.Is / As (so in this case, As because we're checking against an interface type), could you do the change ? 
See https://gosamples.dev/check-error-type/
(and by importing "github.com/cockroachdb/errors" rather than the std lib's error package)

balzdur: La seule manière de contourner que je vois, c'est faire un check sur les 3 erreurs sentinelles en utilisant le `.IS` :

Extract du code source :
```go

// A ParseError is returned for parsing errors.
// Line numbers are 1-indexed and columns are 0-indexed.
type ParseError struct {
	StartLine int   // Line where the record starts
	Line      int   // Line where the error occurred
	Column    int   // Column (1-based byte index) where the error occurred
	Err       error // The actual error
}

func (e *ParseError) Error() string {
	if e.Err == ErrFieldCount {
		return fmt.Sprintf("record on line %d: %v", e.Line, e.Err)
	}
	if e.StartLine != e.Line {
		return fmt.Sprintf("record on line %d; parse error on line %d, column %d: %v", e.StartLine, e.Line, e.Column, e.Err)
	}
	return fmt.Sprintf("parse error on line %d, column %d: %v", e.Line, e.Column, e.Err)
}

func (e *ParseError) Unwrap() error { return e.Err }

// These are the errors that can be returned in ParseError.Err.
var (
	ErrBareQuote  = errors.New("bare \" in non-quoted-field")
	ErrQuote      = errors.New("extraneous or missing \" in quoted-field")
	ErrFieldCount = errors.New("wrong number of fields")

	// Deprecated: ErrTrailingComma is no longer used.
	ErrTrailingComma = errors.New("extra delimiter at end of line")
)

```

Pascal-Delange: Ah sorry, ça marche quand même en faisant:
```
var csvErr *csv.ParseError
if errors.As(err, &csvErr) { ... } 
```

(example from the package doc https://go.dev/play/p/nBACxI0jkdp)

Pascal-Delange (approved):
thanks!

PR #469, opened 2024-02-01
Thomas/mar 412 api keys page

balzdur:
First PR on the topic :
- take into account `deleted_at` (existing but ignored for now, not a real change since we cannot delete by API for now)
- add `description` column

NB: look at the second commit, I propose to add more context to api key name (IIUC, this is only used for logging purpose). I'm not totally sure of what is really collected by logs, but I think this would be cool to identify the used api key in the end (like the email help us identify the user). Still, using user ID or apikey ID fills better for this kind of thing... I'm not sure of what to do for that.

### DIFF
--- a/repositories/dbmodels/db_api_keys.go
+++ b/repositories/dbmodels/db_api_keys.go
@@ -2,25 +2,27 @@ package dbmodels
 
 import (
 	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/utils"
 
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
 type DBApiKey struct {
-	Id             string      `db:"id"`
-	OrganizationId string      `db:"org_id"`
-	Key            string      `db:"key"`
-	DeletedAt      pgtype.Time `db:"deleted_at"`
-	Role           int         `db:"role"`
+	Id             string             `db:"id"`
+	OrganizationId string             `db:"org_id"`
+	Key            string             `db:"key"`
+	Description    string             `db:"description"`
+	DeletedAt      pgtype.Timestamptz `db:"deleted_at"`
+	Role           int                `db:"role"`
 }
-
-var ApiKeyFields = []string{"id", "org_id", "key", "deleted_at", "role"}
 
 const TABLE_APIKEYS = "apikeys"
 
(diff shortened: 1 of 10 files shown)


repositories/dbmodels/db_api_keys.go
```diff
@@ -2,25 +2,27 @@ package dbmodels
...
-	Id             string      `db:"id"`
-	OrganizationId string      `db:"org_id"`
-	Key            string      `db:"key"`
-	DeletedAt      pgtype.Time `db:"deleted_at"`
-	Role           int         `db:"role"`
+	Id             string             `db:"id"`
+	OrganizationId string             `db:"org_id"`
+	Key            string             `db:"key"`
+	Description    string             `db:"description"`
+	DeletedAt      pgtype.Timestamptz `db:"deleted_at"`
```
Pascal-Delange, line 15: please remind me of the difference with pgtype.Time ?

Pascal-Delange (approved):
👍

repositories/dbmodels/db_api_keys.go
balzdur: TBH, when doing this, I just dopy-paste this from `User` code.

After doing some research, it looks like `pgtype.Time` do not take the TS into account :
```go

// Timestamp represents the PostgreSQL timestamp type.
type Timestamp struct {
	Time             time.Time // Time zone will be ignored when encoding to PostgreSQL.
	InfinityModifier InfinityModifier
	Valid            bool
}

// Timestamptz represents the PostgreSQL timestamptz type.
type Timestamptz struct {
	Time             time.Time
	InfinityModifier InfinityModifier
	Valid            bool
}
```

balzdur: I found inside the `Timestamp` code this function :
```go 
func discardTimeZone(t time.Time) time.Time {
	if t.Location() != time.UTC {
		return time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), time.UTC)
	}

	return t
}
```

balzdur: In the end, the rule of thumb I use would be :
> Look at the sql schema, if this is `TIMESTAMP WITH TIME ZONE` use `pgtype. Timestamptz` else `pgtype.Time`

PR #470, opened 2024-02-02
feat(api-key): implement api key creation and management

balzdur:
Second iteration on the API Key implementation :
- create api key
- delete api key
- get list of api keys

Good to know : introduced notion of "hash" vs "value" for keys
- listing keys won't return value or hash to the client
- creating key will return the value (look at the usecase for further info)

Bonus: add really basic test (just so I can introduce myself on those)


integration_test/scenario_flow_test.go
```diff
@@ -80,13 +80,19 @@ func TestScenarioEndToEnd(t *testing.T) {
...
+		OrganizationId: organizationId,
+		Description:    "Test API key",
+		Role:           models.API_CLIENT,
+	})
+	assert.NoError(t, err, "Could not create api key")
+
+	// Generate creds from the created API Key
 	marbleTokenUsecase := usecasesWithCreds.NewMarbleTokenUseCase()
-	creds, err := marbleTokenUsecase.ValidateCredentials(ctx, "", apiKeys[0].Key)
+	creds, err := marbleTokenUsecase.ValidateCredentials(ctx, "", apiKey.Hash)
```
balzdur, line 95: For now, `Hash = raw key value` but it's future proof

models/permission.go
```diff
@@ -37,6 +39,8 @@ func (r Permission) String() string {
 		"DATA_MODEL_READ",
 		"DATA_MODEL_WRITE",
 		"APIKEY_READ",
+		"APIKEY_CREATE",
+		"APIKEY_DELETE",
```
balzdur, line 43: No sure if we want to have two separate permissions for that

repositories/dbmodels/db_api_key.go
```diff
@@ -10,7 +10,7 @@ import (
 type DBApiKey struct {
 	Id             string             `db:"id"`
 	OrganizationId string             `db:"org_id"`
-	Key            string             `db:"key"`
+	Hash           string             `db:"key"`
```
balzdur, line 13: We may want to alter the table column but I think this should be `key` until we implement hashing

repositories/migrations/20240202145805_remove_default_apikey_role.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE apikeys ALTER COLUMN role DROP DEFAULT;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE apikeys ALTER COLUMN role SET DEFAULT 5;
+-- +goose StatementEnd
```
balzdur, line 9: From now on, we need to explicitly set the Api key role

repositories/repositories.go
```diff
@@ -68,9 +67,6 @@ func NewRepositories(
 		UserRepository: &UserRepositoryPostgresql{
 			transactionFactory: transactionFactory,
 		},
-		ApiKeyRepository: &ApiKeyRepositoryImpl{
-			transactionFactory: transactionFactory,
-		},
```
balzdur, line 73: I removed the useless `ApiKeyRepositoryImpl`, taking inspiration from :
- what Chris present in the `postgress` directory
- what latest repositories (like inbox one) done

In the end:
- move `ApiKeyRepository` interface def were it's needed
- implement directly on `MarbleDbRepository` all those interfaces

IMO, final questions on the repository topic are more or less :
- should we still be using squirel + transaction functions or plain SQL when possible ? (or in between)
- should we move the transaction from param to interface implementing repositories (more or less like what is done by Chris) moving out the sugar function 
```go
pgTx := repo.transactionFactory.adaptMarbleDatabaseTransaction(ctx, tx)
```

usecases/marble_token_usecase.go
```diff
@@ -15,10 +15,12 @@ type MarbleTokenUseCase struct {
 	marbleJwtRepository     repositories.MarbleJwtRepository
 	firebaseTokenRepository repositories.FireBaseTokenRepository
 	userRepository          repositories.UserRepository
-	apiKeyRepository        repositories.ApiKeyRepository
-	organizationRepository  repositories.OrganizationRepository
-	tokenLifetimeMinute     int
-	context                 context.Context
+	apiKeyRepository        interface {
+		GetApiKeyByKey(ctx context.Context, tx repositories.Transaction, apiKey string) (models.ApiKey, error)
+	}
```
balzdur, line 20: each usecase declare it's own `ApiKeyRepository` interface

usecases/organization/organization_seeder.go
```diff
@@ -13,42 +10,17 @@ import (
...
 )
 
-func randomAPiKey() string {
-	var key = make([]byte, 8)
-	_, err := rand.Read(key)
-	if err != nil {
-		panic(fmt.Errorf("randomAPiKey: %w", err))
-	}
-	return hex.EncodeToString(key)
-}
```
balzdur, line 23: Moved to the apikey use case

usecases/organization/organization_seeder.go
```diff
@@ -13,42 +10,17 @@ import (
...
-
-	err := o.ApiKeyRepository.CreateApiKey(ctx, nil, models.CreateApiKeyInput{
-		OrganizationId: organizationId,
-		Key:            randomAPiKey(),
-		Description:    "Default API Key",
-	})
-	if err != nil {
-		log.Printf("error creating token: %v", err)
-		return err
-	}
```
balzdur, line 44: Removed, as API key creation will be self servre in the frontend ASAP

usecases/organization_usecase.go
```diff
@@ -113,19 +112,3 @@ func (usecase *OrganizationUseCase) GetUsersOfOrganization(ctx context.Context,
...
-				return []models.ApiKey{}, err
-			}
-			for _, ak := range apiKeys {
-				if err := usecase.enforceSecurity.ReadOrganizationApiKeys(ak); err != nil {
-					return []models.ApiKey{}, err
-				}
-			}
-			return apiKeys, nil
-		})
-}
```
balzdur, line 131: Removed from org usecase.
Everything related to Api keys is included in the new `api_key_usecases` file

usecases/security/enforce_security_api_key.go
```diff
@@ -0,0 +1,30 @@
...
+	return errors.Join(
+		e.Permission(models.APIKEY_DELETE), e.ReadOrganization(apiKey.OrganizationId),
+	)
+}
+
+func (e *EnforceSecurityApiKeyImpl) ListApiKeys() error {
+	return errors.Join(
+		e.Permission(models.APIKEY_READ),
+	)
+}
```
balzdur, line 30: It's not really clear for me :
- sometimes, we check explciitly `e.ReadOrganization(apiKey.OrganizationId)` for read operation too
- sometimes, we do not

Still, usecases on reading info linked to org_id is often based on the orgId from context, but maybe we should do the check explicitly (should be transparent and cost nothing) WDYT ?

usecases/api_key_usecase.go
```diff
@@ -0,0 +1,120 @@
...
+	var key = make([]byte, 32)
+	_, err := rand.Read(key)
+	if err != nil {
+		panic(fmt.Errorf("randomAPiKey: %w", err))
+	}
+	return hex.EncodeToString(key)
+}
+
+func (usecase *ApiKeyUseCase) DeleteApiKey(ctx context.Context, apiKeyId string) error {
+	err := transaction.TransactionFactory.Transaction(
```
Pascal-Delange, line 97: Here I think we can omit doing a transaction

usecases/api_key_usecase.go
```diff
@@ -0,0 +1,120 @@
...
+	if err := usecase.enforceSecurity.ListApiKeys(); err != nil {
+		return []models.ApiKey{}, err
+	}
+
+	organizationId, err := usecase.organizationIdOfContext()
+	if err != nil {
+		return []models.ApiKey{}, err
+	}
+
+	return usecase.apiKeyRepository.ListApiKeys(ctx, nil, organizationId)
```
Pascal-Delange, line 46: Here, I'd like to do: for every api key in output, validate that its org id corresponds to the input org id (see for instance how it's done in cases)

usecases/security/enforce_security_api_key.go
Pascal-Delange: (see comment in the usecase file)

repositories/repositories.go
Pascal-Delange: Great ! 
My take on the two open questions:
- plain sql vs squirrel, I don't have a strong opinion either way - I can live with both, I just think that squirrel serves an important niche in the "mid complexity" queries with templated parameters. Especially for update queries, "where" conditions
- transaction passing: Here I have to say I disagree with Chris and I really like having one method to call, nevermind whether I want to execute it in a transaction or not. Certainly we can make `adaptMarbleDatabaseTransaction` a bit nicer to use (but then again there are probably limits to what we can do) but I prefer this to the other way

Pascal-Delange (commented):
Thanks for the work ! A few answers to your questions are included, just one thing I'd like to change on permissions checks (see usecase file)

dto/apikey_dto.go
```diff
@@ -14,8 +13,24 @@ func AdaptApiKeyDto(apiKey models.ApiKey) ApiKey {
...
+	ApiKey
+	Key string `json:"key"`
+}
+
+func AdaptCreatedApiKeyDto(apiKey models.CreatedApiKey) CreatedApiKey {
+	return CreatedApiKey{
+		ApiKey: AdaptApiKeyDto(apiKey.ApiKey),
+		Key:    apiKey.Value,
+	}
+}
```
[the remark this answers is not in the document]
balzdur: Totally agree with you. I found it weird first, but I think this is due to the lack of "spread" operator in Go (like `...obj` in JS/TS).

In the end, using this kind of composition only for DTO, with the go nesting wrapped inside adapters should not be a problem IMO, but I do not have strong opinion between this, or duplicate all the code every time (since we do have CoPilot to do it for us and this should be written once).

(small) Bonus: having duplication for all those types makes them trully independent (so easier to introduce independant changes on each endpoints).

This is just a lazy way to do things

repositories/repositories.go
balzdur: Ok, I missred Chris code. What I though was more or less :
- same abstraction as current code (meaning a single endpoint for each repository method, with or without transactions)
- transaction handling done outside the repository method : meaning no more `adaptMarbleDatabaseTransaction` done inside each function on the injected `tx` parameter. 

What I had in mind is more or less described by:
```go
// In repository
func (repo *MarbleDbRepository) UpdateUser(ctx context.Context, userId string) error {
    pgTx := repo.transactionFactory.adaptMarbleDatabaseTransaction(ctx)
    
    // do internal logic here,
}

// in usecase
func (usecase *UserUseCase) UpdateUser(ctx context.Context, updateUser models.UpdateUser) (models.User, error) {
	tx, err := usecase.transactionFactory.Begin(ctx, models.DATABASE_MARBLE_SCHEMA)
	defer tx.Rollback(ctx) // should be wrapped inside our Begin implementation but you have the idea

        userRepository, err :=  usecase.getUserRepository(tx) // but by writting this, it feels not necessarly usefull compared to `tx` param injection
	if err != nil {
		return models.User{}, err
	}
	user, err := userRepository.UserByID(ctx, tx, updateUser.UserId)
	if err != nil {
		return models.User{}, err
	}
	if err := usecase.enforceUserSecurity.UpdateUser(user); err != nil {
		return models.User{}, err
	}
	if err := userRepository.UpdateUser(ctx, tx, updateUser); err != nil {
		return models.User{}, err
	}
	user, err = userRepository.UserByID(ctx, tx, updateUser.UserId)
	if err != nil {
		return models.User{}, err
	}

	tx.Commit(ctx)
	analytics.TrackEvent(ctx, models.AnalyticsUserUpdated, map[string]interface{}{"user_id": updatedUser.UserId})

	return updatedUser, nil
}
```

balzdur: FYI, digging inside `pgx` issues to understand how transaction actually work (= can we safelly defer `Rollback`) I found [this issue](https://github.com/jackc/pgx/issues/1667) that may be interesting for us.

As we wrap every call (even simple `SELECT` ones) inside a transaction, it may add a lot of roundtrips to our decision computation (when reading the DB). It's difficult to pronounce w\ measuring but removing tx on simpler call could result in a faster execution (at least, until the lazy thing is implemented inside pgx)

dto/apikey_dto.go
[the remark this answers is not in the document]
Pascal-Delange: yea agreed, let's go for this solution 👍

repositories/repositories.go
Pascal-Delange: Well in fact we're already doing the transaction handling (AKA starting it & committing it) in the usecase, the repo just has an adapter function. Which we can perhaps simplify a bit, but I'm not sure there is that much "fat" to skim on it.

PR #472, opened 2024-02-09
Thomas/mar 471 implement analytics page

balzdur:
Introduce basic usecase to get signed embeded URL from metabase:
- rename old `analytics` package to `tracking`
- implement metabase related stuff inside `infra` package
- abstract away metabse from repo/usecase code as an interface to generate signed URL

The only stuff still linked to Metabase is the untyped Payload data.


models/analytics.go
```diff
@@ -0,0 +1,27 @@
+package models
+
+type EmbeddingId int
```
Pascal-Delange, line 3: I'd call it "type" rather than "id" because "id" implies "unique identifier" to me, but just a matter of personal taste perhaps

repositories/analytics_repository.go
```diff
@@ -0,0 +1,39 @@
...
+	metabase Metabase
+}
+
+func (repo *MarbleAnalyticsRepository) ListAnalytics(ctx context.Context, organizationId string) ([]models.Analytics, error) {
+	var analytics []models.Analytics
+
+	// Add general dashboard
+	generalDashboardUrl, err := repo.metabase.GenerateSignedEmbeddingURL(models.AnalyticsCustomClaims{
+		Resource: map[string]interface{}{
+			"dashboard": 8,
```
Pascal-Delange, line 23: This should be injected from an env variable, no ? (to separate staging from prod)

Pascal-Delange (commented):
Very clear ! One small comment to take into account about the embedding number injection

utils/env_vars.go
```diff
@@ -7,6 +7,18 @@ import (
 	"strconv"
 )
 
+func GetRequiredIntEnv(envVarName string) int {
+	envValue, ok := os.LookupEnv(envVarName)
+	if !ok || envValue == "" {
+		log.Fatalf("%s environment variable is required", envVarName)
+	}
+	intValue, err := strconv.Atoi(envValue)
```
Pascal-Delange, line 15: for what it's worth, technically this by itself covers all the errors you can have above too

models/analytics.go
```diff
@@ -21,7 +32,21 @@ type Analytics struct {
...
+	OrganizationId string
+}
+
+func (analytics GlobalDashboardAnalytics) GetEmbeddingType() EmbeddingType {
+	return GlobalDashboard
+}
+
+func (analytics GlobalDashboardAnalytics) GetParams() map[string]interface{} {
+	return map[string]interface{}{
+		"organization": []string{analytics.OrganizationId},
```
Pascal-Delange, line 50: previously it seemed to be org_id, was it a change on metabase ?

Pascal-Delange (approved):
👍  small typo check question on org_id, otherwise all good for me !

models/analytics.go
balzdur: This part depends on the way embeding is configured.
The last one I tested was using `organization` other `org_id`.

This kind of divergence is one of the reason I moved to a "interface" style directly inside `models.analytics` so each metabase embedding can be easilly customised looking at a single file.

Basically, I do the following :

1. Ask for the Metabase embedding link
2. Open the embedding window from the interface
3. Look at the code section
4. Copy/Paste the information to implement the new embedding as the Go interface described in models (e.g. embeding type, params form...)

utils/env_vars.go
balzdur: Mmh... I think I miss something. How do you use `strconv.Atoi` without calling `os.LookupEnv` first here ?

Pascal-Delange: Hmm non en fait ce que je voulais dire (mais c'était peut-être con) c'était que tu pourrais faire un LookupEnv sans même check le "ok" et passer directement l'output (au pire c'est une string vide) à Atoi() puis check l'erreur juste là
mais c'était pas forcément un commentaire hper pertinent

PR #474, opened 2024-02-14
big refactor: transction injection

Pascal-Delange:


usecases/executor_factory/helpers.go
```diff
@@ -42,4 +42,11 @@ func TransactionReturnValue[ReturnType any](
 type TransactionFactory interface {
 	TransactionInOrgSchema(ctx context.Context, organizationId string, f func(tx repositories.Executor) error) error
 	Transaction(ctx context.Context, fn func(tx repositories.Executor) error) error
+	// Transaction(ctx context.Context, fn func(tx repositories.Executor) error) error
+}
+
+// Interface to be used in usecases, implemented by the DbExecutorFactory class in the usecases/db_executor_factory package
+// which itself has the ExecutorGetter repository class injected in it.
+type ClientSchemaExecutorFactory interface {
```
Pascal-Delange, line 50: I had to move this back here (from usecases) because of circular dependencies

models/ast/ast_function.go
```diff
@@ -55,12 +55,6 @@ const (
 	FUNC_AGGREGATOR
 	FUNC_LIST
 	FUNC_FILTER
-	FUNC_BLANK_FIRST_TRANSACTION_DATE
```
Pascal-Delange, line 58: side change: I removed all code referring to the blank organization

repositories/db_executor.go
```diff
@@ -36,3 +38,20 @@ func (e ExecutorPostgres) Query(ctx context.Context, sql string, args ...any) (p
...
+	if err != nil {
+		return 0, errors.Wrap(err, "can't build sql query")
+	}
+
+	tag, err := exec.Exec(ctx, query, args...)
+	if err != nil {
+		return 0, errors.Wrap(err, fmt.Sprintf("error executing sql query: %s", query))
+	}
+	return tag.RowsAffected(), nil
+}
```
balzdur, line 57: I would be in favour of removing this helper :
- It introduce an abstraction that makes code less understandable when reading repositories methods
- It seems the `tag.RowsAffected()` value is always ignored

We can keep it but IMO we need to either :
- moving it as an `Executor` method like this was done before (basically this is a sugar function that aims to directly inject a `SqlBuilder` in place of the `query`)
- move it in a separate file (this is more or less the same kind of stuff as `sql_to_row` `sql_to_models` : a utility function)

repositories/ingested_data_read_repository.go
```diff
@@ -13,20 +13,22 @@ import (
...
 }
 
 type IngestedDataReadRepositoryImpl struct{}
 
-func (repo *IngestedDataReadRepositoryImpl) GetDbField(ctx context.Context, transaction Transaction_deprec, readParams models.DbFieldReadParams) (any, error) {
-	tx := adaptClientDatabaseTransaction(transaction)
+func (repo *IngestedDataReadRepositoryImpl) GetDbField(ctx context.Context, exec Executor, readParams models.DbFieldReadParams) (any, error) {
+	if err := validateClientDbExecutor(exec); err != nil {
+		return nil, err
+	}
```
balzdur, line 26: Verry nice 🎉 
It's soooo clearer

repositories/ingestion_repository.go
```diff
@@ -13,37 +13,38 @@ import (
...
 
 	if len(payloadsToInsert) > 0 {
-		if err := repo.batchInsertPayloadsAndEnumValues(ctx, tx, payloadsToInsert, table); err != nil {
+		if err := repo.batchInsertPayloadsAndEnumValues(ctx, exec, payloadsToInsert, table); err != nil {
 			return err
 		}
 	}
 
-	logger.Info("Inserted objects in db", slog.String("type", tableNameWithSchema(tx, table.Name)), slog.Int("nb_objects", len(payloadsToInsert)))
+	logger.Info("Inserted objects in db", slog.String("type", tableNameWithSchema(exec, table.Name)), slog.Int("nb_objects", len(payloadsToInsert)))
```
balzdur, line 47: Question unrelated with the scope of this PR: IIUC, we plan to use the logger from the `ctx` in the long term, and remove all injection like this one here ?

usecases/executor_factory/helpers.go
```diff
@@ -42,4 +42,11 @@ func TransactionReturnValue[ReturnType any](
 type TransactionFactory interface {
 	TransactionInOrgSchema(ctx context.Context, organizationId string, f func(tx repositories.Executor) error) error
 	Transaction(ctx context.Context, fn func(tx repositories.Executor) error) error
+	// Transaction(ctx context.Context, fn func(tx repositories.Executor) error) error
```
balzdur, line 45: Probably to delete

usecases/executor_factory/helpers.go
balzdur: No problem, and I think we can create a `MarbleSchemaExecutorFactory` or something like that (linked to another comment on the `isNil`)

usecases/inboxes/inbox_users.go
```diff
@@ -136,11 +136,11 @@ func (usecase *InboxUsers) CreateInboxUser(ctx context.Context, input models.Cre
 }
 
 func (usecase *InboxUsers) UpdateInboxUser(ctx context.Context, inboxUserId string, role models.InboxUserRole) (models.InboxUser, error) {
-	inboxUser, err := transaction.TransactionReturnValue_deprec(
+	inboxUser, err := executor_factory.TransactionReturnValue(
```
balzdur, line 139: It looks weird to me.

Either we have utility function that take `usecase.TransactionFactory` as a param like this, or we have method lile `usecase.TransactionFactory.Transaction(...)` (cf `DeleteInboxUser` implementation)

usecases/user_usecase.go
```diff
@@ -6,28 +6,28 @@ import (
...
 		return models.User{}, err
 	}
-	createdUser, err := transaction.TransactionReturnValue_deprec(
+	createdUser, err := executor_factory.TransactionReturnValue(
 		ctx,
 		usecase.transactionFactory,
 		models.DATABASE_MARBLE_SCHEMA,
-		func(tx repositories.Transaction_deprec) (models.User, error) {
-
+		func(tx repositories.Executor) (models.User, error) {
```
balzdur, line 30: Ideally, we'll go for 
```suggestion
		func(exec repositories.Executor) (models.User, error) {
```

But I can live with that, it's painfull to replace

usecases/usecases.go
```diff
@@ -82,7 +72,7 @@ func (usecases *Usecases) NewExportScheduleExecution() *scheduledexecution.Expor
 
 func (usecases *Usecases) NewPopulateOrganizationSchema() organization.PopulateOrganizationSchema {
 	return organization.PopulateOrganizationSchema{
-		TransactionFactory:           &usecases.Repositories.TransactionFactoryPosgresql_deprec,
+		ClientSchemaExecutorFactory:  usecases.NewClientDbExecutorFactory(),
```
balzdur, line 75: I notice we move from `NewOrgTransactionFactory` to `NewClientDbExecutorFactory` but we still have `NewTransactionFactory`: is there a reason to distinguish the naming clearly like this ?

usecases/scenario_publications_usecase.go
```diff
@@ -120,17 +118,6 @@ func (usecase *ScenarioPublicationUsecase) CreateDatamodelIndexesForScenarioPubl
...
-		ctx,
-		usecase.transactionFactory,
-		organizationId,
-		func(tx repositories.Executor) ([]models.ConcreteIndex, error) {
-			return usecase.ingestedDataIndexesRepository.ListAllValidIndexes(ctx, tx)
-		},
-	)
-	if err != nil {
-		return false, errors.Wrap(err, "Error while fetching existing indexes in CreateDatamodelIndexesForScenarioPublication")
-	}
```
balzdur, line 133: I assume this is done on purpose

usecases/ast_eval/evaluate/evaluate_aggregator.go
```diff
@@ -100,18 +100,11 @@ func (a AggregatorEvaluator) runQueryInRepository(ctx context.Context, tableName
...
-				return nil, err
-			}
-			return result, nil
-		},
-	)
+	if db, err := a.ClientSchemaExecutorFactory.NewClientDbExecutor(ctx, a.OrganizationId); err != nil {
+		return nil, err
+	} else {
+		return a.IngestedDataReadRepository.QueryAggregatedValue(ctx, db, tableName, fieldName, aggregator, filters)
+	}
```
balzdur, line 107: useless else (syntax nazy here, but I like the simple `if / return` structure of go which limit nesting

usecases/ast_eval/evaluate/evaluate_database_access.go
```diff
@@ -72,17 +72,15 @@ func (d DatabaseAccess) getDbField(ctx context.Context, tableName models.TableNa
...
-				TriggerTableName: models.TableName(tableName),
-				Path:             models.ToLinkNames(path),
-				FieldName:        models.FieldName(fieldName),
-				DataModel:        d.DataModel,
-				Payload:          d.Payload,
-			})
+	if db, err := d.ClientSchemaExecutorFactory.NewClientDbExecutor(ctx, d.OrganizationId); err != nil {
+		return nil, err
+	} else {
+		return d.IngestedDataReadRepository.GetDbField(ctx, db, models.DbFieldReadParams{
```
balzdur, line 78: Same here (💅🏽)

balzdur (approved):

repositories/decisions_repository.go
```diff
@@ -15,33 +15,33 @@ import (
...
+	executorGetter ExecutorGetter
 }
 
 // the size of the batch is chosen without any benchmark
 const decisionRulesBatchSize = 1000
 
-func (repo *DecisionRepositoryImpl) DecisionById(ctx context.Context, transaction Transaction_deprec, decisionId string) (models.Decision, error) {
-	tx := repo.transactionFactory.adaptMarbleDatabaseTransaction(ctx, transaction)
+func (repo *DecisionRepositoryImpl) DecisionById(ctx context.Context, exec Executor, decisionId string) (models.Decision, error) {
+	exec = repo.executorGetter.ifNil(exec)
```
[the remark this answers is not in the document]
Pascal-Delange: Alright, as discussed orally I'll rework this a bit

repositories/db_executor.go
Pascal-Delange: I'd still rather keep it because it removes tons of boilerplate in all repo methods, but fine with moving it to elsewhere (probably where the SqlToRow etc are).
Not a fan of making it an Executor Method because it makes the interface more complex.

repositories/ingested_data_read_repository.go
Pascal-Delange: yes 👌

repositories/ingestion_repository.go
Pascal-Delange: yea absolutely, it's supposed to the case already. I'll clean that up, rather in another PR

usecases/user_usecase.go
Pascal-Delange: well actually I left it like that on purpose, to clarify in the code within the closure that it's actually a transaction being used

usecases/usecases.go
Pascal-Delange: NewTransactionFactory is now for generating transactions (on both schema types) while NewClientDbExecutorFactory creates an executor (wrapping a pool)

Pascal-Delange: (and it's the same underlying implementation, but separate interfaces)

usecases/ast_eval/evaluate/evaluate_aggregator.go
Pascal-Delange: hmm yea that was in order to keep the "if err!=nil" condition more inline, but in the end I agree I gain nothing from it

repositories/db_executor.go
balzdur: TBH, from the time I first wrote this comment to the end of the github review, I finally forged my opininion and I'm aligned with you.

Pascal-Delange:
Updated as discussed this morning: executors are passed as parameters everywhere, no more "fallback from nil".
This means an instance of ExecutorFactory in every usecase.
I double checked that it's properly injected everywhere in the usecase constructors, though that was a bit painful. Ultimately I'd like to move for all usecase constructors to something like [here](https://github.com/checkmarble/marble-backend/pull/474/files#diff-e5538d7c2ad6965f30a1be1f9543dd3050ebdbfbecdf51185c4049640988fb70R23) in NewCustomListValuesAccess: have a constructor next to the class definition.
Vivien disliked those because "we're only using it once in the usecases constructor file", but in fact:
- having it next to the class definition makes it easy to check it's up to date with the class definition when adding new fields
- then we no longer need to have public (capitalized) fields on sub-module classes just to explicitly build them in the usecases builder
- when we use builders, it's much easier where we use them to check we didn't forget one (thanks to go type check).
But I'm not going to do it today

Pascal-Delange:
Also to do (I didn't do it here to not cause confusion): I want to do a big re-read of all repos & usecases to rationalize where we use transactions and where not, because it's all over the place (some useless transactions, some places where we should use one)

repositories/decisions_repository.go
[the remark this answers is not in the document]
Pascal-Delange: done, see a later commit !

balzdur (approved):

PR #476, opened 2024-02-15
try new stricter formatting

Pascal-Delange:
- commit 1: general gofumpt rules (no brainer IMO)
- commit 2: more debatable rule on max line length (but I'm very in favor of it)
- commit 3: add Thomas' settings file

See discussion in https://github.com/mvdan/gofumpt/issues/2 on max length formatting. It works with a set max line length of 120 chars and some heuristics to produce reasonable looking code. You can always indent it more manually.
To be clear: I don't think it always results in the single most optimal indentation, which can be achieved manually where needed, but at least it forces something (and does it by being conservative rather than aggressive, so as to not break anyone's manual indentation).
It's an experimental flag so run it in terminal with `GOFUMPT_SPLIT_LONG_LINES=on gofumpt -l -w .`

and https://github.com/mvdan/gofumpt?tab=readme-ov-file for gofumpt generally


balzdur (approved):
I have no strong opinion on lines length.
I really like having a better formatter than the Go standard one (comming from Prettier world, it's not enough work done for me ahha)

Pascal-Delange:
Yep I'm not a line length nazi either, but too much is too much (and ~120 is reasonable, I don't want to horizontally scroll my code).

I still prefer "all params in one line or all on split lines", but I can still do that manually (wasn't motivated to apply it manually everywhere though).

It will apply this format on save with the changes made to settings.json (gopls will choose the formatter from the language server and use gofumpt as configured)

PR #480, opened 2024-02-19
Pascal/actual index creation api

Pascal-Delange:


usecases/scenario_publications_usecase.go
```diff
@@ -114,62 +114,118 @@ func (usecase *ScenarioPublicationUsecase) ExecuteScenarioPublicationAction(
...
+	ctx context.Context,
+	scenarioIterationId string,
+) (toCreate []models.ConcreteIndex, numPending int, err error) {
+	exec := usecase.executorFactory.NewExecutor()
+	iterationToActivate, err := usecase.scenarioFetcher.FetchScenarioAndIteration(ctx, exec, scenarioIterationId)
 	if err != nil {
-		return false, errors.Wrap(err, "Error while fetching active scenario iterations in CreateDatamodelIndexesForScenarioPublication")
+		return toCreate, numPending, err
+	}
+	if err := usecase.enforceSecurity.PublishScenario(iterationToActivate.Scenario); err != nil {
```
Pascal-Delange, line 195: Open question: is this the right permission also for getting the status ? 
At least, we'll have to make sure we don't call it with non publisher users if so

balzdur (approved):

PR #482, opened 2024-02-19
refacto: allow YYYY-MM-DD hh:mm:ss timestamp format, and remove the o…

Pascal-Delange:
…ld payload parser


api/handle_decision.go
```diff
@@ -122,17 +122,6 @@ func (api *API) handlePostDecision(c *gin.Context) {
 		return
 	}
 
-	payload, err := payload_parser.ParseToDataModelObject(table, requestData.TriggerObjectRaw)
```
Pascal-Delange, line 125: goodbye, old parser

usecases/payload_parser/payload_test.go
```diff
@@ -103,20 +125,82 @@ func TestParser_ValidatePayload(t *testing.T) {
...
+				},
+			},
+			input: []byte(`{"object_id": "id", "updated_at": "2023-10-19T00:00:00+03:00"}`),
+			want: models.ClientObject{
+				TableName: "transactions",
+				Data: map[string]any{
+					"nullable":  nil,
+					"object_id": "id",
+					// input is in UTC+3, but the output is in UTC
+					"updated_at": time.Date(2023, time.October, 18, 21, 0, 0, 0, time.UTC),
```
Pascal-Delange, line 177: note this test case

usecases/payload_parser/payload.go
```diff
@@ -151,74 +27,98 @@ var (
...
+	}, nil, nil
 }
 
 func NewParser() *Parser {
-	validators := fieldValidator{
-		models.Timestamp: func(result gjson.Result) error {
-			_, err := time.Parse(time.RFC3339, result.String())
-			if err != nil {
-				return fmt.Errorf("%w: expected format YYYY-MM-DDThh:mm:ss[+optional decimals]Z, got %s", errIsInvalidTimestamp, result.String())
+	validators := fieldParser{
```
balzdur, line 81:
```suggestion
	parsers := fieldParser{
```

usecases/payload_parser/payload.go
```diff
@@ -151,74 +27,98 @@ var (
...
+			t, err1 := time.Parse(time.RFC3339, result.String())
+			if err1 == nil {
+				return t.UTC(), nil
+			}
+			t, err2 := time.Parse("2006-01-02 15:04:05.9", result.String())
+			if err2 != nil {
+				return nil, fmt.Errorf("%w: expected format \"YYYY-MM-DD hh:mm:ss[+optional decimals]\" or \"YYYY-MM-DDThh:mm:ss[+optional decimals]Z\", got \"%s\"", errIsInvalidTimestamp, result.String())
 			}
-			return nil
+			return t.UTC(), nil
```
balzdur, line 91: I think this is clearer like this 
```go
if t, err := time.Parse(time.RFC3339, result.String()); err == nil {
      return t.UTC(), nil				
}
if t, err = time.Parse("2006-01-02 15:04:05.9", result.String()); err == nil {
      return t.UTC(), nil				
}
return nil, fmt.Errorf("%w: expected format \"YYYY-MM-DD hh:mm:ss[+optional decimals]\" or \"YYYY-MM-DDThh:mm:ss[+optional decimals]Z\", got \"%s\"", errIsInvalidTimestamp, result.String())
```

I thiknk the pattern "try to parse, if no error early return else cotinue and return error if no parsing works" is a good "common" pattern we could use for every parsers (still, since others are single case based, it's not really needed)

usecases/ingestion_usecase.go
```diff
@@ -354,11 +353,13 @@ func parseStringValuesToMap(headers []string, values []string, table models.Tabl
 		case models.String:
 			result[fieldName] = value
 		case models.Timestamp:
-			val, err := time.Parse(time.RFC3339, value)
-			if err != nil {
+			if val, err := time.Parse(time.RFC3339, value); err == nil {
+				result[fieldName] = val.UTC()
+			} else if val, err = time.Parse("2006-01-02 15:04:05.9", value); err != nil {
```
balzdur, line 358: Typo
```suggestion
			} else if val, err = time.Parse("2006-01-02 15:04:05.9", value); err == nil {
```

usecases/payload_parser/payload_test.go
balzdur: What about a non TZ timestamp (the new format) ? How does it behaves when the server do `t.UTC()` does it return a TZ based on the server one ? or does it infer it's UTC by default ?

balzdur (approved):
Single feedback, if we do delete the `models.PayloadReader` maybe we should also rename stuff :
- rename everything to have `Payload` (ex: in `models.Decision`)
- or rename `payload` to `clientObject` (since this is the type name now)

I prefer the second option (some of the changes are already reflecting this idea, as inside `usecases/ingestion_usecase.go`)

usecases/payload_parser/payload_test.go
Pascal-Delange: Yes it parses to UTC

usecases/ast_eval/evaluate/evaluate_read_payload.go
```diff
@@ -12,10 +12,10 @@ import (
 
 type Payload struct {
 	Function ast.Function
-	Payload  models.PayloadReader
+	Payload  models.ClientObject
```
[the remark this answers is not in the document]
Pascal-Delange: alright agreed, I'll do that tomorrow

usecases/payload_parser/payload.go
Pascal-Delange: true!

usecases/ast_eval/evaluate/evaluate_read_payload.go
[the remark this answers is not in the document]
Pascal-Delange: ok done

PR #484, opened 2024-02-20
fix: get preparation status works on an invalid iteration (no ast, in…

Pascal-Delange:
…valid ast...)

No fundamental reason why the data preparation endpoint should not be called on a not yet live, even on an as yet invalid scenario iteration. Even if there is no fundamental need for it either.


balzdur (commented):
IIUC, public API is still the same.

Ex: if the Draft/Version is invalid and tested, it may return "no indexes" since all ASTs are invalid but since we do not use the preparation status when invalid, it's totally ok

PR #488, opened 2024-02-22
reduce db pool max connections

Pascal-Delange:
To be considered in the context of https://github.com/checkmarble/terraform/pull/14.
Max_instances*max_connections_per_pool should remain broadly lesser than the db instance's max connections (this is an approximation, because we also have to account for cron jobs)

For context:
- the staging DB with 1 vcpu and 4Gb RAM has max 100 connections
- the prod DB with 4 vcpu and 1Gb RAM has max 500 connections
- max connections are broadly proportional to RAM IIUC

### DIFF
--- a/infra/postgres_connections_pool.go
+++ b/infra/postgres_connections_pool.go
@@ -3,12 +3,16 @@ package infra
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/exaring/otelpgx"
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
-const MAX_CONNECTIONS = 100
+const (
+	MAX_CONNECTIONS          = 40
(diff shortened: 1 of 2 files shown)


balzdur (approved):
Ideally, it would have been good to inject an env var (at least for the `MAX_CONNECTIONS`) so the IaC is responsible for injecting the value (which make sense since this is heavilly corelated to our infra configuration)

Pascal-Delange:
> Ideally, it would have been good to inject an env var (at least for the `MAX_CONNECTIONS`) so the IaC is responsible for injecting the value (which make sense since this is heavilly corelated to our infra configuration)

That's a good point.
I think I'll wait to do that for when I refactor all the env variable injection for the different job types in the backend main though (planned)

PR #491, opened 2024-02-26
Pascal/mar 458 data model links must point to a primary key

Pascal-Delange:


usecases/data_model_usecase.go
```diff
@@ -162,42 +183,171 @@ func (usecase *DataModelUseCase) CreateDataModelField(ctx context.Context, field
...
+	makeNotUnique = input.IsUnique != nil &&
+		!*input.IsUnique &&
+		currentField.UnicityConstraint != models.NoUnicityConstraint
+	if makeNotUnique && len(linksToField) > 0 {
+		return false, false, errors.Wrap(
+			models.BadParameterError,
+			"cannot remove unicity constraint on a field that is linked to another table")
+	}
+
+	if makeUnique && ((field.IsEnum && !makeNotEnum) || makeEnum) {
```
Pascal-Delange, line 286: fundamentally nothing bad would happen if a field were at the same time enum and unique, but it doesn't make sense so I want to prevent it in the front, and I'm adding the same constraint here to avoid the front being in a blocked situation (both fields toggled and disabled)

Pascal-Delange:
last thing to do: document state machine on notion

Pascal-Delange:
In the end I decided not to implement the feature of "return an expected error if multiple rows are actually read for a DB field read". The reason is that with this PR, this should never be able to happen anymore (as enforced by the DB unique index) and I'd rather avoid to include lots of actually unreachable code just for this edge case

usecases/data_model_usecase.go
```diff
@@ -66,10 +74,10 @@ func addUnicityConstraintStatusToDataModel(dataModel models.DataModel, uniqueInd
 				field.UnicityConstraint = models.ActiveUniqueConstraint
 			}
 			// cannot directly modify the struct field in the map, so we need to reassign it
-			dataModel.Tables[index.TableName].Fields[index.Fields[0]] = field
+			dm.Tables[index.TableName].Fields[index.Fields[0]] = field
```
balzdur, line 77: You could, with a deep copy.

usecases/data_model_usecase.go
```diff
@@ -162,42 +173,191 @@ func (usecase *DataModelUseCase) CreateDataModelField(ctx context.Context, field
 				return usecase.organizationSchemaRepository.CreateField(ctx, orgTx, table.Name, field)
 			},
 		)
-	})
-	return fieldId, err
+	}); err != nil {
+		return "", err
+	}
```
balzdur, line 178: I think for long statements like this, I prefer to have 
```go
err := ....

if err != nil {
   ...
}
```
But it's just a matter of formatting

usecases/data_model_usecase.go
```diff
@@ -162,42 +173,191 @@ func (usecase *DataModelUseCase) CreateDataModelField(ctx context.Context, field
...
+
+	if field.IsUnique {
+		if err := usecase.clientDbIndexEditor.CreateUniqueIndexAsync(
+			ctx,
+			getFieldUniqueIndex(models.TableName(tableName), field.Name)); err != nil {
+			return "", err
+		}
```
balzdur, line 185: Same here. Personally, this "shorthand" is meaningfull when it's a short one liner.
IMO, as soon as it's spread accross multiple lines, it's difficult to read.

usecases/data_model_usecase.go
```diff
@@ -114,34 +122,35 @@ func (usecase *DataModelUseCase) CreateDataModelTable(ctx context.Context, organ
 			if err := usecase.organizationSchemaRepository.CreateSchemaIfNotExists(ctx, orgTx); err != nil {
 				return err
 			}
-			return usecase.organizationSchemaRepository.CreateTable(ctx, orgTx, name)
+			if err := usecase.organizationSchemaRepository.CreateTable(ctx, orgTx, name); err != nil {
```
balzdur, line 125: It looks like in the whole data_model usecases file, we do not check for permission either, same question as above

repositories/ingested_data_indexes_repository.go
```diff
@@ -225,16 +225,130 @@ func withDesc(s models.FieldName) string {
...
+	return createUniqueIndex(ctx, exec, index, false)
+}
+
+func (repo *ClientDbRepository) DeleteUniqueIndex(ctx context.Context, exec Executor, index models.UnicityIndex) error {
+	if err := validateClientDbExecutor(exec); err != nil {
+		return err
+	}
+
+	indexName := toUniqIndexName(index.Fields, index.TableName)
+	fmt.Println(dropIdxSqlQuery(indexName, exec))
```
balzdur, line 351: probably to remvoe

models/data_model.go
```diff
@@ -81,6 +92,21 @@ type Table struct {
...
+	links := make(map[LinkName]LinkToSingle)
+	for k, v := range t.LinksToSingle {
+		links[k] = v
+	}
+	out := t
+	out.Fields = fields
+	out.LinksToSingle = links
+	return out
+}
+
```
balzdur, line 109: IMO, we should be "dumb" on those :
- rename `Copy()` to `DeepCopy()` (so it's clear we do not return a shallow copy)
- change this perticular function impelmentation to return a deep copy (fields + links)

By experience, using deep copy protect us against weird bug (for now, we avoid this kind of bug inside the integration using a reasign statement, instead of a mutation)

balzdur (approved):
LGTM, just a general question about usecases w\ any security check, is it intended ?

usecases/indexes/index_editor.go
```diff
@@ -135,3 +138,75 @@ func (editor ClientDbIndexEditor) ListAllUniqueIndexes(ctx context.Context) ([]m
...
+			err,
+			"Error while creating client schema executor in CreateUniqueIndexAsync")
+	}
+	err = editor.ingestedDataIndexesRepository.CreateUniqueIndexAsync(ctx, db, index)
+	if err != nil {
+		return errors.Wrap(err, "Error while creating unique index in CreateUniqueIndexAsync")
+	}
+	logger.InfoContext(
+		ctx,
+		fmt.Sprintf("Unique index pending creation asynchronously: %+v\n", index),
```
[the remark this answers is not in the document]
Pascal-Delange: Hmm I guess in the context of InfoContext we don't indeed (with the basic Printf as opposed to Println we do)

usecases/data_model_usecase.go
Pascal-Delange: 👍  true, I pushed it a bit far here

usecases/data_model_usecase.go
Pascal-Delange: I'm not sure to see what you mean here, every usecase method from dataModelRepository has a permission check (usually towards the beginning).

usecases/indexes/index_editor.go
```diff
@@ -135,3 +138,75 @@ func (editor ClientDbIndexEditor) ListAllUniqueIndexes(ctx context.Context) ([]m
...
+		return err
+	}
+	db, err := editor.executorFactory.NewClientDbExecutor(ctx, organizationId)
+	if err != nil {
+		return errors.Wrap(
+			err,
+			"Error while creating client schema executor in DeleteUniqueIndex")
+	}
+	return editor.ingestedDataIndexesRepository.DeleteUniqueIndex(ctx, db, index)
+}
```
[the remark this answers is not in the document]
Pascal-Delange: Hmm, I see your point. The idea here was that it can be used in different contexts. Similarly to how there is a scenario_fetcher usecase that is used in other usecases (never directly by the api handlers) and that does not enforce permissions. That plus the fact that in practice it's all quite specific to the data model usecase - I rather separated it from it in order to keep it unit-testable (if you have a zillion repository calls to mock to test a usecase, it doesn't work any more).
That being said, you could say there is a stronger case in favor or security checks inside, since it does actions not just reading.

models/data_model.go
Pascal-Delange: It's true that I did get a weird bug in my unit tests (didn't really reach a usecase error though) which is how I worked out that I need to have a deeper copy.
I'm not sure to understand what you mean by `change this perticular function impelmentation to return a deep copy (fields + links)` however, returning a deep copy if links & fields is precisely what I'm doing here

usecases/data_model_usecase.go
Pascal-Delange: actually no: you really can't modify a struct that's in a map - it's not about how we copy, we really have to reassign the struct in the map

Pascal-Delange: because a struct kind of feels like a map (with predefined keys) but it really isn't

PR #493, opened 2024-02-29
disposable: script for migrating unique indexes on cloud

Pascal-Delange:
This work should never be merged, I'm putting it up here only for sharing.
It's a (simple) script I wrote to manually run and create the missing indexes on staging & production tables.

I'm running it by putting the true db password in my env variable and connecting through the cloud sql proxy.

### DIFF
--- a/jobs/migrate_indexes.go
+++ b/jobs/migrate_indexes.go
@@ -1,0 +1,108 @@
+package jobs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/checkmarble/marble-backend/utils"
+)
+
+func generateUsecaseWithCredForMarbleAdmin(ctx context.Context, jobUsecases usecases.Usecases, orgId string) usecases.UsecasesWithCreds {
+	creds := models.Credentials{
+		Role:           models.MARBLE_ADMIN,
+		OrganizationId: orgId,
+	}
+	return usecases.UsecasesWithCreds{
+		Usecases:                jobUsecases,
+		Credentials:             creds,
+		Logger:                  utils.LoggerFromContext(ctx),
+		OrganizationIdOfContext: func() (string, error) { return orgId, nil },
+		Context:                 ctx,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


jobs/migrate_indexes.go
```diff
@@ -0,0 +1,108 @@
...
+					} else if done {
+						logger.InfoContext(ctx, "Added index on "+
+							string(field.Name)+" for table "+string(parentTable.Name))
+					}
+					// refresh data model so that the field.UnicityConstraint is up to date for the next iteration
+					dataModel, err = dmUsecase.GetDataModel(ctx, org.Id)
+					if err != nil {
+						logger.ErrorContext(ctx, "Error getting data model for org "+org.Id, err)
+						return err
+					}
```
balzdur, line 101: Here I understand the needs to do it inside the `for ... range` but this mind bug me ahah

jobs/migrate_indexes.go
```diff
@@ -0,0 +1,108 @@
...
+					return err
+				} else if done {
+					logger.InfoContext(ctx, "Added index on object_id for table "+string(table.Name))
+				}
+				// refresh data model so that the field.UnicityConstraint is up to date for the next iteration
+				dataModel, err = dmUsecase.GetDataModel(ctx, org.Id)
+				if err != nil {
+					logger.ErrorContext(ctx, "Error getting data model for org "+org.Id, err)
+					return err
+				}
```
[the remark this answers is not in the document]
Pascal-Delange: aha yes, my bad I copy pasted a bit from the other case below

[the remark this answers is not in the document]
Pascal-Delange: (as far as I can tell it doesn't actually make a difference though)

jobs/migrate_indexes.go
Pascal-Delange: yes I understand 🙈

Pascal-Delange: anyway in this specific case it'll work: we're iterating over the original map, which we never update (we just replace the struct which contains a pointer to that map).
Because I just use it to get the fields, it's fine.
It's of course a bit fucked up, but this being a one-time script, I can live with it

Pascal-Delange:
Closing this PR as the work is finished

PR #495, opened 2024-03-01
feat: automatically create case from decision

Pascal-Delange:
- New field `decision_to_case_inbox` and `decision_to_case_outcomes` on the scenario (models, dto, db).
 - If both are not nil and the decision's outcome is in decision_to_case_outcomes, then create a automatically
 - Permission check on inbox reading is bypassed: in API context, the inbox id on the scenario is the source of truth. 
 - Pending question (but the code is reviewable as it is): what level of permission to update those settings on the scenario -> Edit: publisher permission required to udpate the settings

Commit 5: [48c8313](https://github.com/checkmarble/marble-backend/pull/495/commits/48c831353629bf9bccb8c22de1a345f7013a4762): just cleaning up dtos & models, many line changes but no logic change

Related front PR: https://github.com/checkmarble/marble-frontend/pull/392

The idea for starters is to set it manually on our side for S&P, later we can add an interface of course.

### DIFF
--- a/dto/case_event_dto.go
+++ b/dto/case_event_dto.go
@@ -4,16 +4,17 @@ import (
 	"time"
 
 	"github.com/checkmarble/marble-backend/models"
+	"github.com/guregu/null/v5"
 )
 
 type APICaseEvent struct {
-	Id             string    `json:"id"`
-	CaseId         string    `json:"case_id"`
-	UserId         string    `json:"user_id"`
-	CreatedAt      time.Time `json:"created_at"`
-	EventType      string    `json:"event_type"`
-	AdditionalNote string    `json:"additional_note"`
-	NewValue       string    `json:"new_value"`
+	Id             string      `json:"id"`
+	CaseId         string      `json:"case_id"`
+	UserId         null.String `json:"user_id"`
+	CreatedAt      time.Time   `json:"created_at"`
+	EventType      string      `json:"event_type"`
+	AdditionalNote string      `json:"additional_note"`
+	NewValue       string      `json:"new_value"`
 }
 
 func NewAPICaseEvent(caseEvent models.CaseEvent) APICaseEvent {
--- a/dto/scenarios.go
+++ b/dto/scenarios.go
@@ -1,21 +1,60 @@
 package dto
 
 import (
+	"time"
+
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/pure_utils"
 	"github.com/guregu/null/v5"
 )
 
+// Read DTO
+type ScenarioDto struct {
+	Id                     string    `json:"id"`
+	CreatedAt              time.Time `json:"createdAt"`
+	DecisionToCaseOutcomes []string  `json:"decision_to_case_outcomes"`
+	DecisionToCaseInboxId  string    `json:"decision_to_case_inbox_id"`
+	Description            string    `json:"description"`
+	LiveVersionID          *string   `json:"liveVersionId,omitempty"`
+	Name                   string    `json:"name"`
+	OrganizationId         string    `json:"organization_id"`
+	TriggerObjectType      string    `json:"triggerObjectType"`
+}
+
+func AdaptScenarioDto(scenario models.Scenario) ScenarioDto {
+	out := ScenarioDto{
+		Id:        scenario.Id,
+		CreatedAt: scenario.CreatedAt,
+		DecisionToCaseOutcomes: pure_utils.Map(scenario.DecisionToCaseOutcomes,
+			func(o models.Outcome) string { return o.String() }),
+		Description:       scenario.Description,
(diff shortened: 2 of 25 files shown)


Pascal-Delange:
As of friday night remains to do: edit scenario workflow settings

usecases/case_usecase.go
```diff
@@ -153,59 +153,92 @@ func (usecase *CaseUseCase) GetCase(ctx context.Context, caseId string) (models.
...
+	createCaseAttributes models.CreateCaseAttributes,
+) (models.Case, error) {
+	// As an automated workflow, we don't need to check the inbox permissions
+	c, err := usecase.createCase(ctx, tx, "", createCaseAttributes, false)
+	if err != nil {
+		return models.Case{}, err
+	}
+
+	return c, err
+}
```
balzdur, line 240: Open question: do we want to track case creation from workflow ?

repositories/dbmodels/db_case_event.go
```diff
@@ -5,19 +5,20 @@ import (
...
-	CreatedAt      time.Time `db:"created_at"`
-	EventType      string    `db:"event_type"`
-	AdditionalNote *string   `db:"additional_note"`
-	ResourceId     *string   `db:"resource_id"`
-	ResourceType   *string   `db:"resource_type"`
-	NewValue       *string   `db:"new_value"`
-	PreviousValue  *string   `db:"previous_value"`
+	Id             string      `db:"id"`
+	CaseId         string      `db:"case_id"`
+	UserId         pgtype.Text `db:"user_id"`
```
balzdur, line 14: If we do save nillable user id, it would be great to adapt the event DTO to return nillable values too (and not `''`)

dto/scenarios.go
```diff
@@ -1,71 +1,76 @@
 package dto
 
 import (
+	"time"
+
 	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/guregu/null/v5"
```
balzdur, line 8: Why do you need this ? I don't find the place where it's used

balzdur (approved):

usecases/case_usecase.go
Pascal-Delange: Good question, I asked myself the same at first. I believe we should not - segment is about user analytics and their actions. We should (once we have the toggle) track users activating the workflow.
In fact, I don't even know how we would track it, because segment expects a user id (we don't have one in the context of machine to machine call)

dto/scenarios.go
Pascal-Delange: Line 60. Basically I want to be able to distinguish between `{"user_id": null}` and `{ }` as far as the parsed user_id is concerned. AKA send null to empty it, send nothing to ignore it in the PATCH

Pascal-Delange: Incidentally I don't need it for the other field (a slice) because the zero value from the parsed string slice is nil, which gives a nil slice in the model, which we ignore in the repository - whereas an empty slice in the input gives a non-nil (empty) slice, which we can use to reset the field to []

repositories/dbmodels/db_case_event.go
Pascal-Delange: Of course, that makes sense. I pushed another commit

dto/scenarios.go
Pascal-Delange: (it's basically the same as pgtype.text. As far as I can tell it's really the same thing and here I'm just choosing to use this rather than pgtype because the latter smells of "db", but same difference)

dto/case_event_dto.go
```diff
@@ -4,16 +4,17 @@ import (
...
-	Id             string    `json:"id"`
-	CaseId         string    `json:"case_id"`
-	UserId         string    `json:"user_id"`
-	CreatedAt      time.Time `json:"created_at"`
-	EventType      string    `json:"event_type"`
-	AdditionalNote string    `json:"additional_note"`
-	NewValue       string    `json:"new_value"`
+	Id             string      `json:"id"`
+	CaseId         string      `json:"case_id"`
+	UserId         null.String `json:"user_id"`
```
balzdur, line 13: IIUC, this is the change that make `user_id` optionnal in the returned JSON payload, is that right ?

balzdur (approved):

dto/case_event_dto.go
balzdur: (I ask because of [this line in the pkg readme](https://github.com/guregu/null?tab=readme-ov-file#bugs))

Pascal-Delange: Yes, that makes it so that a null input (aka value.Valid = false) will encode to JSON `null` while an empty string (`value.valid=true, value.String=""`) will encode to JSON `""`.
I could also get it to completely omit the user_id field in the output if empty string is received by marking `json:"user_id,omitempty"` but I like to have nulls present if they're in the schema

PR #496, opened 2024-03-04
introduce hash field for api keys

Pascal-Delange:
- introduce new field on api keys "hash", store it on creation and backfill it with the old keys hashed value
- for now continue using the non hashed key for matching
- in follow-up PR, use the hashed key for matching
- then finally, drop the key column (or reuse it to store just the prefix of the key, for display like `123##############`)
- while I'm at it, add a created_at field to return them in order

Security:
- we use 32 bytes of entropy to generate the keys, so no need to salt them
- SHA256 to hash the keys (industry standard)

### DIFF
--- a/models/api_key.go
+++ b/models/api_key.go
@@ -1,26 +1,19 @@
 package models
 
+import "time"
+
 type ApiKey struct {
 	Id             string
-	OrganizationId string
-	Hash           string
+	CreatedAt      time.Time
 	Description    string
+	Key            string
+	Hash           []byte
+	OrganizationId string
 	Role           Role
-}
-
-type CreatedApiKey struct {
-	ApiKey
-	Value string
 }
 
 type CreateApiKeyInput struct {
-	OrganizationId string
 	Description    string
+	OrganizationId string
 	Role           Role
-}
-
-type CreateApiKey struct {
-	CreateApiKeyInput
-	Id   string
-	Hash string
 }
--- a/repositories/migrations/20240304155400_hash_apikeys.sql
+++ b/repositories/migrations/20240304155400_hash_apikeys.sql
@@ -1,0 +1,24 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE apikeys
+ADD COLUMN key_hash bytea;
+
+ALTER TABLE apikeys
+ADD COLUMN created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now();
+
+UPDATE apikeys
+SET
+      key_hash = sha256(key::bytea)
+where
+      key_hash is null;
+
+-- +goose StatementEnd
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE apikeys
+DROP COLUMN key_hash;
+
+ALTER TABLE apikeys
+DROP COLUMN created_at;
+
+-- +goose StatementEnd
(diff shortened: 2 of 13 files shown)


repositories/migrations/20240304155400_hash_apikeys.sql
```diff
@@ -0,0 +1,24 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE apikeys
+ADD COLUMN key_hash bytea;
+
+ALTER TABLE apikeys
+ADD COLUMN created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now();
+
+UPDATE apikeys
```
Pascal-Delange, line 9: this is fine because there are few rows in the table

usecases/api_key_usecase.go
```diff
@@ -54,65 +54,51 @@ func (usecase *ApiKeyUseCase) ListApiKeys(ctx context.Context) ([]models.ApiKey,
...
-func (usecase *ApiKeyUseCase) CreateApiKey(ctx context.Context, input models.CreateApiKeyInput) (models.CreatedApiKey, error) {
-	apiKey, err := executor_factory.TransactionReturnValue(
+	if input.Role != models.API_CLIENT {
+		return models.ApiKey{}, errors.Wrap(
+			models.BadParameterError,
+			fmt.Sprintf("role %s is not supported", input.Role),
+		)
+	}
+
+	err := usecase.apiKeyRepository.CreateApiKey(
```
Pascal-Delange, line 81: untangling a bit of code here - we actually never need to read back the value from the DB, so no need for a transaction either

models/api_key.go
```diff
@@ -1,26 +1,19 @@
...
-}
-
 type CreateApiKeyInput struct {
-	OrganizationId string
 	Description    string
+	OrganizationId string
 	Role           Role
 }
-
-type CreateApiKey struct {
```
Pascal-Delange, line 22: I'm simplifying things a bit here - the CreateApiKey basically _is_ the ApiKey. 
In a follow-up PR I might still reintroduce a notion of "CreatedApiKey" at the model level, but it's easier to bring it back when I need it.

repositories/postgres/api_keys.go
```diff
@@ -33,8 +20,13 @@ func (db *Database) GetApiKeyByKey(ctx context.Context, key string) (models.ApiK
...
 	var apiKey dbmodels.DBApiKey
-	err := db.pool.QueryRow(ctx, query, key).Scan(&apiKey.Id, &apiKey.OrganizationId,
-		&apiKey.Hash, &apiKey.Description, &apiKey.Role)
+	err := db.pool.QueryRow(ctx, query, key).Scan(
+		&apiKey.Id,
+		&apiKey.OrganizationId,
+		&apiKey.Key,
+		&apiKey.Description,
+		&apiKey.Role,
+	)
```
balzdur, line 29: IIUC, for now we store the hash but since it's nneither used, we still do not need it here

Pascal-Delange: yes correct.
I could have done it in the same PR, but it's mergeable as it is so let's do it more simply in a follow-up PR

usecases/token/generator_test.go
```diff
@@ -18,7 +18,7 @@ func TestGenerator_GenerateToken_APIKey(t *testing.T) {
 	key := models.ApiKey{
 		Id:             "api_key_id",
 		OrganizationId: "organization_id",
-		Hash:           apiKeyHash,
+		Key:            apiKeyHash,
```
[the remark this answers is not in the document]
Pascal-Delange: ah yes, thanks

PR #497, opened 2024-03-05
fix error on db field read

Pascal-Delange:
Db field read:
- would fail with unexpected error if the child field was not present in the payload (e.g. account_id absent in transaction payload)
- would not work if the link was pointing to a parent table field other than object_id (would generate a wrong query)

### DIFF
--- a/repositories/ingested_data_read_repository.go
+++ b/repositories/ingested_data_read_repository.go
@@ -21,6 +21,7 @@ type IngestedDataReadRepository interface {
 
 type IngestedDataReadRepositoryImpl struct{}
 
+// "read db field" methods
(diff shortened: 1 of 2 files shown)


repositories/ingested_data_read_repository.go
```diff
@@ -55,10 +71,10 @@ func createQueryDbForField(ctx context.Context, exec Executor, readParams models
 			readParams.Path[0], models.NotFoundError)
 	}
 
-	firstTableObjectId, err := getFirstTableObjectIdFromPayload(readParams.ClientObject, link.ChildFieldName)
+	firstTableLinkValue, err := getParentTableJoinField(readParams.ClientObject, link.ChildFieldName)
```
Pascal-Delange, line 74: the function and the logic within was wrong, partly because it was incorrectly named

balzdur (approved):

PR #498, opened 2024-03-06
small changes for deployment with docker compose

Pascal-Delange:
- new env variable "enable_gc_tracing" (requires application default credentials to run)
- use alpine instead of distroless image to have access to wget to use in docker-compose health checks
- better documentation of what env variables do
- drop "VerifySessionCookie" logic in auth middleware, that was no longer supposed to be used but messed up all our error logging related to token validation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ FROM alpine:3.19
 
 COPY --from=build /go/bin/app /
 
-ENV API_PORT=${API_PORT:-8080}
-EXPOSE $API_PORT
+ENV PORT=${PORT:-8080}
+EXPOSE $PORT
 
 ENTRYPOINT ["/app"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ FROM alpine:3.19
 
 COPY --from=build /go/bin/app /
 
-ENV PORT=8080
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -7,11 +7,11 @@ RUN go get
 
 RUN CGO_ENABLED=0 go build -o /go/bin/app -ldflags="-X 'main.version=`git rev-parse --short HEAD`'"
 
-FROM gcr.io/distroless/static
```
balzdur, line 10: Same here, IIRC, this is prefarable for GCP to build from this image, so maybe we should have two separate dockerfile

balzdur (approved):

Dockerfile
Pascal-Delange: I don't know, as far as I understand it just makes for small images, but alpine is good enough.
And I want it for the health checks in docker-compose, where our users will use the same docker image as us

Pascal-Delange: Ok here are some ideas on how to do it better :https://www.mattknight.io/blog/docker-healthchecks-in-distroless-node-js

But let's keep that as a subject for later - for now, I can live with an alpine based image, it's small enough and surely not _that_ much of a security threat.

PR #502, opened 2024-03-11
tentative fix for timezone load

Pascal-Delange:
See
https://lalatron.hashnode.dev/a-story-about-go-docker-and-time-zones
https://stackoverflow.com/questions/59094236/error-unknown-time-zone-america-los-angeles-in-time-loadlocation

The error was introduced by using alpine as the base image, which does not include the tz definitions by default (turns out the tz info is handled by the environment not hard coded in the std lib)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ RUN CGO_ENABLED=0 go build -o /go/bin/app -ldflags="-X 'main.version=`git rev-pa
 FROM alpine:3.19
 
 COPY --from=build /go/bin/app /
+COPY --from=build /usr/local/go/lib/time/zoneinfo.zip /
+ENV ZONEINFO=/zoneinfo.zip
 
 ENV PORT=${PORT:-8080}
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,6 +10,8 @@ RUN CGO_ENABLED=0 go build -o /go/bin/app -ldflags="-X 'main.version=`git rev-pa
 FROM alpine:3.19
 
 COPY --from=build /go/bin/app /
+COPY --from=0 /usr/local/go/lib/time/zoneinfo.zip /
```
balzdur, line 13: IIUC, `--from=0` mean the same as `--from=build` in this context. I think `--from=build` is more explicit: it doesn't require to know about internal Docker convention.

balzdur (approved):

Dockerfile
Pascal-Delange: 👍  agreed

PR #506, opened 2024-03-12
Pascal/feat iteration id on decisions

Pascal-Delange:
Next step (but I'd rather do that in the next release) migrate to add the missing ids, leveraging the information we have

### DIFF
--- a/dto/decision_dto.go
+++ b/dto/decision_dto.go
@@ -37,6 +37,7 @@ type APIDecisionRule struct {
 	ScoreModifier int       `json:"score_modifier"`
 	Result        bool      `json:"result"`
 	Error         *APIError `json:"error"`
+	RuleId        string    `json:"rule_id"`
 }
 
 type APIError struct {
@@ -62,6 +63,7 @@ type APIDecision struct {
 	Rules                []APIDecisionRule   `json:"rules"`
 	Score                int                 `json:"score"`
 	ScheduledExecutionId *string             `json:"scheduled_execution_id,omitempty"`
+	ScenarioIterationId  string              `json:"scenario_iteration_id"`
 }
 
 func NewAPIDecision(decision models.Decision) APIDecision {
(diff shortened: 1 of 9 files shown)


dto/decision_dto.go
```diff
@@ -80,6 +82,7 @@ func NewAPIDecision(decision models.Decision) APIDecision {
 		Score:                decision.Score,
 		Rules:                make([]APIDecisionRule, len(decision.RuleExecutions)),
 		ScheduledExecutionId: decision.ScheduledExecutionId,
+		ScenarioIterationId:  decision.ScenarioIterationId,
```
balzdur, line 85: Just in case: there is `Version` info in the `Scenario` property. It may be better to flatten everything or group everything or 🤷🏽 

I do not have strong opinion but I think grouping `Version` + `ScenarioIterationId` at the same place (or droping `Version` make sense)

Having a copy of Scenario `Name` + `Description` can make sense since they are mutable

repositories/dbmodels/db_decision.go
```diff
@@ -23,6 +23,7 @@ type DbDecision struct {
 	CreatedAt            time.Time   `db:"created_at"`
 	Outcome              string      `db:"outcome"`
 	ScenarioId           string      `db:"scenario_id"`
+	ScenarioIterationId  pgtype.Text `db:"scenario_iteration_id"`
```
balzdur, line 26: IIUC, this should be `string` as soon as we have backfilled existing decision

dto/decision_dto.go
```diff
@@ -62,6 +63,7 @@ type APIDecision struct {
 	Rules                []APIDecisionRule   `json:"rules"`
 	Score                int                 `json:"score"`
 	ScheduledExecutionId *string             `json:"scheduled_execution_id,omitempty"`
+	ScenarioIterationId  string              `json:"scenario_iteration_id"`
```
balzdur, line 66: I would rather do this until we have backfilled existing values
```suggestion
	ScenarioIterationId  string              `json:"scenario_iteration_id,omitempty"`
```

balzdur (approved):

dto/decision_dto.go
Pascal-Delange: ah yes, of course 👍

repositories/dbmodels/db_decision.go
Pascal-Delange: yes

dto/decision_dto.go
Pascal-Delange: ok, can do 👍

PR #512, opened 2024-03-14
hotfix: table name case was lost at creation, failing field creation …

Pascal-Delange:
…if table name was not lower case


Pascal-Delange:
To merge in main, then I'll cherry pick to hotfix prod

PR #515, opened 2024-03-14
feat(api): add decision rule ast to decision response

balzdur:
When getting a decision by id, populate rule asts

EDIT: I needed to add `score_modifier` since on the DTO we expose the "result" score as `score_modifier` (I need the real score modif of the rule for detail display)

What I really need here, is a validation of the direction I took. I'm not sure the way I build struct is the good one... Still, I choose to create `rule_detail` in order to avoid breaking change on the API (IIUC, this endpoint is on the public API inside the readme, we may need to update it)

### DIFF
--- a/api/handle_decision.go
+++ b/api/handle_decision.go
@@ -29,7 +29,7 @@ func (api *API) handleGetDecision(c *gin.Context) {
 	if presentError(c, err) {
 		return
 	}
-	c.JSON(http.StatusOK, dto.NewAPIDecision(decision))
+	c.JSON(http.StatusOK, dto.NewAPIDecisionWithRule(decision))
 }
 
 func (api *API) handleListDecisions(c *gin.Context) {
@@ -141,5 +141,5 @@ func (api *API) handlePostDecision(c *gin.Context) {
 		presentError(c, errors.Wrap(err, "Error creating decision in handlePostDecision"))
 		return
 	}
-	c.JSON(http.StatusOK, dto.NewAPIDecision(decision))
+	c.JSON(http.StatusOK, dto.NewAPIDecisionWithRule(decision))
 }
--- a/dto/decision_dto.go
+++ b/dto/decision_dto.go
@@ -61,9 +61,13 @@ type APIDecision struct {
 	TriggerObjectType    string              `json:"trigger_object_type"`
 	Outcome              string              `json:"outcome"`
 	Scenario             APIDecisionScenario `json:"scenario"`
-	Rules                []APIDecisionRule   `json:"rules"`
 	Score                int                 `json:"score"`
 	ScheduledExecutionId *string             `json:"scheduled_execution_id,omitempty"`
+}
+
+type APIDecisionWithRules struct {
+	APIDecision
+	Rules []APIDecisionRule `json:"rules"`
 }
 
 func NewAPIDecision(decision models.Decision) APIDecision {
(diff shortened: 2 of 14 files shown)


repositories/decisions_repository.go
```diff
@@ -434,15 +434,22 @@ func (repo *DecisionRepositoryImpl) rulesOfDecision(ctx context.Context, exec Ex
...
+	var columns []string
+	columns = append(columns, columnsNames("dr", dbmodels.SelectDecisionRuleColumn)...)
+	columns = append(columns, "r.formula_ast_expression")
+	query := NewQueryBuilder().
+		Select(columns...).
+		From(fmt.Sprintf("%s AS dr", dbmodels.TABLE_DECISION_RULE)).
+		LeftJoin(fmt.Sprintf("%s AS r ON r.id = dr.rule_id", dbmodels.TABLE_RULES)).
+		Where(squirrel.Eq{"dr.decision_id": decisionId}).
+		OrderBy("dr.id")
+
```
balzdur, line 446: TBH on this: inside the `DecisionById` impl, we do :
- get all rules exec (this is the code above)
- get the decision + join on case (if any)

So in the end, we will do 2 joins. Still this is two separated querries (one for rule exec, one for decision)

repositories/decisions_repository.go
```diff
@@ -434,15 +434,27 @@ func (repo *DecisionRepositoryImpl) rulesOfDecision(ctx context.Context, exec Ex
...
 		exec,
-		NewQueryBuilder().Select(dbmodels.SelectDecisionRuleColumn...).
-			From(dbmodels.TABLE_DECISION_RULE).
-			Where(squirrel.Eq{"decision_id": decisionId}).
-			OrderBy("id"),
-		func(r dbmodels.DbDecisionRule) (models.RuleExecution, error) {
-			return dbmodels.AdaptRuleExecution(r), nil
+		query,
+		func(row pgx.CollectableRow) (models.RuleExecution, error) {
+			db, err := pgx.RowToStructByPos[dbmodels.DbDecisionRuleWithDetail](row)
```
Pascal-Delange, line 452: I think I'm a bit troubled by this utility function, in this context.
I fail to understand how it reconciles tables and fields. You have prefixes on fields and table names in the query (that's good), but the prefixes are of course not present on the (composed) struct attribute names. Especially here where the two joined tables share some field name score_modifier (though not the nested structs).

So by default (assuming you tested it and it worked locally) it does just take them in order, and understands that the first one goes to the DbDecisionRule and the second to the DbDecisionRuleDetail. (what happens if you invert the order in the columns by switching lines 438 and 439 btw ?)

Anyway, this is where pgx.RowToStructByPos becomes hard to follow and we're perhaps better off by explicitly scanning into the struct.

api/handle_decision.go
```diff
@@ -29,7 +29,12 @@ func (api *API) handleGetDecision(c *gin.Context) {
 	if presentError(c, err) {
 		return
 	}
-	c.JSON(http.StatusOK, dto.NewAPIDecision(decision))
+	decisionWithFormulaDto, err := dto.NewAPIDecisionWithDetail(decision)
```
Pascal-Delange, line 32: ok so hold on.
This may be getting a bit confusing:
1. get decision by id returns rules with details
2. list decisions does not return rules
3. create decision returns rules but not rule detail

It's fine that the first two are different, it's harder to explain that 1 and 3 are.
But then, as you said, we need to modify the create decisions endpoint (which is a public endpoint).


Now if I understand your PR correctly, we are here replacing the "actual" executed rule score modifier (0 if not trigger) by the "raw" value, which is an API change.
It's probably still early enough to make a breaking change on the public API - I'm comfortable with that - but I'm less sure if that is specifically a change we want to make. Let's discuss it when you're back on Wed

dto/decision_dto.go
```diff
@@ -40,6 +40,14 @@ type APIDecisionRule struct {
 	RuleId        string    `json:"rule_id"`
 }
 
+type APIDecisionRuleWithDetail struct {
+	APIDecisionRule
+	RuleDetail struct {
+		FormulaAstExpression *NodeDto `json:"formula_ast_expression"`
+		ScoreModifier        int      `json:"score_modifier"`
```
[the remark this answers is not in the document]
Pascal-Delange: I mean, ok I get why it's there but I don't quite get how it works in this case.
You compose structs, so which of the two values does it take to fill in the json score_modifier field ? The second one just because it comes second ?

api/handle_decision.go
balzdur: Yes you're right. An alternative (but I'm not sure yet how it may impact the front) would be to fetch sequentially :
- the decision (the one previous to this PR)
- the rule using the `rule_id`

It may be simpler to understand and if it impact perf, I can fetch rule detail without blocking the navigation (so loading the decision detail won't be affected by chaining requests) as we only need detail when openning rule detail

Pascal-Delange: No that's fine, I think with some minor tweaks to your code we can proceed as it is.
The basis of the misunderstanding is that we're using a field with the same name from 2 different tables we join, with a subtly different meaning. The way to go is to clarify what our API exposes, and from there we can work back to the sql query.
If you want we can do it together on Wed.

Pascal-Delange (approved):

PR #517, opened 2024-03-18
fix allowed origin for cors outside of dev mode

Pascal-Delange:
That is the correct setting, right ?

### DIFF
--- a/router.go
+++ b/router.go
@@ -16,11 +16,11 @@ import (
 )
 
 func corsOption(env string) cors.Config {
-	allowedOrigins := []string{"*"}
+	allowedOrigins := []string{"*.checkmarble.com"}
 
 	if env == "development" {
-		allowedOrigins = append(allowedOrigins, "http://localhost:3000",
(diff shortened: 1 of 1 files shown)


router.go
```diff
@@ -16,11 +16,11 @@ import (
...
 
 func corsOption(env string) cors.Config {
-	allowedOrigins := []string{"*"}
+	allowedOrigins := []string{"*.checkmarble.com"}
 
 	if env == "development" {
-		allowedOrigins = append(allowedOrigins, "http://localhost:3000",
-			"http://localhost:3001", "http://localhost:3002", "http://localhost:5173")
+		allowedOrigins = append(allowedOrigins,
+			"http://localhost:3000", "http://localhost:3001", "http://localhost:3002", "http://localhost:5173")
```
Pascal-Delange, line 23: do you have any idea why we allow 5173 ?

balzdur: No idea 🤷🏽

balzdur (approved):

PR #518, opened 2024-03-18
fix: no 500 error on admin user login

Pascal-Delange:
To do perhaps: handle the backend's 400 error more gracefully on the front (currently: stays stuck on an error page)


balzdur (approved):
I created [this](https://linear.app/checkmarble/issue/MAR-507/gracefully-handle-marble-admin-user) Linear issue in order to track the improvment for frontend

PR #521, opened 2024-03-21
Pascal/transfercheck

Pascal-Delange:
To do:

- [x] Create the basic API as specified
- [x] Usecases that store, query, update, score the transfers, using our ingested data repo as a storage layer
- [x] Create marble ids for the transfers, store them in a mapping table (marble `transfer_id` <=> ingested `transfer_id` AKA `object_id`)
- [x] Lookup the latest decision's score (for the transfer) on GET/PUT endpoints
- [x] Update openapi doc
- [x] Lookup the transfercheck scenario from the organization (& store it on the organization)
- [x] Index json fields in decisions (& use jsonb instead of json type)
- [x] Documentation on how to create the transfercheck org

Will do in a follow-up PR:
- [ ] Multi-tenant org (with partner_id on keys)
- [ ] Handle permissions with the key's partner_id
- [ ] Add the necessary new operators (string fuzzy matching...)
- [ ] Check that the API does not expose the details (AST) of our scenario to users with API keys

Link [openapi doc](https://www.notion.so/checkmarble/Draft-openapi-d3a71ab6212f4613a3f6362ec5e1a074)

The [org setup guide](https://www.notion.so/checkmarble/Org-setup-guide-984e368f113e434f97f42aaa6424e60c) on notion

Notable points:
- we ingest first and then score. If for whatever reason an error occurs during scoring, the ingested row will remain
- we first insert the "transfer mapping" (which does the link between our generated id and the client transfer id), then we ingest the object. So there can be orphan transfer mappings. However, reads check that both sides are present so this should be transparent for the users


repositories/ingested_data_read_repository.go
```diff
@@ -15,8 +15,20 @@ import (
...
 	ListAllObjectsFromTable(ctx context.Context, exec Executor, table models.Table) ([]models.ClientObject, error)
-	QueryAggregatedValue(ctx context.Context, exec Executor, tableName models.TableName,
-		fieldName models.FieldName, aggregator ast.Aggregator, filters []ast.Filter) (any, error)
+	QueryIngestedObject(
+		ctx context.Context,
+		exec Executor,
+		table models.Table,
+		objectId string,
+	) ([]map[string]any, error)
+	QueryAggregatedValue(
```
balzdur, line 24: Why do we use `Query*` here instead ?

balzdur: Ok probably just because `QueryAggregatedValue` already exists ^^'

balzdur (approved):

Pascal-Delange:
@balzdur 
Following the discussion on Friday, in the latest commit ([f44241f](https://github.com/checkmarble/marble-backend/pull/521/commits/f44241f391bdf9ba7caaded84ab47cf6cb57856b)):
- store the marble_id in the ingested db (that's cheap)
- try to reduce the risk of "orphan" transfer_mapping rows by trying to delete them if ingestion fails (but catch deletion errors, so the possibility of orphan rows is still there
- in hindsight, I didn't want to wrap the two in transactions, because while removing the risk of orphan rows in transfer_mappings it would have created a risk for orphan ingested objects, which is worse

I'm keeping the transfer_mappings table because at the cost of a bit more code/tables, it makes it easier to 
- query them and handle the required indexes explicitly
- use it to store the partner_id of a transfer explicitly
- use it in metabase analytics to query anonymous stats on transferchecks usage more easily

PR #522, opened 2024-03-22
feat(decision): store evaluation of decision rules

balzdur:
Store and return node evaluation info inside rule execution


models/ast/ast_node_evaluation.go
```diff
@@ -35,24 +35,11 @@ func (root NodeEvaluation) AllErrors() (errs []error) {
...
 
-type RootNodeEvaluation struct {
-	ReturnValue bool
-	Errors      []error
-
-	Children      []NodeEvaluation
-	NamedChildren map[string]NodeEvaluation
-}
-
-func AdaptRootNodeEvaluation(root NodeEvaluation) (RootNodeEvaluation, error) {
```
balzdur, line 46: Roll back because it's anoying to use inside recursive function (need to handle recurisve func for `NodeEvaluation` due to chil/named chil and a "root" one for `RootNodeEvaluation`)

models/decision.go
```diff
@@ -46,22 +48,23 @@ type ScenarioExecution struct {
 type RuleExecution struct {
 	Rule                Rule
 	Result              bool
+	Evaluation          *ast.NodeEvaluation
 	ResultScoreModifier int
 	Error               error
 }
 
-type RuleExecutionError int
+type ExecutionError int
```
balzdur, line 56: Rename since it's not rule related

usecases/ast_eval/evaluate_ast_expression.go
```diff
@@ -29,12 +29,12 @@ func (evaluator *EvaluateAstExpression) EvaluateAstExpression(
...
 	}
 
-	rootEvaluation, err := ast.AdaptRootNodeEvaluation(evaluation)
+	returnValue, err := evaluation.GetBoolReturnValue()
 	if err != nil {
-		return ast.RootNodeEvaluation{}, errors.Join(models.ErrRuntimeExpression, err)
+		return false, evaluation, errors.Join(models.ErrRuntimeExpression, err)
 	}
-	return rootEvaluation, nil
+	return returnValue, evaluation, nil
```
balzdur, line 39: Now return `(bool, ast.NodeEvaluation, error)` where.

`evaluation` is always returned (so we can store intermediate calc info to get context on errors)

repositories/dbmodels/db_ast_evaluation.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := json.Unmarshal(serializedNodeEvaluation, &dbNodeEvaluation); err != nil {
+		return nil, err
+	}
+
+	nodeEvaluation, err := adaptNodeEvaluation(dbNodeEvaluation)
+	if err != nil {
+		return nil, err
+	}
+	return &nodeEvaluation, nil
+}
```
balzdur, line 85: Inspired from ast de/ser

balzdur (commented):
It works as it is, still we have some particular cases to handle...

Ex: when evaluating a list, the whole list is loaded in memory, making the `return_value` really huge, we probably do not want to store this in the decision rule db

Pascal-Delange (commented):
A few questions !

repositories/dbmodels/db_ast_evaluation.go
```diff
@@ -0,0 +1,85 @@
+package dbmodels
```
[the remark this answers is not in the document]
balzdur: This is one of my brain freeze from last week.

IF you look at the code, there is already a model really close to `dbNodeEvaluation` : `NodeEvaluationDto`. Thet both serve the same purpose : adapting from `ast.NodeEvaluation`.

Still there is some problems :
- `NodeEvaluationDto` use `EvaluationErrorDto` which do not include `models.ExecutionError` internally (could be fixed by handling them)
- `NodeEvaluationDto` only adapt from `ast.NodeEvaluation` (no adapters yet for the other way required to load from the DB if used for the de/ser): I may look into it, maybe it's feasable

repositories/dbmodels/db_ast_evaluation.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/models/ast"
+	"github.com/checkmarble/marble-backend/pure_utils"
+)
+
+type dbNodeEvaluation struct {
+	ReturnValue   any                         `json:"return_value"`
+	ErrorCodes    []models.ExecutionError     `json:"error_codes"`
```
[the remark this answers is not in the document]
balzdur: Yes, only error codes like those go inside this prop.
This private struct is only created/used to store/unstore from the DB, we currently use the `NodeEvaluationDto` model for API (so no error codes are sent to the frontend for now).
Cf other comment for more information

repositories/dbmodels/db_ast_evaluation.go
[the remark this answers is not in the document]
Pascal-Delange: from a quick look, it seems that when we serialize the AST for rules (for storage in db), we use the same DTO & adapters as when we return them from the API (cf `dbmodels/db_ast_expression.go`).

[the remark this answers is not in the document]
Pascal-Delange: It looks like it would be good to do something similar here - we can have a look together whenever you're back today

repositories/migrations/20240322130353_evaluation_on_decisionrules.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
```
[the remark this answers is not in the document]
Pascal-Delange: Je vais merge ma PR staging, tu devras renommer la migration pour mettre un timestamp plus tard donc (la mienne a un timestamp 1h plus tard en l'état)

repositories/dbmodels/db_ast_evaluation.go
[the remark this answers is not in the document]
balzdur: Reworked based on our discussion

models/decision.go
```diff
@@ -53,22 +55,23 @@ type ScenarioExecution struct {
...
-	NoRowsRead           RuleExecutionError = 201
-	PayloadFieldNotFound RuleExecutionError = 202
-	Unknown              RuleExecutionError = -1
+	NoError              ExecutionError = 0
+	DivisionByZero       ExecutionError = 100
+	NullFieldRead        ExecutionError = 200
+	NoRowsRead           ExecutionError = 201
+	PayloadFieldNotFound ExecutionError = 202
+	Unknown              ExecutionError = -1
 )
```
balzdur, line 72: Not yet handled in this PR

repositories/migrations/20240325142444_evaluation_on_decisionrules.sql
```diff
@@ -0,0 +1,12 @@
...
+ALTER TABLE decision_rules
+ADD COLUMN rule_evaluation jsonb;
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE decision_rules
+DROP COLUMN rule_evaluation;
+-- +goose StatementEnd
```
balzdur, line 12: Moved to a new named migration (after rebasing on `main`) so it prevent issues

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -185,9 +185,11 @@ func evalScenarioRule(
 			fmt.Sprintf("error while evaluating rule %s (%s)", rule.Name, rule.Id))
 	}
 
+	ruleEvaluationDto := ast.AdaptNodeEvaluationDto(ruleEvaluation)
 	ruleExecution := models.RuleExecution{
-		Rule:   rule,
-		Result: ruleReturnValue.ReturnValue,
+		Rule:       rule,
+		Evaluation: &ruleEvaluationDto,
+		Result:     returnValue,
```
balzdur, line 192: Adapt to `NodeEvaluationDto` so we manipulate the same `models.RuleExecution` everywhere

balzdur (commented):
Last changes are based on our previous discussion.

Handling specific case like "List" is not done in those changes

PR #524, opened 2024-03-25
drop clear api keys, store their prefix

Pascal-Delange:
- store the api key prefix (instead of its clear value)
- expose it to the front


repositories/migrations/20240325133100_store_api_key_prefix.sql
```diff
@@ -0,0 +1,32 @@
...
+-- +goose StatementBegin
+ALTER TABLE api_keys
+RENAME TO apikeys;
+
+ALTER TABLE apikeys
+RENAME COLUMN prefix TO key;
+
+CREATE UNIQUE INDEX apikey_key_idx ON apikeys (prefixkey);
+
+-- +goose StatementEnd
```
balzdur, line 32: Same warning, depending on who merge first, we may need to change the filename

usecases/token/generator.go
```diff
@@ -60,7 +60,7 @@ func (g *Generator) fromAPIKey(ctx context.Context, apiKey string) (string, time
 			fmt.Errorf("GetOrganizationByID error: %w", err)
 	}
 
-	name := fmt.Sprintf("ApiKey Of %s", organization.Name)
+	name := fmt.Sprintf("Api key %s*** of %s", key.Prefix, organization.Name)
```
balzdur, line 63: IIUC, the Prefix is used by convenience :
- to track which api key is used
- to communicate with a client, in case we need to investigate issues about a perticular apikey (= impossible to use postgress ID for that)

Is the name prop used to create event ?

balzdur (approved):

repositories/migrations/20240325133100_store_api_key_prefix.sql
Pascal-Delange: yep

PR #525, opened 2024-03-25
add partners entity - link it (optional) to api keys

Pascal-Delange:
## Left over from previous PR
- [x] Multi-tenant org (with partner_id on keys)
- [x] Handle permissions with the key's partner_id
- [x] Check that the API does not expose the details (AST) of our scenario to users with API keys
- [x] Use partner_id as prefix in the ingested object_id to ensure clients do not read or overwrite each other's objects

------
Introduce a new permission type for Transfercheck API key. The difference with regular API key is that it does not have the permission to read a scenario. We want this because our proprietary algo (not open source) for transfercheck is stored in the DB, hence in the scenario.

------ 

## Illustration
**Postman calls**
- Post transfer:
Capture d’écran 2024-03-26 à 14 07 13

- Query transfers by (client) id: (notice the output of size 1)
Capture d’écran 2024-03-26 à 14 07 05

**Db extract**:
Capture d’écran 2024-03-26 à 14 08 10


usecases/decision_usecase.go
```diff
@@ -188,8 +189,10 @@ func (usecase *DecisionUsecase) CreateDecision(
 		return models.DecisionWithRuleExecutions{},
 			errors.Wrap(err, "error getting scenario")
 	}
-	if err := usecase.enforceSecurityScenario.ReadScenario(scenario); err != nil {
-		return models.DecisionWithRuleExecutions{}, err
+	if !skipScenarioPermissionForTransferCheck {
```
Pascal-Delange, line 192: I'm not all too happy with that, but it's a fair enough workaround for right now.
The reason I'm introducing this is that I don't want to give scenario_read permissions to the API keys used for transfer check (because we don't want clients to read the logic details we put in the rules).
A long term solution would be to factorize code differently here... but now I need to move on to the string fuzzy matching with higher priority.

models/role_permission.go
```diff
@@ -37,6 +37,9 @@ var ROLES_PERMISSIOMS = map[Role][]Permission{
 		SCENARIO_READ, SCENARIO_CREATE, DECISION_READ, DECISION_CREATE, INGESTION,
 		DATA_MODEL_READ, CUSTOM_LISTS_READ, CUSTOM_LISTS_CREATE,
 	},
+	TRANSFER_CHECK_API_CLIENT: {
```
Pascal-Delange, line 40: Cf the other comment: I want distinct (& more restricted) permissions for the transfer check api than for regular decisions

usecases/transfer_check_usecase.go
```diff
@@ -87,16 +113,29 @@ func (usecase *TransferCheckUsecase) CreateTransfer(
...
 		if err != nil {
 			return models.Transfer{}, err
 		}
+		transferMappings, err = usecase.transferMappingsRepository.ListTransferMappings(
+			ctx,
+			exec,
+			organizationId,
+			partnerId,
+			transfer.TransferData.TransferId,
+		)
```
balzdur, line 129: In this perticular case, you can directly use `GetTransferMapping` since you generate `transferMappingId = uuid.New().String()`

balzdur (approved):

models/transfer_check.go
```diff
@@ -145,16 +148,16 @@ type TransferCreateBody struct {
 	SkipScore    *bool
 }
 
-func (t TransferDataCreateBody) ToIngestionMap(id string) map[string]any {
+func (t TransferDataCreateBody) ToIngestionMap(mapping TransferMapping) map[string]any {
 	return map[string]any{
 		// there is a trap here: we map it to "object_id" to match what we do elsewhere on data model tables
-		"object_id": t.TransferId,
+		"object_id": ObjectIdWithPartnerIdPrefix(mapping.PartnerId, t.TransferId),
```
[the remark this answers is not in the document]
Pascal-Delange: because the object_id is the unique key for the purpose of identifying ingested objects in the org, and we want to guarantee partners cannot overwrite/read each other's ingested objects.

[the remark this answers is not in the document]
Pascal-Delange: so we must make object_id unique per partner in our code

usecases/decision_usecase.go
Pascal-Delange: I'm not sure that works, the permission is held on the credentials which are in the context, at first look this seems unpractical (the usecase with its context is built at the API handler level)

usecases/transfer_check_usecase.go
Pascal-Delange: ah yes, thank you

usecases/transfer_check_usecase.go
```diff
@@ -87,16 +113,29 @@ func (usecase *TransferCheckUsecase) CreateTransfer(
...
+			return models.Transfer{}, err
+		}
 	}
 
-	clientObject := models.ClientObject{Data: transfer.TransferData.ToIngestionMap(
-		transferMappingId), TableName: TransferCheckTable}
+	clientObject := models.ClientObject{
+		Data:      transfer.TransferData.ToIngestionMap(transferMappings[0]),
+		TableName: TransferCheckTable,
+	}
```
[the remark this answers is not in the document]
Pascal-Delange: well not really since if it not present, in the else branch I insert it.

usecases/decision_usecase.go
Pascal-Delange: (and TBH I'm not 100% with how we handle that context injection, but we'll see about that issue in due time)

PR #526, opened 2024-03-25
Thomas/handle runtime errors

balzdur:
Handle runtime errors :
- move definition to `ast` pkg
- handle those errors in `EvaluationErrorDto`
- refactor legacy decision rule error handling

### DIFF
--- a/models/ast/ast_node_evaluation.go
+++ b/models/ast/ast_node_evaluation.go
@@ -14,24 +14,19 @@ type NodeEvaluation struct {
 	NamedChildren map[string]NodeEvaluation
 }
 
-func (root NodeEvaluation) AllErrors() (errs []error) {
-	var addEvaluationErrors func(NodeEvaluation)
+func (root NodeEvaluation) FlattenErrors() []error {
+	errs := make([]error, 0)
 
-	addEvaluationErrors = func(child NodeEvaluation) {
-		if child.Errors != nil {
-			errs = append(errs, child.Errors...)
-		}
+	errs = append(errs, root.Errors...)
 
-		for _, child := range child.Children {
-			addEvaluationErrors(child)
-		}
-
-		for _, child := range child.NamedChildren {
-			addEvaluationErrors(child)
-		}
+	for _, child := range root.Children {
+		errs = append(errs, child.FlattenErrors()...)
 	}
 
-	addEvaluationErrors(root)
+	for _, child := range root.NamedChildren {
+		errs = append(errs, child.FlattenErrors()...)
+	}
+
 	return errs
 }
 
--- a/models/ast/ast_node_evaluation.go
+++ b/models/ast/ast_node_evaluation.go
@@ -7,6 +7,7 @@ import (
 )
 
 type NodeEvaluation struct {
+	Function    Function
 	ReturnValue any
 	Errors      []error
 
--- a/models/decision.go
+++ b/models/decision.go
@@ -91,13 +91,13 @@ func AdaptExecutionError(err error) ExecutionError {
 	switch {
 	case err == nil:
 		return NoError
-	case errors.Is(err, ErrNullFieldRead):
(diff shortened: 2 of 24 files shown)


models/decision.go
```diff
@@ -59,47 +58,3 @@ type RuleExecution struct {
...
-	case errors.Is(err, ErrNoRowsRead):
-		return NoRowsRead
-	case errors.Is(err, ErrDivisionByZero):
-		return DivisionByZero
-	case errors.Is(err, ErrPayloadFieldNotFound):
-		return PayloadFieldNotFound
-	default:
-		return Unknown
-	}
-}
```
balzdur, line 105: Moved to `dbmodels`: this is only a storage implementation detail ATM

repositories/dbmodels/db_decision_rule.go
```diff
@@ -1,44 +1,90 @@
...
+type ExecutionError int
+
+const (
+	NoError              ExecutionError = 0
+	DivisionByZero       ExecutionError = 100
+	NullFieldRead        ExecutionError = 200
+	NoRowsRead           ExecutionError = 201
+	PayloadFieldNotFound ExecutionError = 202
+	Unknown              ExecutionError = -1
+)
```
balzdur, line 41: Some context on the current code :
- this error is a "single error" attached to each rule evaluation
- the error is derived from the flatten rule evaluation errors (the logic which choose the single error is defined in `AdaptExecutionError` above)
- errors codes are only used for storage ATM (artificially rebuild the error on READ using `adaptErrorCodeAsError`)

This could be handled totally differently :
1. **remove**: we do store every nested errors in the rule execution now, so this "root" error is a sugar prop
2. **kept**: we could adapt `error` to `EvaluationErrorDto` as we do for other evaluation errors

Why I did not choose one of the above solutions : 
- this error "root" prop is documented in [our public API](https://docs.checkmarble.com/reference/ressource-definitions#error):
  - in practice, we always return `code: 1` so error codes are not really respecting the doc
  - explanation of error code `100` looks wrong (comparing public API vs our codebase)
- on existing decisions, this is the only valuable stored information on decision Errors

What we could do :
- Choose the 2. solution (since we do not respect public API error codes, it's not really a problem to break IMO)
- Migrate existing decisions to store equivalent `EvaluationErrorDto` (so we do not loose information on past decisions)

models/ast/ast_node_evaluation.go
```diff
@@ -14,24 +14,19 @@ type NodeEvaluation struct {
 	NamedChildren map[string]NodeEvaluation
 }
 
-func (root NodeEvaluation) AllErrors() (errs []error) {
-	var addEvaluationErrors func(NodeEvaluation)
+func (root NodeEvaluation) FlattenErrors() []error {
```
balzdur, line 17: Rename + refactor so it's clearer (at least to me 😅)

repositories/dbmodels/db_decision_rule.go
Pascal-Delange: > in practice, we always return code: 1 so error codes are not really respecting the doc

I don't think that's the case, the error code is properly stored on rule executions (however there is an error code on the decision level which may be always at 1)

Pascal-Delange: > explanation of error code 100 looks wrong (comparing public API vs our codebase)

is it ? 🤔

Pascal-Delange (approved):

repositories/dbmodels/db_decision_rule.go
balzdur: Taken from the documentation :
code | message
-- | --
100 | Scenario was not able to compute a score because all rules failed.

What is returned for code 100 in our code base :
code | message
-- | --
1 | Division by zero (alike message)

-> All rules did not failed, the error is specific to Division by zero

PR #527, opened 2024-03-25
Thomas/skip long results

balzdur:
Add support for skipping long result (DTO + storage)

### DIFF
--- a/models/ast/node_evaluation_dto.go
+++ b/models/ast/node_evaluation_dto.go
@@ -2,31 +2,35 @@ package ast
 
 import "github.com/checkmarble/marble-backend/pure_utils"
 
+// When too long, the ReturnValue can be omitted (ex: when function is `FUNC_CUSTOM_LIST_ACCESS“).
+// In such cases, ReturnValue will be nil and Omitted will be true.
+type ReturnValueDto struct {
+	Value   any  `json:"value,omitempty"`
+	Omitted bool `json:"is_omitted"`
+}
+
 type NodeEvaluationDto struct {
-	// When too long, the ReturnValue can be omitted (ex: when function is `FUNC_CUSTOM_LIST_ACCESS``).
-	// In such cases, ReturnValue will be the constant `omittedReturnValue``
-	ReturnValue   any                          `json:"return_value"`
+	ReturnValue   ReturnValueDto               `json:"return_value"`
 	Errors        []EvaluationErrorDto         `json:"errors"`
 	Children      []NodeEvaluationDto          `json:"children,omitempty"`
 	NamedChildren map[string]NodeEvaluationDto `json:"named_children,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


models/ast/node_evaluation_dto.go
```diff
@@ -10,10 +13,22 @@ type NodeEvaluationDto struct {
...
+		nodeEvaluationDto.ReturnValue = omittedReturnValue
+	}
+
+	return nodeEvaluationDto
+}
+
+const omittedReturnValue = "__omitted_return_value"
+
+func isReturnValueOmitted(evaluation NodeEvaluation) bool {
+	return evaluation.Function == FUNC_CUSTOM_LIST_ACCESS
```
balzdur, line 33: I'm open to discussion on this. What I need, is an "explicit" way to discover omitted result in the frontend (and I prefer a rule that is not based on business logic like the one implemented in `isReturnValueOmitted`).

models/ast/node_evaluation_dto.go
```diff
@@ -3,17 +3,31 @@ package ast
...
-	return NodeEvaluationDto{
+	nodeEvaluationDto := NodeEvaluationDto{
 		ReturnValue:   evaluation.ReturnValue,
 		Errors:        pure_utils.Map(evaluation.Errors, AdaptEvaluationErrorDto),
 		Children:      pure_utils.Map(evaluation.Children, AdaptNodeEvaluationDto),
 		NamedChildren: pure_utils.MapValues(evaluation.NamedChildren, AdaptNodeEvaluationDto),
 	}
+
+	if isReturnValueOmitted(evaluation) {
+		nodeEvaluationDto.ReturnValue = omittedReturnValue
```
Pascal-Delange, line 23: How about storing an empty array (or maybe the N first values?) rather than an arbitrary constant string (in the case of list access) ?

balzdur: I plan to store the arbitrary constant for every ommitted return values (so I can easilly parce the result frontend side and explicitly know when the result is skipped).

I know this is not elegant, but with the current form I can't get the information w\ adding another prop to the DTO.

Pascal-Delange: hmm ok, and it needs to be something else than NULL right

Pascal-Delange: fine then
maybe use an all caps string constant but that's just a matter of taste

Pascal-Delange: I'm afraid I don't have better either and we can always change this storage later if needed

Pascal-Delange (approved):

models/ast/node_evaluation_dto.go
balzdur: An alternative would be to :
- store `null`
- store a `boolean` prop like `isResultOmmitted`(or alike)

PR #531, opened 2024-03-28
return decision app link in API decision object

Pascal-Delange:
For an API response body like this 
```json
{
    "id": "585906d7-ea00-4181-9946-26246f561a58",
    "app_link": "https://localhost:3000/decisions/585906d7-ea00-4181-9946-26246f561a58",
    "created_at": "2024-03-26T13:52:17.238933+01:00",
[...]
}
```

Returns `null` if the env variable is not set, and also for decisions nested in the "cases" object return by the cases API

-----
Caveat: I use https:// by default as the scheme, which means if the url is used locally it will give an error "Ce site ne peut pas fournir de connexion sécurisée", but I still prefer that to exposing one more env variable to allow unsecured http links which will work only in dev mode anyway.

### DIFF
--- a/dto/decision_dto.go
+++ b/dto/decision_dto.go
@@ -2,10 +2,13 @@ package dto
 
 import (
 	"encoding/json"
+	"fmt"
+	"net/url"
 	"time"
 
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/models/ast"
+	"github.com/guregu/null/v5"
 )
 
 type GetDecisionInput struct {
@@ -57,6 +60,7 @@ type APIDecisionScenario struct {
 
 type APIDecision struct {
 	Id                   string              `json:"id"`
+	AppLink              null.String         `json:"app_link"`
 	Case                 *APICase            `json:"case,omitempty"`
 	CreatedAt            time.Time           `json:"created_at"`
 	TriggerObject        map[string]any      `json:"trigger_object"`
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.21.0
 	golang.org/x/net v0.18.0
 	golang.org/x/sync v0.3.0
+	golang.org/x/text v0.14.0
```
Pascal-Delange, line 43: unrelated to this PR's subject, just the result of a `go mod tidy`

dto/decision_dto.go
```diff
@@ -98,9 +103,22 @@ func NewAPIDecision(decision models.Decision) APIDecision {
...
 }
 
-func NewAPIDecisionWithRule(decision models.DecisionWithRuleExecutions) APIDecisionWithRules {
+func toDecisionUrl(marbleAppHost string, decisionId string) null.String {
+	if marbleAppHost == "" {
+		return null.String{}
+	}
+
+	url := url.URL{
+		Scheme: "https",
```
balzdur, line 112: If you want to allow specific unsafe protocol for dev, you can do a special case for local development (cf suggestion on CORS)

balzdur (approved):

dto/decision_dto.go
Pascal-Delange: Here it's more annoying, mostly because I really don't want to inject the environment variable all the way down to the DTO adapter. 
So while it works for CORS, I kinda don't want to do it here

PR #532, opened 2024-03-28
Pascal/mar 516 cors setting is not open source friendly on backend

Pascal-Delange:
I'm making this one branch off https://github.com/checkmarble/marble-backend/pull/531 because it comes kind of naturally, as I already needed the app host for the previous PR

### DIFF
--- a/router.go
+++ b/router.go
@@ -15,17 +15,24 @@ import (
 	"github.com/checkmarble/marble-backend/utils"
 )
 
-func corsOption(env string) cors.Config {
-	allowedOrigins := []string{"https://app*.checkmarble.com", "https://backoffice*.checkmarble.com"}
+func corsOption(conf AppConfiguration) cors.Config {
+	protocol := "https://"
+	if conf.env == "development" {
+		protocol = "http://"
+	}
 
-	if env == "development" {
+	allowedOrigins := []string{
+		protocol + conf.config.MarbleAppHost,
+		protocol + conf.config.MarbleBackofficeHost,
+	}
+
+	if conf.env == "development" {
 		allowedOrigins = append(allowedOrigins,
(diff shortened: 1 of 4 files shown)


router.go
```diff
@@ -15,17 +15,19 @@ import (
...
 
-	if env == "development" {
+	if conf.env == "development" {
 		allowedOrigins = append(allowedOrigins,
 			"http://localhost:3000", "http://localhost:3001", "http://localhost:3002", "http://localhost:5173")
 	}
 
 	return cors.Config{
-		AllowOrigins:  allowedOrigins,
-		AllowWildcard: true,
```
Pascal-Delange, line 28: no longer needed as each env will have its own specific values (which is a better practice anyway)

router.go
```diff
@@ -15,17 +15,19 @@ import (
 	"github.com/checkmarble/marble-backend/utils"
 )
 
-func corsOption(env string) cors.Config {
-	allowedOrigins := []string{"https://app*.checkmarble.com", "https://backoffice*.checkmarble.com"}
+func corsOption(conf AppConfiguration) cors.Config {
+	allowedOrigins := []string{
+		"https://" + conf.config.MarbleAppHost,
+		"https://" + conf.config.MarbleBackofficeHost,
+	}
```
balzdur, line 22: I would go for something like that :
```go
protocol := "https://"
if conf.env == "development" {
  protocol = "http://"
}

allowedOrigins := []string{
  protocol + conf.config.MarbleAppHost,
  protocol + conf.config.MarbleBackofficeHost,
}
```

balzdur (approved):

router.go
Pascal-Delange: ah, yes pretty easy to do here, I'll do that

PR #534, opened 2024-03-28
feat: retry ingestion in case of conflict

Pascal-Delange:
The context is that, since the introduction of a unique index on object_id on ingested tables, we can have conflicts at insertion.
Actually, also on unique fields other than the object_id, which is kind of annoying.
In here, I propose to retry (3 times) the repository ingestPayload method, if a conflict is detected.
The problem is that conflicts on object_id are sort of expected (can happen in case of concurrent calls) and should be retried (normally, it should pass on the first retry, unless there is an unreasonable number of concurrent calls on the same object). While conflict errors on other unique fields really shouldn't be retried (because we don't have logic in the code that looks it up by that key, and updates accordingly).

But in practice it's hard for me to differenciate between the two scenarios, so I just retry 3 times. In the worst case it leads to some extra work by the db if the used sends a conflicting row (different object_id but same value of another unique field).

### DIFF
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -11,8 +11,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/avast/retry-go/v4"
 	"github.com/cockroachdb/errors"
-
 	"github.com/google/uuid"
 
 	"github.com/checkmarble/marble-backend/models"
@@ -20,6 +20,7 @@ import (
 	"github.com/checkmarble/marble-backend/repositories"
 	"github.com/checkmarble/marble-backend/usecases/executor_factory"
 	"github.com/checkmarble/marble-backend/usecases/security"
+	"github.com/checkmarble/marble-backend/utils"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


usecases/ingestion_usecase.go
```diff
@@ -45,9 +46,13 @@ func (usecase *IngestionUseCase) IngestObjects(ctx context.Context, organization
...
-	return usecase.transactionFactory.TransactionInOrgSchema(ctx, organizationId, func(tx repositories.Executor) error {
-		return usecase.ingestionRepository.IngestObjects(ctx, tx, payloads, table, logger)
-	})
+
+	ingestClosure := func() error {
+		return usecase.transactionFactory.TransactionInOrgSchema(ctx, organizationId, func(tx repositories.Executor) error {
+			return usecase.ingestionRepository.IngestObjects(ctx, tx, payloads, table, logger)
+		})
+	}
+	return retryIngestion(ctx, ingestClosure)
```
balzdur, line 55: By curiosity, do we need to retry outside the transaction to make it "works" or is it independant ? (I must admit I fail to grasp why retry is needed here)

balzdur (approved):

usecases/ingestion_usecase.go
Pascal-Delange: The idea is the following:
Everything that's done inside the transaction, either is committed or everything is rolled back. It will be rolled back if there is an error (for instance an insertion conflict error).
It's fine and expected that there may be such insertion conflicts, and they may or may not be transient (they are if it's just a concurrency issue).
The thing to do then is to retry everything that's done within the transaction, from the beginning.

PR #540, opened 2024-04-02
feature: audit trail on custom list value insertion

Pascal-Delange:
The three options on the table were:
1. audit information directly on the table (e.g. "added_by", "deleted_by" columns, etc)
2. centralized audit table, handled in our code (run the actions in transactions, try to insert rows by calling a repo from the usecase, etc)
3. or as done here: use postgresql triggers to handle the logic.

Pros:
- it kind of just works out of the box
- very low code overhead
- could be branched off to solution 2 if needed sometime later
Cons:
- it's a bit too magic to be totally reliable (in the sense of "people are aware of it and won't accidentally break it")
- harder to have really fine control on which events we want to audit or not (well, we can possibly fine tune it more but at the cost of making the point above even worse)

All in all, I'm happy with what this does in 100 lines of code, and it would be super easy to replicate this for the other tables that have events to audit (scenario publication, ...)

### DIFF
--- a/repositories/migrations/20240402153800_audit_trail_trigger.sql
+++ b/repositories/migrations/20240402153800_audit_trail_trigger.sql
@@ -1,0 +1,55 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TYPE audit_operation AS ENUM('INSERT', 'UPDATE', 'DELETE');
+
+-- CreateTable
+CREATE TABLE
+	audit (
+		"id" UUID NOT NULL DEFAULT gen_random_uuid () PRIMARY KEY,
+		"operation" audit_operation NOT NULL,
+		"user_id" TEXT,
+		"table" VARCHAR NOT NULL,
+		"entity_id" UUID NOT NULL,
+		"data" JSONB NOT NULL DEFAULT '{}',
+		"created_at" TIMESTAMPTZ (6) NOT NULL DEFAULT CURRENT_TIMESTAMP
+	);
+
+-- Order audit trigger function
+CREATE
+OR REPLACE FUNCTION global_audit () RETURNS TRIGGER AS $$
+    BEGIN
+        IF (TG_OP = 'DELETE') THEN
+            INSERT INTO audit ("operation", "user_id", "table", "entity_id", "data", "created_at")
+            VALUES ('DELETE', current_setting('custom.current_user_id', TRUE), TG_TABLE_NAME, OLD.id, to_jsonb(OLD), now());
+
+        ELSIF (TG_OP = 'UPDATE') THEN
+            INSERT INTO audit ("operation", "user_id", "table", "entity_id", "data", "created_at")
+            VALUES ('UPDATE', current_setting('custom.current_user_id', TRUE), TG_TABLE_NAME, NEW.id, to_jsonb(NEW), now());
+
+        ELSIF (TG_OP = 'INSERT') THEN
+            INSERT INTO audit ("operation", "user_id", "table", "entity_id", "data", "created_at")
+            VALUES ('INSERT', current_setting('custom.current_user_id', TRUE), TG_TABLE_NAME, NEW.id, to_jsonb(NEW), now());
+        END IF;
+        RETURN NULL;
+    END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER audit
+AFTER INSERT
+OR
+UPDATE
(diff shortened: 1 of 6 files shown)


repositories/migrations/20240402153800_audit_trail_trigger.sql
```diff
@@ -0,0 +1,55 @@
...
+		"operation" audit_operation NOT NULL,
+		"user_id" TEXT,
+		"table" VARCHAR NOT NULL,
+		"entity_id" UUID NOT NULL,
+		"data" JSONB NOT NULL DEFAULT '{}',
+		"created_at" TIMESTAMPTZ (6) NOT NULL DEFAULT CURRENT_TIMESTAMP
+	);
+
+-- Order audit trigger function
+CREATE
```
Pascal-Delange, line 18: Directly inspired from a snippet shared on hexa_cto, which itself I believe is directly inspired from some ressources found online.

repositories/migrations/20240402153800_audit_trail_trigger.sql
```diff
@@ -0,0 +1,55 @@
...
+OR REPLACE FUNCTION global_audit () RETURNS TRIGGER AS $$
+    BEGIN
+        IF (TG_OP = 'DELETE') THEN
```
Pascal-Delange, line 21: `TG_*` contains some metadata on the trigger event, similarly `NEW` and `OLD` hold the data of the row acted on (with obvious naming).

repositories/migrations/20240402153800_audit_trail_trigger.sql
```diff
@@ -0,0 +1,55 @@
...
+        END IF;
+        RETURN NULL;
+    END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER audit
+AFTER INSERT
+OR
+UPDATE
+OR DELETE ON custom_list_values FOR EACH ROW
```
Pascal-Delange, line 41: To repeat for every other table where we want to activate the audit logs. Presumably, every sensitive table that is not high I/O.

repositories/migrations/20240402153800_audit_trail_trigger.sql
```diff
@@ -0,0 +1,55 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TYPE audit_operation AS ENUM('INSERT', 'UPDATE', 'DELETE');
+
+-- CreateTable
+CREATE TABLE
+	audit (
+		"id" UUID NOT NULL DEFAULT gen_random_uuid () PRIMARY KEY,
+		"operation" audit_operation NOT NULL,
+		"user_id" TEXT,
```
Pascal-Delange, line 10: Keeping user_id nullable, as it will not always be set

repositories/utils.go
```diff
@@ -1,11 +1,27 @@
...
+func setCurrentUserIdContext(ctx context.Context, exec Executor, userId *models.UserId) error {
+	if userId != nil {
+		_, err := exec.Exec(
+			ctx,
+			fmt.Sprintf("SELECT SET_CONFIG('%s', $1, false)", postgres_audit_user_id_parameter),
+			*userId,
+		)
+		return err
+	}
+	return nil
```
Pascal-Delange, line 22: no error here if userId is null. If we want to make double sure that it's properly filled, that should happen at the usecase level

repositories/utils.go
```diff
@@ -1,11 +1,27 @@
...
 	"github.com/checkmarble/marble-backend/pure_utils"
 )
 
+const postgres_audit_user_id_parameter = "custom.current_user_id"
+
+func setCurrentUserIdContext(ctx context.Context, exec Executor, userId *models.UserId) error {
+	if userId != nil {
+		_, err := exec.Exec(
+			ctx,
+			fmt.Sprintf("SELECT SET_CONFIG('%s', $1, false)", postgres_audit_user_id_parameter),
```
Pascal-Delange, line 17: `, false)` here means that the config will not be kept after the transaction. This also means that it only makes sense if called within a transaction.
We could of course let it survive for the duration of the the connection, and make sure we set if before each audited table write in a reused connection, but I'm close to 100% sure that we'll forget it and insert garbage data at some point if we do (at least, if we do without some more advanced bookkeeping - it may be possible to implement a cleanup when the connection is returned to the pool, for instance)

balzdur (approved):

PR #544, opened 2024-04-09
Thomas/mar 524 filter decisions by scheduled execution

balzdur:
In this PR :
- allow to filter decisions by `scheduled_execution_id`
- rename dto to use internal conventions (snake case + singuler for arrays)
- update `public_api.yaml` spec

Caveat: the public api is not yet pushed to readme

### DIFF
--- a/specs/public-api.yaml
+++ b/specs/public-api.yaml
@@ -68,15 +68,80 @@ paths:
         - Decision
       security:
         - ApiKeyAuth: []
-      description: List existing decisions for the organization.
-      summary: Query a list of decisions
+      description: List matching decisions for the organization
+      summary: List decisions based on the provided filters and pagination
+      parameters:
+        - name: outcome[]
+          description: decision outcomes used to filter the list
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              $ref: "#/components/schemas/outcome"
+        - name: scenario_id[]
+          description: scenario IDs used to filter the list
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+              format: uuid
+        - name: trigger_object[]
+          description: trigger objects used to filter the list
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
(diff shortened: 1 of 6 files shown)


specs/public-api.yaml
```diff
@@ -68,15 +68,80 @@ paths:
...
+          description: scheduled execution IDs used to filter the list
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+              format: uuid
+        - name: sorting
+          description: the field used to sort the items
```
Pascal-Delange, line 117: could you please double check what happens if you pass this param with a value other than created_at ?

specs/public-api.yaml
```diff
@@ -68,15 +68,80 @@ paths:
...
         200:
-          description: A decision was successfully taken.
+          description: List of corresponding decisions
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/decisions"
+                allOf:
+                  - $ref: "#/components/schemas/pagination"
+                  - type: object
```
Pascal-Delange, line 137: ah yes good fix here, the old doc was wrong

specs/public-api.yaml
balzdur: Error 500 is returned
```JSON
{
    "message": "An unexpected error occurred. Please try again later, or contact support if the problem persists.",
    "error_code": ""
}
```

There is a check in the repository to ensure we respect the whitelist
```go
	var offsetField any
	switch p.Sorting {
	case models.DecisionSortingCreatedAt:
		offsetField = offset.CreatedAt
	default:
		// only pagination by created_at is allowed for now
		return query, fmt.Errorf("invalid sorting field: %w", models.BadParameterError)
	}
```

balzdur: Ok for me, but I will handle this in a separated PR.

I need to merge this to match the merged PR in the front end

Pascal-Delange: yep sure 👍

PR #547, opened 2024-04-12
remove data model custom types

Pascal-Delange:
Seemed like a good idea at first, actually it just cluttered the whole codebase with casting from string to custom string types or vice versa.

### DIFF
--- a/models/data_model.go
+++ b/models/data_model.go
@@ -57,7 +57,6 @@ type DataModel struct {
 
 type (
 	TableName string
-	FieldName string
 )
 
 func (dm DataModel) Copy() DataModel {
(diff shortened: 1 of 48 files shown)


models/data_model.go
```diff
@@ -52,25 +52,11 @@ func DataTypeFrom(s string) DataType {
 
 type DataModel struct {
 	Version string
-	Tables  map[TableName]Table
-}
-
-type (
-	TableName string
-	FieldName string
-	LinkName  string
```
Pascal-Delange, line 61: basically this change - and all the rest follows to make it compile again

balzdur (approved):

PR #548, opened 2024-04-16
WIP - transfercheck input validation & hashing

Pascal-Delange:
- actually hash the beneficiary iban
- but do it only if we don't already receive a hash (well, something that looks like a hash - namely 32 bytes in hexa)
- while we're at it, add some more input validation on the other fields that are expected
    - caveat: some fields are format validate by gin at the binding layer, while some fields are validated in the code, and that results in different error api returns - but good enough for now
- expected formats:
    - bic: 8 or 11 chars, alphanumeric
    - iban: alphanumeric, 16 to 34 chars (or already a SHA256 hash in hexadecimal representation)
    - currency: a valid ISO3 currency code
    - ip address: a valid ip address
    - status: one of "neutral", "suspected_fraud", "confirmed_fraud"
    - beneficiary_name, label, sender_device: free string, max 140 chars
    - sender_account_id, transfer_id: free string, max 50 chars
    - value: greater than or equal 0
- as before:
    - timestamps in iso format "2000-01-01T00:00:00Z"

example api response with bad input 
```
{
    "transfer_data": {
        "beneficiary_bic": "SOGEFRPPqzefqzeqzef",
        "beneficiary_iban": "qzef",
        "beneficiary_name": "",
        "currency": "EUREA",
        "label": "my first transfer with high riskmy first transfer with high riskmy first transfer with high riskmy first transfer with high riskmy first transfer with high risk",
        "sender_account_id": "123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456123456",
        "sender_bic": "BOUSFR",
        "sender_device": "Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10Iphone 10",
        "sender_ip": "127.0.0",
        "status": "neutra",
        "timezone": "Europe/Paris",
        "transfer_id": "5",
        "transfer_requested_at": "2024-03-20T00:00:00Z",
        "value": 1
    }
}
```

```
{
    "errors": {
        "beneficiary_bic": "bic must be 8 or 11 characters: bad parameter",
        "beneficiary_iban": "iban must be between 15 and 34 characters: bad parameter",
        "currency": "currency EUREA is not valid",
        "label": "string is too long",
        "sender_account_id": "string is too long",
        "sender_bic": "bic must be 8 or 11 characters: bad parameter",
        "sender_device": "string is too long",
        "sender_ip": "sender_ip 127.0.0 is not a valid IP address",
        "status": "status neutra is not valid"
    }
}
```

### DIFF
--- a/models/errors.go
+++ b/models/errors.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	"fmt"
+
 	"github.com/cockroachdb/errors"
 )
 
@@ -59,22 +61,9 @@ var (
 // ingestion and decision creating payload related errors
 var FormatValidationError = errors.New("The input object is not valid")
 
-type PayloadValidationErrors struct {
-	message string
-	errors  map[string]string
-}
+// transfercheck errors
+type FieldValidationError map[string]string
 
-func (p PayloadValidationErrors) Error() string {
-	return p.message
-}
-
-func (p PayloadValidationErrors) Errors() map[string]string {
-	return p.errors
-}
-
-func NewPayloadValidationErrors(message string, errors map[string]string) PayloadValidationErrors {
-	return PayloadValidationErrors{
-		message: message,
-		errors:  errors,
-	}
+func (e FieldValidationError) Error() string {
+	return fmt.Sprintf("%v", map[string]string(e))
 }
--- a/models/transfer_check.go
+++ b/models/transfer_check.go
@@ -195,7 +195,7 @@ func (t TransferDataCreateBody) FormatAndValidate() (TransferDataCreateBody, err
 	errs := make(FieldValidationError, 10)
 
 	// hash the iban if it's clear - otherwise keep it unchanged
-	t.BeneficiaryIban, err = hashIbanIfClear(t.BeneficiaryIban)
+	t.BeneficiaryIban, err = validateIbanOrHashIfClear(t.BeneficiaryIban)
 	if err != nil {
 		errs["beneficiary_iban"] = err.Error()
 	}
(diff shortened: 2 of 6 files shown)


models/errors.go
```diff
@@ -59,22 +61,9 @@ var (
 // ingestion and decision creating payload related errors
 var FormatValidationError = errors.New("The input object is not valid")
 
-type PayloadValidationErrors struct {
```
Pascal-Delange, line 62: no longer used

models/transfer_check.go
```diff
@@ -175,6 +190,103 @@ func (t TransferDataCreateBody) ToIngestionMap(mapping TransferMapping) map[stri
...
+	if !isAlphanumeric.MatchString(iban) {
+		return "", errors.Wrap(BadParameterError, "iban must be alphanumeric")
+	}
+	if len(iban) < 15 || len(iban) > 34 {
+		return "", errors.Wrap(BadParameterError, "iban must be between 16 and 34 characters")
+	}
+
+	hash := sha256.Sum256([]byte(trimAndUpper(iban)))
+	return strings.ToUpper(hex.EncodeToString(hash[:])), nil
+}
```
balzdur, line 284: IIUC, the spec is :
- compute the sha256 HASH
- encode in human readable hex format
- transform to UPPPERCASE since the format is case insensitive (to ensure we can easilly compare similar hash by string equality)

balzdur (approved):

models/transfer_check.go
Pascal-Delange: Yes, with an additional step (first) to remove spaces and upper case (so that the hash is also comparable).
Actually, I'll change it a bit to "remove spaces" rather than simply "trim spaces" on the raw iban.

PR #553, opened 2024-04-23
Pivot values on decisions

Pascal-Delange:
Mostly as described here https://www.notion.so/checkmarble/Defining-a-pivot-67cf95c5a94249fa95ba8c183e54751b.

One important point (to discuss from a product point of view but I think it makes sense): 
- _**because**_ we choose (for now) that if using links, the pivot value is the value of the parent field of the last link (e.g. accounts.object_id if the link is tx.account_id=>accounts.object_id but not accounts.some_other_field)
- and _**because**_ the value if the penultimate table child field is the same as the last table's parent field (as in the link - unless the last table object has not been ingested) 
- **_then_** we can just use the last-table-but-one's "foreign key" value.

It's clearer with two examples:
- use the account_id from transactions in payload rather than the ingested account's object_id
- use the company_id of the ingested account (assuming base table transactions) rather than the ingested company's object_id 

This is implemented in the function getPivotValue in evaluate_scenario.go

--- 
Filtering of decisions by pivot value in a follow up PR


models/data_model_pivot.go
```diff
@@ -77,16 +78,18 @@ func FieldFromPath(dm DataModel, pathLinkIds []string, baseTableName string) (Fi
...
+	var fieldId string
 	if firstLink.ChildTableName != baseTableName {
 		return Field{}, errors.Wrap(
 			BadParameterError,
 			fmt.Sprintf(`first link's (%s) child table must be the base table "%s" (is "%s" instead)`,
 				firstLink.Id, baseTableName, firstLink.ChildTableName,
 			),
 		)
+	} else {
+		fieldId = firstLink.ParentFieldId
```
Pascal-Delange, line 90: this was an error (wasn't working for path of length one)

usecases/decision_usecase.go
```diff
@@ -219,10 +219,17 @@ func (usecase *DecisionUsecase) CreateDecision(
 		return models.DecisionWithRuleExecutions{}, err
 	}
 
+	pivotsMeta, err := usecase.dataModelRepository.ListPivots(ctx, exec, input.OrganizationId, nil)
+	if err != nil {
+		return models.DecisionWithRuleExecutions{}, err
+	}
+	pivot := models.FindPivot(pivotsMeta, input.TriggerObjectTable, dataModel)
```
balzdur, line 226: I would probably go for a solution like :
```go
triggerTable := dataModel.GetTableByName(input.TriggerObjectTable) //To implement
usecase.dataModelRepository.GetPivot(triggerTable.ID)
```
 but both solution work

models/data_model_pivot.go
balzdur: What about this refactoring proposition ?
```go
// separate validation from field access, use it `data_model_usecase.go`
func ValidatePathPivot(dm DataModel, pathLinkIds []string, baseTableName string) error {
        linksMap := dm.AllLinksAsMap()
        
	// check that the first link is from the base table
	firstLink := linksMap[pathLinkIds[0]]
	if firstLink.ChildTableName != baseTableName {
		return Field{}, errors.Wrap(
			BadParameterError,
			fmt.Sprintf(`first link's (%s) child table must be the base table "%s" (is "%s" instead)`,
				firstLink.Id, baseTableName, firstLink.ChildTableName,
			),
		)
	}

	// check that the links are chained consistently
	for i := 1; i < len(pathLinkIds); i++ {
		previousLink := linksMap[pathLinkIds[i-1]]
		currentLink := linksMap[pathLinkIds[i]]
		if previousLink.ParentTableName != currentLink.ChildTableName {
			return Field{}, errors.Wrap(
				BadParameterError,
				fmt.Sprintf(`link %s (parent table "%s") is not a child of link %s (child table "%s")`,
					previousLink.Id, previousLink.ParentTableName, currentLink.Id, currentLink.ChildTableName,
				),
			)
		}
	}
}

// you can transform this method to return an error if len(pathLinkIds) == 0, but you get the idea
func FieldFromPath(dm DataModel, pathLinkIds []string, baseTableName string) Field {
        linksMap := dm.AllLinksAsMap()
        fieldId = linksMap[len(pathLinkIds)-1]
        return dm.AllFieldsAsMap()[fieldId]
}
```

An alternative is to keep both functions merged together (validation + access) but at least, separate validation from access (the current mix inside the loop for `fieldId` seems unecessary)

balzdur (approved):

usecases/decision_usecase.go
Pascal-Delange: Ok let me see.
`usecase.dataModelRepository.GetPivot(triggerTable.ID)` kind of already exists (ListPivots takes a last pointer tableId argument to filter on a specific table). It still returns a slice of pivots.
`dataModel.GetTableByName(` also exists, it's basically `dataModel.Tables[name]`. 
The advantage of my FindPivot is that I hide a call to adapt the pivot metadata to a full pivot inside.
I think both are very viable, but I don't see a strong incentive to change it now that it's already written.
Thanks for the idea though.

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -139,11 +147,15 @@ func EvalScenario(
...
 		ScenarioVersion:     publishedVersion.Version,
+		PivotValue:          pivotValue,
 		RuleExecutions:      ruleExecutions,
 		Score:               score,
 		Outcome:             outcome,
 		OrganizationId:      params.Scenario.OrganizationId,
 	}
+	if params.Pivot != nil {
+		se.PivotId = &params.Pivot.Id
+	}
```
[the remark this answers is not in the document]
Pascal-Delange: yep totally ok to do that

models/data_model_pivot.go
Pascal-Delange: Ok j'achète la proposition 👍

PR #555, opened 2024-04-25
query decisions by pivot value

Pascal-Delange:
Add a new possibility to filter decisions by pivot_value in the API. Add an index on `decisions(org_id,pivot_value,created_at DESC)`.
Still as discussed here : https://www.notion.so/checkmarble/Defining-a-pivot-67cf95c5a94249fa95ba8c183e54751b.
I propose to wait a bit before we include this in the public API doc, on purpose.

--- 

The last commit is not directly related to this PR's subject, but cleans up indexes on the decisions table:
Before 
```
CREATE INDEX decisions_org_id_idx ON decisions(org_id uuid_ops,created_at DESC);
CREATE INDEX decisions_scheduled_execution_id_idx ON decisions(scheduled_execution_id,created_at DESC);
CREATE INDEX scenario_id_idx ON decisions(org_id,scenario_id,created_at DESC);
CREATE INDEX outcome_idx ON decisions(org_id,outcome,created_at DESC);
CREATE INDEX trigger_object_type_idx ON decisions(org_id,trigger_object_type,created_at DESC);
CREATE INDEX decisions_case_id_idx ON decisions(org_id uuid_ops,case_id);
CREATE INDEX decision_object_id_idx ON decisions(org_id uuid_ops,(trigger_object ->> 'object_id'::text));
CREATE INDEX decisions_pivot_value_index ON decisions(org_id,pivot_value,created_at DESC);
```
After 
```
CREATE INDEX decisions_case_id_idx ON decisions(org_id,case_id);
CREATE INDEX decision_object_id_idx ON decisions(org_id,(trigger_object ->> 'object_id'::text));
CREATE INDEX decisions_pivot_value_index ON decisions(org_id,pivot_value,created_at DESC);
CREATE INDEX decisions_by_org_id_index ON decisions(org_id,created_at DESC) INCLUDE(scenario_id,outcome,trigger_object_type,case_id);
CREATE INDEX decisions_scheduled_execution_id_idx_2 ON decisions(org_id,scheduled_execution_id,created_at DESC);
```

Based on the following guidelines broadly:
- if we want to allow to filter decisions by something very specific (AKA few decisions have this attribute), probably it's worth having its own index (and let the query optimizer do its job): e.g. if we filter by pivot_value, presumably only a handful out of many decisions match a given value, we don't want to do a full index scan. Is the case for:
    - scheduled_execution_id
    - case_id
    - pivot_value
    - trigger_object.object_id
- inversely if it's reasonable to assume that a kinda large % of decisions match a given value, it's reasonable _until performance issues prove otherwise_ to just use one index (on org_id, created_at desc) and include the other tables we want to filter => `CREATE INDEX decisions_by_org_id_index ON decisions(org_id,created_at DESC) INCLUDE(scenario_id,outcome,trigger_object_type,case_id);`

### DIFF
--- a/repositories/decisions_repository.go
+++ b/repositories/decisions_repository.go
@@ -26,13 +26,18 @@ type DecisionRepository interface {
 		decisionIds []string,
 	) ([]models.DecisionWithRuleExecutions, error)
 	DecisionsById(ctx context.Context, exec Executor, decisionIds []string) ([]models.Decision, error)
-	DecisionsByCaseId(ctx context.Context, exec Executor, caseId string) (
-		[]models.DecisionWithRuleExecutions, error)
+	DecisionsByCaseId(
+		ctx context.Context,
+		exec Executor,
+		organizationId, caseId string,
+	) ([]models.DecisionWithRuleExecutions, error)
 	DecisionsByObjectId(ctx context.Context, exec Executor, organizationId string, objectId string) ([]models.DecisionCore, error)
 	DecisionsOfScheduledExecution(
 		ctx context.Context,
 		exec Executor,
-		scheduledExecutionId string) (<-chan models.DecisionWithRuleExecutions, <-chan error)
+		organizationId string,
+		scheduledExecutionId string,
+	) (<-chan models.DecisionWithRuleExecutions, <-chan error)
 	StoreDecision(
 		ctx context.Context,
 		exec Executor,
@@ -134,6 +139,7 @@ func (repo *DecisionRepositoryImpl) DecisionsById(ctx context.Context, exec Exec
 func (repo *DecisionRepositoryImpl) DecisionsByCaseId(
 	ctx context.Context,
 	exec Executor,
+	organizationId string,
 	caseId string,
 ) ([]models.DecisionWithRuleExecutions, error) {
 	if err := validateMarbleDbExecutor(exec); err != nil {
@@ -141,6 +147,7 @@ func (repo *DecisionRepositoryImpl) DecisionsByCaseId(
 	}
 
 	query := selectDecisions().
+		Where(squirrel.Eq{"org_id": organizationId}).
 		Where(squirrel.Eq{"case_id": caseId}).
 		OrderBy("created_at DESC")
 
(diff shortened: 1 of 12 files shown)


usecases/decision_usecase.go
```diff
@@ -97,36 +97,31 @@ func (usecase *DecisionUsecase) ListDecisions(
 		return []models.DecisionWithRank{}, err
 	}
 
-	return executor_factory.TransactionReturnValue(
```
Pascal-Delange, line 100: non-necessary transaction

repositories/decisions_repository.go
```diff
@@ -393,6 +401,7 @@ func (repo *DecisionRepositoryImpl) DecisionsOfScheduledExecution(
 		ctx,
 		exec,
 		selectDecisions().
+			Where(squirrel.Eq{"org_id": organizationId}).
```
Pascal-Delange, line 404: First level of fix for permissions when querying decisions by scheduled_execution_id. It's still not perfect because I'd rather this permission is enforced in the usecase, as in the rest of our code. However, because this function uses a channel of decisions that are streamed directly to the writer higher up the stack, it's a bit more complicated to do so here (we should probably add an intermediate layer in the usecase where we read the repo's channel and send to another channel higher up.
Anyway, this is the minimum change so that one cannot query another org's scheduled ex decisions.

usecases/scheduledexecution/export_schedule_execution.go
```diff
@@ -122,6 +144,13 @@ func (exporter *ExportScheduleExecution) ExportDecisions(ctx context.Context, sc
 	var number_of_exported_decisions int
 
 	for decision := range decisionChan {
+		if decision.OrganizationId != organizationId {
```
Pascal-Delange, line 147: To be more precise and expand on the comment above (repositories/decisions_repository.go), this makes sure in the usecases layer we don't send a decision to an org it doesn't belong to - but I'm not totally sure what happens with the channel if we return here. Probably if we return an error the context will end up being canceled higher up and it's fine, but I'd like to double check. I made a note on linear.
(We're not supposed to close the channel from the receiver side)

balzdur (approved):

PR #557, opened 2024-04-29
fix(dto): handle optionnal path_link_ids

balzdur:
**The issue :**

When sending this HTTP request, an internal error is raised
```
POST /data-model/pivots

{
    "base_table_id": "476e49fe-8d08-4c35-9dd5-6abd6a028527",
    "field_id": "74114c55-7fb7-43e6-9867-34162b72827a"
}
```

Before the fix, it was required to always send `path_link_ids` in the payload, even for "field" pivot (= no need for links)


Pascal-Delange (approved):
thanks!

PR #561, opened 2024-05-02
Pascal/mar 560 custom roles permissions for users as part of a

Pascal-Delange:
As it is, I get something like this if I connect as a user with role 9 "transfercheck_user"
Capture d’écran 2024-05-02 à 16 56 45
 (though if I check the scheduled executions page or inboxes page, I do get a non empy view)


balzdur (approved):

balzdur:
After digging a little bit:
- `scheduled executions`: if there is none, the backend return a 200, if there are some, you get a proper 403
- `cases`: it looks the same

Pascal-Delange:
thanks for checking, yes that was my intuition

PR #562, opened 2024-05-03
create extension pg_cron

Pascal-Delange:
## Edit after some testing and debugging:
- the pg_cron extension is installed by default [in a cloud SQL instance](https://cloud.google.com/sql/docs/postgres/extensions), but needs to be installed when using a local db
- if the db runs in docker, that means adding it in the image, as below (inspiration from [here](https://datacraze.io/postgresql-docker-image-adding-extensions/))
```dockerfile
FROM postgres:15

RUN apt-get update

RUN apt-get update \
      && apt-get install -y curl \
      && apt-get -y install postgresql-15-cron

RUN echo "shared_preload_libraries='pg_cron'" >> /usr/share/postgresql/postgresql.conf.sample
RUN echo "cron.database_name='marble'" >> /usr/share/postgresql/postgresql.conf.sample
```
- I did this and pushed the custom image to `europe-west1-docker.pkg.dev/marble-infra/marble/postgresql-db:latest`
- for some reason I can't understand, this image works fine locally, but if used instead of `postgres:15` with dockertest (see `integration_test/init_test.go`) it properly starts the db but fails to ping it
- because I really only need the extension on cloud and local (not in customer docker), I won't fight this right now: I just changed the local postgres image definition in `docker_compose.yaml` and will add some doc on how to make it work in local dev mode => [the doc](https://www.notion.so/checkmarble/Making-pg_cron-work-locally-c61b860bff5e41a78605300c3ba347c8)


Pascal-Delange:
ah shit, it's broken on the test db

balzdur (approved):

balzdur (approved):

PR #563, opened 2024-05-06
fix: sort columns when reading data model columns, for consistent ingestion queries

Pascal-Delange:
Currently we end up with this in the query insights
Capture d’écran 2024-05-06 à 16 47 58
we may still have duplicates corresponding to different numbers of rows to ingest, but it should be a lot better

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
   db:
     container_name: postgres
     image: europe-west1-docker.pkg.dev/marble-infra/marble/postgresql-db:latest # custom image of postgres 15 with pg_cron extension added
+    shm_size: 1g
     restart: always
     environment:
       POSTGRES_USER: postgres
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
   db:
     container_name: postgres
     image: europe-west1-docker.pkg.dev/marble-infra/marble/postgresql-db:latest # custom image of postgres 15 with pg_cron extension added
+    shm_size: 1g
```
Pascal-Delange, line 7: unrelated - needed so I could VACUUM locally

repositories/ingestion_repository.go
```diff
@@ -162,20 +162,18 @@ func (repo *IngestionRepositoryImpl) batchInsertPayloadsAndEnumValues(ctx contex
 	columnNames := models.ColumnNames(table)
```
Pascal-Delange, line 162: This is where I want to insert values in a consistent order

balzdur (approved):
I agree with the changes, I just may miss the point of this PR (related to the comment).

IIUC, the screenshot is showing distinct queries but by ordering inserted columns, it's the same, is that right ?
It won't reduce the average amount of time necessary to insert which is close to ~1sec

Pascal-Delange:
No, you're right about that
(that's ~1s for 1000 rows though, so ~1ms per row)
This PR is more about getting usable monitoring metrics

PR #569, opened 2024-05-13
feat(partner): allow to create users with partner_id

balzdur:
Small changes to allow create users by API


Pascal-Delange (approved):
alright 👍  When we want to allow transfercheck to create users themselves, we'll need to worry about permissions and input validation - so long as it's just marble admins that create users, this should be fine

PR #571, opened 2024-05-20
create workflow to add decision to an existing case

Pascal-Delange:
## Context
Implemented according to the specs discussed in https://www.notion.so/checkmarble/Decision-to-case-add-to-open-case-ac05bcd9d9304e1ca72ee9cafc65fe22.

### Selection rules
We try to find a case, with the following preferences, in order:
- prefer the least number of distinct pivot values present in the case (in particular, prefer a case that has only one pivot value matching the context decision)
- prefer a case that is "open" rather than "investigating"
- all else equal, take the most recent case

### DB queries & indexes
For now, the choice is to do the matching synchronously. There will be lots of decisions overall, and possibly quite a few cases as well, so it is important that we design well the query that will select the candidate matching cases. Here, I tried to do this by making sure we select:
- on one side, the cases that are in the right org, inbox, and are open, ordered by case id
- on the other side, the decisions in the right org, matching pivot_value, non null case_id, ordered by case_id
- both sides of the join being ordered (they also should not be too large), we can expect a merge join

It still requires a new index on the decisions table, I didn't find a good way to avoid it.

Still, to be monitored.

### Tests
I did some manual operations to test all the base cases end to end, as best as I could
Capture d’écran 2024-05-21 à 17 34 12

### DIFF
--- a/models/case.go
+++ b/models/case.go
@@ -45,11 +45,10 @@ type CreateCaseAttributes struct {
 }
 
 type UpdateCaseAttributes struct {
-	Id          string
-	DecisionIds []string // Decision Ids to add to the case - not list of decision ids to replace on the case
-	InboxId     string
-	Name        string
-	Status      CaseStatus
+	Id      string
+	InboxId string
+	Name    string
+	Status  CaseStatus
 }
 
 type CreateCaseCommentAttributes struct {
--- a/repositories/migrations/20240521171800_add_to_case_workflow_indexes.sql
+++ b/repositories/migrations/20240521171800_add_to_case_workflow_indexes.sql
@@ -2,7 +2,8 @@
 -- +goose StatementBegin
 CREATE INDEX decisions_add_to_case_idx ON decisions (org_id, pivot_value, case_id)
 WHERE
-      pivot_value IS NOT NULL;
+      pivot_value IS NOT NULL
+      AND case_id IS NOT NULL;
 
 CREATE INDEX cases_add_to_case_workflow_idx ON cases (org_id, inbox_id, id)
 WHERE
--- a/repositories/case_and_decisions.go
+++ b/repositories/case_and_decisions.go
@@ -20,7 +20,7 @@ func (repo *MarbleDbRepository) SelectCasesWithPivot(
 	query := `SELECT c.id, c.status, c.created_at, c.org_id
 	FROM cases AS c
 	INNER JOIN (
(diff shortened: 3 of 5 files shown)


Pascal-Delange:
Remaining to do at this point: implement the two new repository methods necessary for the case matching

Pascal-Delange:
At this commit, the feature works - last thing to do is check if we need some more indexes for the new db queries

models/case.go
```diff
@@ -39,7 +46,7 @@ type CreateCaseAttributes struct {
 
 type UpdateCaseAttributes struct {
 	Id          string
-	DecisionIds []string
+	DecisionIds []string // Decision Ids to add to the case - not list of decision ids to replace on the case
```
balzdur, line 49: If this is confusing, you may pick an explcit prop name like `DecisionToAddIds`

repositories/migrations/20240521171800_add_to_case_workflow_indexes.sql
```diff
@@ -0,0 +1,26 @@
...
+-- +goose StatementBegin
+DROP INDEX decisions_case_id_idx;
+
+DROP INDEX cases_add_to_case_workflow_idx;
+
+DROP INDEX decisions_add_to_case_idx;
+
+CREATE INDEX decisions_case_id_idx ON decisions (org_id, case_id);
+
+-- +goose StatementEnd
```
balzdur, line 26: FYI, I found an existing index defined like this in our migration files:
```sql
CREATE INDEX decisions_pivot_value_index ON decisions (org_id, pivot_value, created_at DESC);
```

balzdur (approved):

models/case.go
Pascal-Delange: true 👍  good point

repositories/migrations/20240521171800_add_to_case_workflow_indexes.sql
Pascal-Delange: Yea I know, it's close enough. Maybe it would have been sufficient, at least in most cases. The thing is, to provide for the case where there are many entries in both sides of the join (which can happen), I wanted to make sure we do  a merge join rather than a loop join, for which I want the decisions to be filtered by org_id/pivot_value and ordered by case_id, which is what I added this index for. I'll make sure to monitor this once it's live

models/case.go
Pascal-Delange: ok turned out it's even better, it's not even the struct I was really using for the update: this field was never actually used anywhere (I did the rename trick on another variable though, cf last commit)

PR #572, opened 2024-05-23
fix: ingestion logic in usecase, not handler. Don't read data model e…

Pascal-Delange:
…num values every time

### DIFF
--- a/api/handle_ingestion.go
+++ b/api/handle_ingestion.go
@@ -2,82 +2,36 @@ package api
 
 import (
 	"encoding/csv"
-	"encoding/json"
 	"fmt"
 	"io"
-	"log/slog"
 	"net/http"
 
-	"github.com/cockroachdb/errors"
 	"github.com/gin-gonic/gin"
 
 	"github.com/checkmarble/marble-backend/dto"
-	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/pure_utils"
-	"github.com/checkmarble/marble-backend/usecases/payload_parser"
 	"github.com/checkmarble/marble-backend/utils"
 )
 
 func (api *API) handleIngestion(c *gin.Context) {
-	logger := utils.LoggerFromContext(c.Request.Context())
 	organizationId, err := utils.OrgIDFromCtx(c.Request.Context(), c.Request)
 	if presentError(c, err) {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


api/handle_ingestion.go
```diff
@@ -2,81 +2,36 @@ package api
...
-		return
-	}
-	logger = logger.With(slog.String("object_type", objectType))
-
-	tables := dataModel.Tables
-	table, ok := tables[objectType]
-	if !ok {
-		logger.ErrorContext(c.Request.Context(), "Table not found in data model for organization")
-		http.Error(c.Writer, "", http.StatusNotFound)
+	if presentError(c, err) {
```
balzdur, line 24: This no longer return `http.StatusUnprocessableEntity` with the current implementation of `presentError`

usecases/ingestion_usecase.go
```diff
@@ -40,17 +44,63 @@ type IngestionUseCase struct {
...
+		return 0, errors.Wrap(models.BadParameterError, string(encoded))
+	}
+
+	var nb int
 	ingestClosure := func() error {
 		return usecase.transactionFactory.TransactionInOrgSchema(ctx, organizationId, func(tx repositories.Executor) error {
-			return usecase.ingestionRepository.IngestObjects(ctx, tx, payloads, table)
+			var err error
+			nb, err = usecase.ingestionRepository.IngestObjects(ctx, tx, []models.ClientObject{payload}, table)
+			return err
```
balzdur, line 100: For my curiosity, this syntax ensure we redeclare a `var err error` specitific variable inside the scope of the inline `func` that do not rely on the "outter scope" err ?

balzdur (approved):

usecases/ingestion_usecase.go
Pascal-Delange: Hmm, well yes, though in fact the line 98 is useless here (I removed it), because there already exists a variable `err` higher up in the function.
What I don't want specifically is to use `:=` line 99 because I want to update the variable `nb`
defined outside of the inline function. So if there was no variable err higher in the IngestObjects method, I may be tempted to do `nb, err:= ...` which would be wrong because it would define a new `nb` that is not the outer scope `nb`

api/handle_ingestion.go
Pascal-Delange: Yes that's right, I'll fix that

PR #574, opened 2024-05-23
try to silence sentry errors when cloud run fails to create unix socket

Pascal-Delange:
I spent some time trying to find exactly what kind of error this should be. After failing, I went for stupid string matching 🤷 

Edit 24/05: Ok while I was at it I also took the opportunity to add sentry monitoring and tracing to the other jobs, with a helper function to avoid repeating all the code

### DIFF
--- a/jobs/execute_with_monitoring.go
+++ b/jobs/execute_with_monitoring.go
@@ -1,0 +1,72 @@
+package jobs
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/checkmarble/marble-backend/tracing"
+	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/checkmarble/marble-backend/utils"
+	"github.com/getsentry/sentry-go"
+)
+
+func executeWithMonitoring(
+	ctx context.Context,
+	uc usecases.Usecases,
+	config tracing.Configuration,
+	jobName string,
+	fn func(context.Context, usecases.Usecases) error,
+) error {
+	telemetryRessources, err := tracing.Init(config)
+	if err != nil {
+		return fmt.Errorf("error initializing tracing: %w", err)
+	}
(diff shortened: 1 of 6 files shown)


jobs/schedule_due_scenarios.go
```diff
@@ -31,17 +42,13 @@ func ScheduleDueScenarios(ctx context.Context, usecases usecases.Usecases) error
 	runScheduledExecution := usecasesWithCreds.NewRunScheduledExecution()
 
 	for _, scenario := range scenarios {
-		logger.DebugContext(ctx, "executing scenario",
+		logger.DebugContext(ctx, "Scheduled scenario scenario",
```
balzdur, line 45: Typo ?

jobs/execute_with_monitoring.go
```diff
@@ -0,0 +1,72 @@
...
+		nil,
+	)
+
+	err = fn(ctx, uc)
+	if err != nil {
+		// Known issue where Cloud Run will sometimes fail to create the unix socket to connect to CloudSQL. In this case, we don't log the error in Sentry.
+		if strings.Contains(err.Error(), "failed to connect to `host=/cloudsql/") {
+			logger.WarnContext(ctx, "Failed to create unix socket to connect to CloudSQL. Wait for the next execution of the job.")
+			return nil
+		}
```
balzdur, line 44: You may try to log more error detail if you plan to get rid of the string matching

balzdur (approved):

jobs/schedule_due_scenarios.go
Pascal-Delange: ah yes, probably

jobs/execute_with_monitoring.go
Pascal-Delange: Hmm well I don't know, either we get an error that matches this (and we silence it on sentry because there's nothing we can do with it), or we get something different and it's caught in the base error case. I think it's ok like this

PR #581, opened 2024-06-05
WIP: implement data enrichment for ip addresses on transfercheck

Pascal-Delange:
## Recap
Overall I'm satisfied with this PR as it is.
It expects to find a set of csv files in the transfercheck bucket on GCS, containing ip ranges. At the first transfercheck request received by the instance, they are read and put in memory. For an incoming sender ip address, we do a dichotomy search on the known ip ranges to find if it matches a given country/vpn/tor address range.

Because the GCS files are downloaded at the first request, that means we have a bit of a cold start (e.g. ~1sec on the first request) on the instance for transfercheck routes. But so be it, the possible workarounds are just too cumbersome compared to what we lose, for now anyway.

## To do:
- implement an expiry date on the data held in memory, preparing for...
- ... refresh the files on GCS on a schedule (daily at least?) with a script (I'll see where I put it, I kind of don't want to pollute the marble OS codebase with background scripts for transfercheck...)

## Prerequisite
 https://github.com/checkmarble/terraform/pull/21

## Examples of csv files with the right format
[ip_tor_ranges.csv](https://github.com/user-attachments/files/15592486/ip_tor_ranges.csv)
[ip_vpn_ranges.csv](https://github.com/user-attachments/files/15592487/ip_vpn_ranges.csv)
[ip_country_ranges.csv](https://github.com/user-attachments/files/15592488/ip_country_ranges.csv)
Capture d’écran 2024-06-05 à 16 50 56

### DIFF
--- a/repositories/repositories.go
+++ b/repositories/repositories.go
@@ -10,24 +10,25 @@ import (
 )
 
 type Repositories struct {
-	ExecutorGetter                ExecutorGetter
-	FirebaseTokenRepository       FireBaseTokenRepository
-	MarbleJwtRepository           func() MarbleJwtRepository
-	UserRepository                UserRepository
-	OrganizationRepository        OrganizationRepository
-	IngestionRepository           IngestionRepository
-	DataModelRepository           DataModelRepository
-	IngestedDataReadRepository    IngestedDataReadRepository
-	DecisionRepository            DecisionRepository
-	MarbleDbRepository            MarbleDbRepository
-	ClientDbRepository            ClientDbRepository
-	ScenarioPublicationRepository ScenarioPublicationRepository
-	OrganizationSchemaRepository  OrganizationSchemaRepository
-	AwsS3Repository               AwsS3Repository
-	GcsRepository                 GcsRepository
-	CustomListRepository          CustomListRepository
-	UploadLogRepository           UploadLogRepository
-	MarbleAnalyticsRepository     MarbleAnalyticsRepository
+	ExecutorGetter                    ExecutorGetter
+	FirebaseTokenRepository           FireBaseTokenRepository
+	MarbleJwtRepository               func() MarbleJwtRepository
+	UserRepository                    UserRepository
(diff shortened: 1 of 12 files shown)


repositories/repositories.go
```diff
@@ -66,9 +69,13 @@ func NewRepositories(
 		CustomListRepository:          &CustomListRepositoryPostgresql{},
 		UploadLogRepository:           &UploadLogRepositoryImpl{},
 		AwsS3Repository:               AwsS3Repository{s3Client: NewS3Client()},
-		GcsRepository:                 &GcsRepositoryImpl{},
+		GcsRepository:                 &gcsRepository,
 		MarbleAnalyticsRepository: MarbleAnalyticsRepository{
 			metabase: metabase,
 		},
+		TransferCheckEnrichmentRepository: NewTransferCheckEnrichmentRepository(
```
Pascal-Delange, line 76: Now, I'm not 100% happy with this, because we inject a repository (gcsRepository) into another. The issue is that with this, it's harder to do like we do on the ingestion usecase where we allow the option to use a "fake" gcs repository (with local file system) rather than the real GCS repo connecting to GCP - because that should be decided at the usecase level and I don't really want to create a "fakeGcsTransfercherckEnrichmentRepo".
But it's fine for now I guess.

repositories/transfercheck_enrichment.go
```diff
@@ -2,24 +2,248 @@ package repositories
...
+		} else if range1.ipRange.Addr().Less(range2.ipRange.Addr()) {
+			return -1
+		}
+		return 1
+	})
+
+	return nil
+}
+
+func (r *TransferCheckEnrichmentRepository) GetSenderBicRiskLevel(ctx context.Context, bic string) (string, error) {
```
Pascal-Delange, line 247: For this one: pending discussion with Arnaud on what to do with it. But we can already go ahead with the current scope of the PR.

balzdur (approved):
Reminder: refactor lock for lazy initialisation

PR #586, opened 2024-06-12
fix(api): harmonize custom list return

balzdur:
Small breaking change to harmonize api return.

NB: I checked the frontend repo, the return type is not used. Indeed, the repository implementation is **wrong** so it would not work if used.


Pascal-Delange (approved):
👍

PR #589, opened 2024-06-13
feat(batch-ingestion): allow org from param for batch ingestion

balzdur:


Pascal-Delange (approved):
got it 👍 
I'm curious though how API calls for other endpoints work from the BO ? They should al have the same issue ? Do they all make use of OrgIdFromRequest ?

balzdur:
> got it 👍 I'm curious though how API calls for other endpoints work from the BO ? They should al have the same issue ? Do they all make use of OrgIdFromRequest ?

Yes, most of API calls inject the orgId in querry param

PR #598, opened 2024-06-19
required sender account type for transfercheck

Pascal-Delange:
Remains to do: define the exact list of values we want to use: https://checkmarble.slack.com/archives/C06Q995LQCB/p1718790927395749?thread_ts=1718790335.969849&cid=C06Q995LQCB
=> done

---
I added the field in my local transfercheck org and the staging one (our staging) and the sandbox in our prod env one. Remember to add it in your local org if you have one.
I also updated the [org setup guide](https://www.notion.so/checkmarble/Org-setup-guide-984e368f113e434f97f42aaa6424e60c) with it

### DIFF
--- a/dto/transfer_check.go
+++ b/dto/transfer_check.go
@@ -32,6 +32,7 @@ type TransferData struct {
 	Currency            string     `json:"currency"`
 	Label               string     `json:"label"`
 	SenderAccountId     string     `json:"sender_account_id"`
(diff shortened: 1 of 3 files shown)


Pascal-Delange:
Oh I also need to add it to the api doc, I'll do that tonight or tomorrow

balzdur (approved):

PR #601, opened 2024-06-21
feat(license): add license endpoints

balzdur:
Introduce License endpoints in the backend :
- Manage License (~CRUD) when you are a MARBLE ADMIN
- Validate a given license key: public endpoint

NB: I do not know if GCP provide an "anti spam" strategy by default on endpoints, to prevent people spamming the endpoint to find valid licenses (like rate limitting)


api/routes.go
```diff
@@ -156,4 +156,10 @@ func (api *API) routes(auth *Authentication, tokenHandler *TokenHandler) {
 	router.PATCH("/transfers/:transfer_id", api.handleUpdateTransfer)
 	router.GET("/transfers/:transfer_id", api.handleGetTransfer)
 	router.POST("/transfers/:transfer_id/score", api.handleScoreTransfer)
+
+	router.GET("/licenses", api.handleListLicenses)
+	router.POST("/licenses", api.handleCreateLicense)
+	router.PATCH("/licenses/:license_id", api.handleUpdateLicense)
+	router.GET("/licenses/:license_id", api.handleGetLicenseById)
+	router.GET("/validate-license/:license_key", api.handleValidateLicense)
```
Pascal-Delange, line 164: certainly this one should be using the base router (as line 32), not the routed with auth middleware ?

usecases/license_usecase.go
```diff
@@ -0,0 +1,150 @@
...
+				ExpirationDate:      input.ExpirationDate,
+				OrganizationName:    input.OrganizationName,
+				Description:         input.Description,
+				LicenseEntitlements: input.LicenseEntitlements,
+			},
+		)
+		if err != nil {
+			return models.License{}, err
+		}
+		return usecase.licenseRepository.GetLicenseById(ctx, exec, licenseId)
```
Pascal-Delange, line 84:
```suggestion
	return executor_factory.TransactionReturnValue(ctx, usecase.transactionFactory, func(
		tx repositories.Executor,
	) (models.License, error) {
		licenseId := uuid.NewString()
		err := usecase.licenseRepository.CreateLicense(
			ctx,
			tx,
			models.License{
				Id:                  licenseId,
				Key:                 generateLicense(),
				ExpirationDate:      input.ExpirationDate,
				OrganizationName:    input.OrganizationName,
				Description:         input.Description,
				LicenseEntitlements: input.LicenseEntitlements,
			},
		)
		if err != nil {
			return models.License{}, err
		}
		return usecase.licenseRepository.GetLicenseById(ctx, tx, licenseId)
```

Pascal-Delange (approved):
👍 

No completely obvious way I know of to magically rate limit a specific endpoint for such attacks (how would it distinguish from legit calls ?), but we could look into  Cloud Armor configuration.
Anyway they're not going to bruteforce find a 32 byte key, so that's ok IMO.

PR #604, opened 2024-06-26
better logging in gcp env: log trace ids & spanid in logs so that we can correlate logs

Pascal-Delange:
## Inspiration
 [this](https://github.com/jussi-kalliokoski/slogdriver/blob/main/handler.go)

Also see [this](https://cloud.google.com/trace/docs/trace-log-integration#:~:text=To%20associate%20a%20log%20entry%20with%20a%20trace,project%20ID%20and%20TRACE_ID%20is%20the%20trace%20identifier)

## Requirements
~This means I need to put back the env var in cloud run~ or maybe not, see second commit

Goes with https://github.com/checkmarble/terraform/pull/27

### DIFF
--- a/utils/context_logging.go
+++ b/utils/context_logging.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 
 	"github.com/gin-gonic/gin"
+	"golang.org/x/oauth2/google"
 )
 
 func NewLogger(format string) *slog.Logger {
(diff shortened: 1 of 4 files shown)


utils/context_logging.go
```diff
@@ -27,9 +27,13 @@ func NewLogger(format string) *slog.Logger {
 		}.NewLocalDevHandler(os.Stdout)
 		logger = slog.New(logHandler)
 	case "json":
-		slogOption := slog.HandlerOptions{ReplaceAttr: GCPLoggerAttributeReplacer}
-		jsonHandler := slog.NewJSONHandler(os.Stdout, &slogOption)
-		logger = slog.New(jsonHandler)
+		logger = slog.New(slog.NewJSONHandler(os.Stdout, nil))
+	case "gcp":
+		creds, err := google.FindDefaultCredentials(context.Background())
```
Pascal-Delange, line 32: I want to try this, the thing is that it does not work in local dev mode (you get credentials but they don't have a project id attached (though they have a quota project id but that would be hacky)
but that's fine because this only makes sense in a gcp setup anyway

balzdur (approved):

PR #610, opened 2024-06-27
add bic field on partner with related CRUD operations

Pascal-Delange:
The first commit is a bugfix: a new key was generated at every server startup. This probably has caused some unwanted 401s in the front.

---- 
The BIC is supposed to be the 8 char bic, but I'm not strictly enforcing this for now

### DIFF
--- a/repositories/transfercheck_partners.go
+++ b/repositories/transfercheck_partners.go
@@ -2,6 +2,7 @@ package repositories
 
 import (
 	"context"
+	"strings"
 
 	"github.com/Masterminds/squirrel"
 	"github.com/checkmarble/marble-backend/models"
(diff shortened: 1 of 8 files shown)


usecases/partner_usecase.go
```diff
@@ -63,6 +65,7 @@ func (usecase *PartnerUsecase) CreatePartner(
 		tx repositories.Executor,
 	) (models.Partner, error) {
 		partnerId := uuid.New().String()
+		partnerCreateInput.Bic = strings.TrimSpace(strings.ToUpper(partnerCreateInput.Bic))
```
Pascal-Delange, line 68: For now and since it's an internal endpoint, I'm doing minimal data cleaning rather than full input validation

repositories/transfercheck_partners.go
```diff
@@ -14,15 +15,20 @@ func selectPartners() squirrel.SelectBuilder {
...
-func (repo MarbleDbRepository) ListPartners(ctx context.Context, exec Executor) ([]models.Partner, error) {
+func (repo MarbleDbRepository) ListPartners(ctx context.Context, exec Executor, filters models.PartnerFilters) ([]models.Partner, error) {
 	if err := validateMarbleDbExecutor(exec); err != nil {
 		return nil, err
 	}
 
+	query := selectPartners().OrderBy("created_at DESC")
+	if filters.Bic.Valid {
+		query = query.Where(squirrel.Eq{"LOWER(bic)": strings.ToLower(filters.Bic.String)})
+	}
```
balzdur, line 26: Here bic is "lowered" but on create + update it is not.
I know the index is created using "lower", my question: does this "where" use the index or the stored "non lowered" value in the column ?

balzdur (approved):

repositories/transfercheck_partners.go
Pascal-Delange: Good question :-) 
The answer you're looking for is that since I created the index on a computed field (not on the raw column value), it should use the index to apply the filter.
The slightly longer answer is that because the partners table is small, and is expected to remain kinda small, it's possible that the DB still decides "fuck the index I can scan 10 rows" (especially if you SELECT fields that are not in the index). 
I still created the index for the sake of completeness.

Pascal-Delange: that being said, you're right it's a bit dumb that I UPPER the bic on insertion and LOWER it on filtering (though technically it works just the same). I'll change that.

PR #612, opened 2024-07-02
add return value: "beneficiary is in network" for transfercheck transfers

Pascal-Delange:


usecases/transfer_check_usecase.go
```diff
@@ -471,7 +499,8 @@ func (usecase *TransferCheckUsecase) ScoreTransfer(
 				Data:      transferData.ToIngestionMap(transferMapping),
 				TableName: models.TransferCheckTable,
 			},
-			OrganizationId: organizationId,
+			OrganizationId:     organizationId,
+			TriggerObjectTable: models.TransferCheckTable,
```
Pascal-Delange, line 503: This is a bugfix, beyond the original intent of the PR

balzdur (approved):
LGTM !
I only have an independent question: why the `QueryTransfers` use case return 0 or only the first element and not the whole list ? Is it because only 1 elem max is available ?

Pascal-Delange:
> LGTM ! I only have an independent question: why the `QueryTransfers` use case return 0 or only the first element and not the whole list ? Is it because only 1 elem max is available ?

Yes, because by construction it's "query by the client's unique id" (in practice our "object_id"), so it's a 0 or 1 case

PR #615, opened 2024-07-08
refactor(repositories): add options pattern to NewRepositories

balzdur:
In this PR, a refac to fix the issue with local use of fake GCS repository :
- need to handle fake GCS repository in `NewRepositories`
- refactor interface to use the `With*` options pattern, so we no longer need to explicitly assign "null" values for non used options in init phase

### DIFF
--- a/repositories/repositories.go
+++ b/repositories/repositories.go
@@ -7,6 +7,47 @@ import (
 	"github.com/jackc/pgx/v5/pgxpool"
 )
 
+type options struct {
+	firebaseClient                *auth.Client
+	metabase                      Metabase
+	transfercheckEnrichmentBucket string
+	fakeGcsRepository             bool
+}
+
+type Option func(*options)
+
+func getOptions(opts []Option) *options {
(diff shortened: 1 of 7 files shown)


repositories/repositories.go
```diff
@@ -32,18 +80,21 @@ func NewQueryBuilder() squirrel.StatementBuilderType {
...
 }
 
-func NewRepositories(
-	firebaseClient *auth.Client,
-	marbleConnectionPool *pgxpool.Pool,
-	metabase Metabase,
-	transfercheckEnrichmentBucket string,
-) Repositories {
-	executorGetter := NewExecutorGetter(marbleConnectionPool)
+func NewRepositories(opts ...Option) Repositories {
```
Pascal-Delange, line 83: Since the db pool seems to always be a necessary option, let's keep it as a mandatory param (and the optional ones like you defined here) ?

Pascal-Delange (approved):
nice!

PR #616, opened 2024-07-09
feat(webhooks): basic convoy integration

balzdur:
# Implement webhooks with Convoy

## Setup

Setup intruction are for now defined in Notion [here](https://www.notion.so/checkmarble/Webhooks-features-77d00705704543a99a52df38f741f081?pvs=4#f9e9c96c9f574e338411156910c803a8).

> It could be improved by defining our own docker compose file + env file for Convoy but this works for now, and it is easier to update (since we use Convoy repository directly)

## Archi

Convoy is defined in the `infra` pkg. I created : 
- a webhook DB + repository to manage events creation (= events to be sent to Convoy or whatever webhook solution we use)
- a convoy repository to send events to convoy

The usecase expose :
- `CreateWebhook`: used by other usecases in order to create event in our DB + try to send it to Convoy in a separate go routine
- `SendWebhooks`: used by a dedicated cron job to send all pending events every minute

## Security

> TL;DR: I think WEBHOOK permissions are useless but the org/partner access is a good safety check

I tried to add some permissions, but since every users may at some point interact with a usecase that may produce webhooks, it looks like every single user may need this permission.

Concerning Convoy, multi-tenancy like implementation is done using `owner_id` property. I crafted them following this pattern :
```
# For Marble core users
org:123456

# For Transfercheck users
org:123456-partner:123456
```

In the `enforceSecurity` usecase implementation I did some basic org + partner access check to ensure the current user create a webhook for it's `owner_id` only. In order to allow the cron job to send all pending events for every (org,partner), I created `ANY_PARTNER_ID_IN_CONTEXT` permission (= same behaviour as `ANY_ORGANIZATION_ID_IN_CONTEXT`).

## Business logic

For now, webhooks events state machine follow this pattern :

```mermaid
graph TD;
    Scheduled-->Success;
    Scheduled-->Retry;
    Retry-->Success;
    Retry--"send_attempt_count < 3"-->Retry;
    Retry--"send_attempt_count >= 3"-->Failed
```

The link Marble webhook --> Convoy is done using our internal ID (it's used as idempotency key in Convoy).
When an event is successfully fan out to, Convoy internal logic do the following :
1. If the provided `owner_id` do have endpoints configured for the provided `event_type`, Convoy dispatch the event
2. If not, nothing happen

I tested locally, if I create a matching endpoint, past events are not sent to it.


models/webhook.go
```diff
@@ -0,0 +1,101 @@
...
+		return Failed
+	}
+	panic(fmt.Errorf("unknown webhook delivery status: %s", s))
+}
+
+type WebhookEventType int
+
+const (
+	WebhookEventType_CaseStatusUpdated WebhookEventType = iota
+)
```
balzdur, line 55: I started to experiment missing namespace pain here (as `CaseStatusUpdated` is a "global" constant defined in `models/cases.go` already).

repositories/convoy_repository.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/checkmarble/marble-backend/models"
+	convoy "github.com/frain-dev/convoy-go/v2"
+)
+
+type convoyResources interface {
+	GetClient() *convoy.Client
+}
```
balzdur, line 14: For now, instead of consumming directly `*convoy.Client` (as Firebase, Metabase do) I inject a "getter" interface.

IMO, it will help define a dynamic "getter" if required in the future (eg, separate client per application (transfercheck, marble core...) or no client when no license). Still, this is too abstract to fully decide if it's a good choice or not.

balzdur (commented):
What may comes next:
- decide how to handle webhooks for different app
- handle license/unlicense in the backend
- find an way to manage webhooks (inside our product or using Convoy Portal Links)

usecases/webhooks_usecase.go
```diff
@@ -0,0 +1,224 @@
...
+					EventType:        input.EventType,
+					EventData:        input.EventData,
+				})
+			if err != nil {
+				return models.Webhook{}, fmt.Errorf("error while creating webhook: %w", err)
+			}
+			return usecase.webhookRepository.GetWebhook(ctx, tx, webhookId)
+		})
+
+	go func() {
```
Pascal-Delange, line 106: Ok I'm thinking this through again.
We would like to insert the webhook in the same transaction in which we do the insert/update.
Once it's committed (and not before), we want to send the webhook to convoy, doing any necessary bookkeeping of the webhook stored on our side. So in practice it's not clear to me that we can or should call `sendWebhooks` from within this function - as the transaction would not yet be committed.

usecases/webhooks_usecase.go
```diff
@@ -0,0 +1,224 @@
...
+	return nil
+}
+
+func (usecase WebhooksUsecase) SendWebhooks(
+	ctx context.Context,
+) error {
+	logger := utils.LoggerFromContext(ctx)
+	exec := usecase.executorFactory.NewExecutor()
+
+	pendingWebhooks, err := usecase.webhookRepository.ListWebhooks(ctx, exec, models.WebhookFilters{
```
Pascal-Delange, line 123: See comment in the repository: this should be bounded in size

usecases/webhooks_usecase.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	logger.InfoContext(ctx, fmt.Sprintf("Webhooks sent: %d success, %d retry, %d failed, %d errors",
+		successCount, retryCount, failedCount, errorCount))
+
+	return firstError
+}
+
+// sendWebhook sends a webhook and updates its status in the database.
+// It returns the delivery status of the webhook and an error if updating the webhook fails.
+func (usecase *WebhooksUsecase) sendWebhook(
```
Pascal-Delange, line 190: This function should do what it does in a transaction. We can discuss the details if you want.

Pascal-Delange (commented):
Thanks, great first version !
There are a few tricky points left on how the actions are orchestrated, in particular from the point of view of DB transactions.
It's perhaps made yet a bit more tricky if we consider that (for batch decisions at least) the decisions are written in batch in a transaction, and hence so should the webhooks (though sending them, probably should not?)
Let's discuss it in the afternoon if you want

cmd/run_job_scheduler.go
```diff
@@ -68,6 +73,7 @@ func RunJobScheduler() error {
 
 	repositories := repositories.NewRepositories(pool,
 		repositories.WithFakeGcsRepository(gcpConfig.FakeGcsRepository),
+		repositories.WithConvoyResources(infra.InitializeConvoyRessources(convoyConfiguration)),
```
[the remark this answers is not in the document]
balzdur: Yes, still the code may break during `convoyConfiguration` init since we do use `utils.GetEnv(...)`

balzdur (commented):
I pushed a new commit with changes based on your comments. For now I only do a single "batch" webhook send (no loop until we reached the end of pending webhooks)

infra/convoy.go
```diff
@@ -0,0 +1,25 @@
...
+		config.ProjectID,
+	)
+
+	return ConvoyRessources{
+		convoyClient: convoyClient,
+	}
+}
+
+func (r ConvoyRessources) GetClient() (*convoy.Client, error) {
+	return r.convoyClient, nil
```
Pascal-Delange, line 24: the idea was to return an error here if the client is nil, corresponding to what you were previously doing in the repository :-)

repositories/migrations/20240708141735_create_webhook_table.sql
```diff
@@ -0,0 +1,29 @@
...
+            CONSTRAINT fk_webhooks_partner FOREIGN KEY(partner_id) REFERENCES partners(id) ON DELETE CASCADE
+      );
+
+CREATE INDEX webhooks_delivery_status_idx ON webhooks(delivery_status) WHERE delivery_status IN ('scheduled', 'retry');
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
```
Pascal-Delange, line 26: Also drop index (first)

Pascal-Delange (approved):
Looks good (a last few comments).
So in this first version, webhooks are only sent by the batch job (and we'll add near-sync sending & scaling options of the batch job in a 2nd time) correct ?

usecases/webhooks_usecase.go
balzdur: I tried to change for a tx implementation. Let's try this way and if transaction make webhook bulk send slow, we can remove it by leveradging the idempotency key.

PR #619, opened 2024-07-11
Thomas/use generated convoy client

balzdur:
In this PR :
- Generate a Convoy API Client based on the OAS spec
- Remove `convoy-go` SDK

Bonnus: I only generate a whitelist of endpoints (+ depending models) so we are only exposed to the pertinent subset of endpoints we use (I expect it will make diff changes in API update more readable for us, by ditching any noise on other endpoints)


Makefile
```diff
@@ -0,0 +1,9 @@
+help: ## Display this help
```
Pascal-Delange, line 1: here we are, we have a makefile now :D

Pascal-Delange (approved):
it's only the client codegen part that I handn't read yet, right?

PR #620, opened 2024-07-11
Thomas/manage webhooks

balzdur:
First iteration for webhook creation by API.

NB: still discovered some issues in the Open API doc... but fixable easy. I also openned a PR to fix issues on convoy repository


models/webhook.go
```diff
@@ -110,3 +80,26 @@ func (f WebhookEventFilters) MergeWithDefaults() WebhookEventFilters {
 	}
 	return defaultFilters
 }
+
+type WebhookCreate struct {
+	OrganizationId    string
+	PartnerId         null.String
+	EventType         WebhookEventType
```
Pascal-Delange, line 87: as discussed orally: the PR looks globally good to me, the one thing I'm unsure about is I'd rather see a mapping 1 registered webhook <-> list of event types

PR #622, opened 2024-07-12
Thomas/manage webhooks

balzdur:
In this PR, improved work on webhook creation (with your previous comments)

NB: you may review changes for the second time due to the last PR auto-closing... I didn't rebase + merge already reviewed changes and I must admit I did some rebase + commit amend 🙈


usecases/case_usecase.go
```diff
@@ -304,6 +304,15 @@ func (usecase *CaseUseCase) UpdateCase(ctx context.Context, userId string,
...
 }
 
+func webhookEventCaseUpdatedFrom(updateCaseAttributes models.UpdateCaseAttributes) models.WebhookEventContent {
+	if updateCaseAttributes.Status != "" {
+		return models.WebhookEventCaseStatusUpdated{
+			CaseStatus: updateCaseAttributes.Status,
+		}
+	}
+	return nil
+}
```
balzdur, line 314: Experimentation : use an interface to handel Webhook Event content and create a struct for each custom event

balzdur: My first idea was to direclty implement interface on struct like `updateCaseAttributes models.UpdateCaseAttributes` here, but I'm not totally sure it's a good idea...

Another alternative would be to rollback to the monomorphic event data struct + add suggar functions to create each event (to prevent inline def of `event_data`)

Pascal-Delange: yes the second option makes sense to me

models/webhook.go
```diff
@@ -110,3 +91,36 @@ func (f WebhookEventFilters) MergeWithDefaults() WebhookEventFilters {
...
+	Secret            string
+	Url               string
+	HttpTimeout       *int
+	RateLimit         *int
+	RateLimitDuration *int
+}
+
+func (input WebhookRegister) Validate() error {
+	for _, eventType := range input.EventTypes {
+		if _, err := ParseWebhookEventType(eventType); err != nil {
```
Pascal-Delange, line 108: unless ParseWebhookEventType is used somewhere else (I don't think so?), you can just replace this with something like 
```go
if !slices.Contains(validWebhookEventTypes, WebhookEventType(eventType)) {...
```

repositories/convoy_repository.go
```diff
@@ -27,23 +28,89 @@ func getOwnerId(organizationId string, partnerId null.String) string {
 	return fmt.Sprintf("org:%s", organizationId)
 }
 
+// Ensure event type is included in the event data
+func getEventContent(eventContent models.WebhookEventContent) (string, map[string]interface{}) {
+	eventType := string(eventContent.Type)
+	eventData := eventContent.Data
+	eventData["event_type"] = eventType
+	return eventType, eventData
```
Pascal-Delange, line 36: so, just to have it all clear in my head: the expectation with convoy (and other solutions) is that we send a map with the event type and id of the object at the root level, along with (optionally) the event body that is nested, and any relevant metadata too if we want ?

Pascal-Delange: if all the function does is effectively to add the type to the event data, let's rename it to something along the lines of "addEventTypeToData" and make it return just the map ? (and use the original webhookContent type where we need it)

repositories/convoy_repository.go
```diff
@@ -27,23 +28,89 @@ func getOwnerId(organizationId string, partnerId null.String) string {
...
+	fanoutEvent, err := convoyClient.CreateEndpointFanoutEventWithResponse(ctx, projectId, convoy.ModelsFanoutEvent{
+		OwnerId:        &ownerId,
+		EventType:      &eventType,
+		IdempotencyKey: &webhookEvent.Id,
+		Data:           &eventData,
+	})
+	if err != nil {
+		return errors.Wrap(err, "can't create convoy event: request error")
+	}
+	if fanoutEvent.JSON201 == nil {
```
Pascal-Delange, line 58: question: how redundant is this with the check of line 55 ? Do we expect a non nil err every time there is no JSON201 ?

repositories/convoy_repository.go
```diff
@@ -27,23 +28,89 @@ func getOwnerId(organizationId string, partnerId null.String) string {
...
+		Url:               &input.Url,
+		Secret:            &input.Secret,
+		HttpTimeout:       input.HttpTimeout,
+		RateLimit:         input.RateLimit,
+		RateLimitDuration: input.RateLimitDuration,
+	})
+	if err != nil {
+		return errors.Wrap(err, "can't create convoy endpoint: request error")
+	}
+	if endpoint.JSON201 == nil {
```
Pascal-Delange, line 92: same here

Pascal-Delange (commented):
thanks ! 
Two small last points, almost good to merge

PR #623, opened 2024-07-12
feat(webhooks): list all webhooks

balzdur:
NB: Full required endpoints are not yet done but I want you to share your opinion

### DIFF
--- a/repositories/convoy_repository.go
+++ b/repositories/convoy_repository.go
@@ -29,6 +29,22 @@ func getOwnerId(organizationId string, partnerId null.String) string {
 	return fmt.Sprintf("org:%s", organizationId)
 }
 
+func parseOwnerId(ownerId string) (string, null.String) {
+	parts := strings.Split(ownerId, "-partner:")
+	if len(parts) == 2 {
+		return parts[0][4:], null.StringFrom(parts[1])
+	}
+	return ownerId[4:], null.String{}
+}
+
+func getName(ownerId string, eventTypes []string) string {
+	eventLabel := "all-events"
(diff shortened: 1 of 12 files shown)


repositories/convoy_repository.go
```diff
@@ -114,3 +114,93 @@ func (repo ConvoyRepository) RegisterWebhook(ctx context.Context, input models.W
...
+		convoyEndpoint, ok := endpointMap[*convoySubscription.EndpointMetadata.Uid]
+		if !ok {
+			return nil, errors.New("can't find convoy endpoint")
+		}
+
+		webhooks = append(webhooks, adaptWebhook(convoyEndpoint, convoySubscription))
+	}
+
+	return webhooks, nil
+}
```
balzdur, line 167: Due to the opinionated way we handle webhook on Convoy (create imediately a pair endpoint/subscription) I need to fetch and reconcile both entities on Convoy (to return something close to what we expect for webhook register action).

An alternative would be :
- `LIST` : fetch only endpoints and display part of the data (= in a list view) 
- `GET /:id` : get full info per endpoint  (= with dedicated detail view)

Still, I think we can live with that, assuming we won't create too many webhooks in the near future (-> that's why I hardcoded a 100 page size and do not bother with pagination)

Pascal-Delange: I can live with the setup where we retrieve both pieces of information and merge them.
We'll see in due time if it ever becomes an issue.
For good measure, let's use a larger page size and/or log a warning if nb of items retrieved == page size ?

Pascal-Delange (approved):

PR #624, opened 2024-07-15
rework webhooks event usecase (sync with semaphore)

Pascal-Delange:
- send the webhooks asynchronously after creating the event
   - in practice this now requires to call 2 methods in the usecase where we want to send the webhook - but I think it's hard to avoid that
- the batch job runs every 10 minutes to retry the failed ones
- stop writing the "failed" status after 3 failed retries (basically I think this kind of status update isn't really suited in the webhook sending routine, especially because the frequency at which retries are run is an externality unknown from the sending routine. Most likely we want to apply relevant filters in the query that reads events to retry, though ideally we're still aiming for a "100% webhooks eventually reached convoy")
- inject page size of events to retry from env variable (optionally): if we end up with a backlog because of a failure of convoy, we want to be able to increase the size to catch up with the backlog

---- 
Things I want to do next:
- make the proper content of webhooks (with object id) when the event happens
- send the original event timestamp along with the base event metadata

--- 
The PR is probably best read commit by commit at first


usecases/webhook_events_usecase.go
```diff
@@ -108,102 +141,98 @@ func (usecase WebhookEventsUsecase) SendWebhookEvents(
...
+
+	err = usecase.enforceSecurity.SendWebhookEvent(ctx, webhookEvent.OrganizationId, webhookEvent.PartnerId)
+	if err != nil {
+		return models.Scheduled, err
+	}
+
+	logger.InfoContext(ctx, fmt.Sprintf("Start processing webhook event %s", webhookEvent.Id))
+
+	err = usecase.convoyRepository.SendWebhookEvent(ctx, webhookEvent)
+	// sending the webhook to convoy was successful, so we can (no need for transaction)
```
balzdur, line 205:
```suggestion
	// sending the webhook to convoy was successful, so we can update the status (no need for transaction). If update fail, idempotency on convoy will prevent sending an event twice)
```

usecases/webhook_events_usecase.go
```diff
@@ -77,28 +90,48 @@ func (usecase WebhookEventsUsecase) CreateWebhookEvent(
...
+			trace.WithAttributes(attribute.String("webhook_event_id", webhookEventId)))
+		defer span.End()
+
+		status, err := usecase._sendWebhookEvent(ctx, webhookEventId)
+		if err != nil {
+			logger.ErrorContext(ctx, fmt.Sprintf("Error sending webhook event %s: %s", webhookEventId, err.Error()))
+		}
+		if status == models.Retry {
+			logger.WarnContext(ctx, fmt.Sprintf("Webhook event %s will be retried", webhookEventId))
+		}
```
balzdur, line 122: IMO, not really accurate: normally we do not use returned value when `err != nil`, here by reading the `_sendWebhookEvent` implementation, there is some branching where the returned status may not be the one currently present in the DB

Maybe use a `else` or add a `return` in previous `if`

usecases/webhook_events_usecase.go
```diff
@@ -108,102 +141,98 @@ func (usecase WebhookEventsUsecase) SendWebhookEvents(
...
+			case <-ctx.Done():
+				return errors.Wrapf(ctx.Err(), "context cancelled before retrying webhook %s", webhookEvent.Id)
+			default:
+			}
 
-	waitGroup.Wait()
-	close(uploadErrorChan)
-	close(deliveryStatusChan)
+			deliveryStatus, err := usecase._sendWebhookEvent(ctx, webhookEvent.Id)
+			deliveryStatuses[i] = deliveryStatus
```
balzdur, line 162: Same here, we may use some "invalid" deliveryStatus

balzdur (approved):
LGTM 👍🏽 

Just few comments :
- If `Failed` status is not used, we should remove it totally (it's easier to re-add if required in the future)
- In the DB, the "count" column is `SendAttemptCount` (= number of attempt which is different from number of "retry"): with the current business logic, we do not increment it on Success. Either change the business logic or change the db schema.

usecases/webhook_events_usecase.go
Pascal-Delange: oopsies, thanks

usecases/webhook_events_usecase.go
Pascal-Delange: Ok, thanks. Took me a while to understand your point but i'm there now.
Yes, it is correct that I'm breaking a convention here - partly because I'm mixing up the error from sending the webhook and the error from udpating the webhook event status in db

usecases/webhook_events_usecase.go
Pascal-Delange: Well, here in fact we don't, at least we don't use it: any error returned will return an error to the waitgroup (line 167 `err = group.Wait()`) after which we don't use the deliveryStatuses slice.
Still I get your point.

Pascal-Delange:
Ok I'll rename SendAttemptCount into RetryCount (with the provision that we will never have a guarantee that it's in proper sync, but that's ok it's indicative)

PR #625, opened 2024-07-16
refactor(webhooks): refactor webhook dtos

balzdur:
Small refactoring to be closer to other resources exposed on the API :
- POST and PATCH should return data
- Only return webhook secret on POST (closer to what we have on API_KEYS, it will help in case we want to implement finer permission strategy)

NB: currently there is no way to retrive an existing webhook secret but it's possible (the secret is stripped off at the dto adapter layer). I let this work for another iteration, focused on secret management.


Pascal-Delange (approved):
that makes sense, thanks 👍

PR #626, opened 2024-07-17
remove go-fuzzywuzzy dependency

Pascal-Delange:
Library replaced for licensing reasons. https://github.com/adrg/strutil which comes as a partial replacement (for the edit distance) is under MIT license.

--- 

Last thing to do (but no strong urgency): update the naming of the functions in the front and the doc. I changed "token_set_ratio" it to "bag_of_words_similarity" because it's less library specific and ultimately more self-explanatory, though I kept backwards compatibility of course for old ASTs.

--- 

Making if fully backwards compatible was a bit tricky because of subtle differences in how the Levensthtein distance is defined (and computed), see comment in the strings.go file

### DIFF
--- a/usecases/ast_eval/evaluate/eval_fuzzy_match.go
+++ b/usecases/ast_eval/evaluate/eval_fuzzy_match.go
@@ -4,19 +4,16 @@ import (
 	"context"
 
 	"github.com/cockroachdb/errors"
-	fuzzy "github.com/paul-mannino/go-fuzzywuzzy"
 
 	"github.com/checkmarble/marble-backend/models/ast"
 	"github.com/checkmarble/marble-backend/pure_utils"
 )
 
-// Implements a fuzzy match using the go-fuzzywuzzy library.
-// List of strign cleaning steps applied:
+// List of string cleaning steps applied:
 // - normalize
 // - remove diacritics
 // - set to lower case
-// - keep only letters and numbers
-// (- keep non-ASCII characters)
+// - only letters and numbers
 type FuzzyMatch struct{}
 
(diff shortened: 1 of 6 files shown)


usecases/ast_eval/evaluate/eval_fuzzy_match.go
```diff
@@ -4,19 +4,17 @@ import (
...
 	"github.com/cockroachdb/errors"
-	fuzzy "github.com/paul-mannino/go-fuzzywuzzy"
 
 	"github.com/checkmarble/marble-backend/models/ast"
 	"github.com/checkmarble/marble-backend/pure_utils"
 )
 
-// Implements a fuzzy match using the go-fuzzywuzzy library.
-// List of strign cleaning steps applied:
+// Implements a fuzzy match using
```
balzdur, line 12:
```suggestion
```

usecases/ast_eval/evaluate/eval_fuzzy_match.go
```diff
@@ -69,31 +64,25 @@ func (fuzzyMatcher FuzzyMatchAnyOf) Evaluate(ctx context.Context, arguments ast.
...
+
 	switch s {
 	case "ratio":
-		f = func(s1 string, s2 string, opts ...bool) int { return fuzzy.Ratio(s1, s2) }
-	case "partial_ratio":
-		f = func(s1 string, s2 string, opts ...bool) int { return fuzzy.PartialRatio(s1, s2) }
-	case "token_sort_ratio":
-		f = fuzzy.TokenSortRatio
+		return pure_utils.DirectSimilarity, nil
 	case "token_set_ratio":
```
balzdur, line 81: IMO, we should add acomment to explain to future "us" (or devs) why this case exists

balzdur (approved):

PR #627, opened 2024-07-18
send webhooks on case updates

Pascal-Delange:
- send webhooks on every case update + decision created
- revamp webhook format: final format is 
```json
{
  "content": {
    "case": {
      "id": "3ef6392f-acb1-4aab-ac1d-d8353f225363"
    }
  },
  "timestamp": "2024-07-19T11:15:01.092728+02:00",
  "type": "case.decisions_updated"
}   
```
inspired from the format of webhooks in chargebee, which I worked with and were pretty nice to work on.
In particular it leaves the door open to easily send more details on the object (here the case) or to send other objects alongside it without having the change the existing format. + we can add more metadata where needed.
- current list of webhooks treated:

> case.updated
> case.created_manually
> case.created_from_workflow
> case.decisions_updated
> case.tags_updated
> case.comment_created
> case.file_created
> decision.created


usecases/decision_usecase.go
```diff
@@ -263,13 +269,39 @@ func (usecase *DecisionUsecase) CreateDecision(
...
+		}
 
 		return usecase.decisionRepository.DecisionWithRuleExecutionsById(ctx, tx, decision.DecisionId)
 	})
+
+	for _, webhookEventId := range sendWebhookEventId {
+		usecase.webhookEventsSender.SendWebhookEventAsync(ctx, webhookEventId)
+	}
+
+	return newDecision, err
```
balzdur, line 304: Here you do not take into account possible transaction error, is it intentionnal ?

usecases/decision_usecase.go
```diff
@@ -375,14 +409,35 @@ func (usecase *DecisionUsecase) CreateAllDecisions(
...
+			}
 		}
 
 		return usecase.decisionRepository.DecisionsWithRuleExecutionsByIds(ctx, tx, ids)
 	})
+
+	for _, caseWebhookEventId := range sendWebhookEventIds {
+		usecase.webhookEventsSender.SendWebhookEventAsync(ctx, caseWebhookEventId)
+	}
+
```
balzdur, line 440: Idem

balzdur (approved):

usecases/decision_usecase.go
Pascal-Delange: you are right, thanks for pointing it out

balzdur (approved):

PR #637, opened 2024-07-31
models & dto & basic API scaffold for rule snoozes

Pascal-Delange:
See https://www.notion.so/checkmarble/Rule-Snoozing-b9ccd81c197f4816a1ce70d2dfb501b3


balzdur (approved):

usecases/rule_snoozes.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		return models.SnoozesOfDecision{}, err
+	}
+
+	return models.NewSnoozesOfDecision(decisionId, snoozes, it), nil
+}
+
+func (usecase RuleSnoozeUsecase) SnoozeDecision(
+	ctx context.Context, input models.SnoozeDecisionInput,
+) (models.SnoozesOfDecision, error) {
```
balzdur, line 113: It looks like you never return any `models.SnoozesOfDecision`, do we still need those or not ? (impactfull for the return type on the API for me).

In short, I won't use the return value in the frontend so you can return it or not.

balzdur (commented):
As a reminder, you also need to update the license (but it can be done in a future PR)

usecases/rule_snoozes.go
pascaldelange: yes yes, I just commited before this part was finished when I left for China :)

pascaldelange:
Continued in https://github.com/checkmarble/marble-backend/pull/638 for security reasons while I'm working abroad

pascaldelange:
can be closed

PR #639, opened 2024-08-06
models & dto & basic API scaffold for rule snoozes

pascaldelange:
Continued work from https://github.com/checkmarble/marble-backend/pull/637, PR created from a fork on my personal github account

---- 

## Scope:
- Introduce the rule snooze & snooze group models
- allow to snooze a rule for a certain pivot value, from within the context of a case
- allow to read existing snoozes:
   - on all rules of a scenario iteration (existence or not of currently active rule snoozes for each rule)
   - on all rules in the context of a decision - meaning on all rules of the scenario iteration used to create the decision, active at the current time for the decision's pivot id (only if present)
- [Sneaked in] license server DTO for rule snoozes
- updated doc (in-app openapi + readme public api doc, including backfilling some missing fields that had not been added on the doc
- new copies of a rule (as per new draft from iteration, not cloning a rule) inherit the parent rule's snooze groups

Broadly based on the implementation plan discussed on [notion](https://www.notion.so/checkmarble/Rule-Snoozing-b9ccd81c197f4816a1ce70d2dfb501b3), with small changes where necessary.

---- 

## Gotchas & things to keep in mind

- I created a "snooze creator" permission, which I added on the viewer role but the actual permission check in the usecase is on a “by inbox” basis (any member of an inbox can view the cases & act on them, including adding snoozes)
- loosely enforce the “one active snooze per rule” constraint (a stricter enforcement would require serializable tx which is totally feasible but a bit overkill here): we could end up with duplicates, but it's unlikely & no particular bad impact
- I'm reusing the inbox permission check from the case usecase which is injected as a dependency to read the decision’s case
- New field `outcome` on decisions (DB & API). Nb: can be empty string in db for old decisions, migration to be added in a later step if needed, but beware if we add filtering by outcome (option not added for now)
- Detail of snooze reason is not in the API decision object (don’t want to overload it & make querying more complex): the source of truth exists and can be retrieved if needed later
- a snooze group is created and attached to a rule if a snooze is added & no snooze group existed yet for this rule. This is an exception to the rule that we do not mutate rules, but is permissible because it is only additive (unless race conditions, but worst case a snooze is lost so not bug impact)

--- 
## To Do
- [ ] create case event on new snooze created
- [ ] apply doc on readme


usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -114,9 +124,35 @@ func EvalScenario(
...
+	snoozes := make([]models.RuleSnooze, 0)
+	if params.Pivot != nil {
+		snoozeGroupIds := make([]string, 0, len(publishedVersion.Body.Rules))
+		for _, rule := range publishedVersion.Body.Rules {
+			if rule.SnoozeGroupId != nil {
+				snoozeGroupIds = append(snoozeGroupIds, *rule.SnoozeGroupId)
+			}
+		}
+		snoozes, err = repositories.SnoozeReader.ListRuleSnoozesForDecision(ctx, exec, snoozeGroupIds, *pivotValue)
+	}
```
balzdur, line 146: I would do one of the following instead (IIUC, the pivot value can be null, so snooze if statement should depend on it):
```suggestion
	var pivotValue *string
	if params.Pivot != nil {
		pivotValue, err = getPivotValue(ctx, *params.Pivot, dataAccessor)
		if err != nil {
			return models.ScenarioExecution{}, errors.Wrap(
				err,
				"error getting pivot value in EvalScenario")
		}
	}

	snoozes := make([]models.RuleSnooze, 0)
	if pivotValue != nil {
		snoozeGroupIds := make([]string, 0, len(publishedVersion.Body.Rules))
		for _, rule := range publishedVersion.Body.Rules {
			if rule.SnoozeGroupId != nil {
				snoozeGroupIds = append(snoozeGroupIds, *rule.SnoozeGroupId)
			}
		}
		snoozes, err = repositories.SnoozeReader.ListRuleSnoozesForDecision(ctx, exec, snoozeGroupIds, *pivotValue) // ensure *pivotValue won't fail (independant from a change in above if statement
	}
```

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -144,16 +180,10 @@ func EvalScenario(
...
+		PivotValue:          pivotValue,
 	}
 	if params.Pivot != nil {
-		pivotValue, err := getPivotValue(ctx, *params.Pivot, dataAccessor)
-		if err != nil {
-			return models.ScenarioExecution{}, errors.Wrap(
-				err,
-				"error getting pivot value in EvalScenario")
-		}
 		se.PivotId = &params.Pivot.Id
```
balzdur, line 186: I would keep all pivot related stuff closed to each other (for readability)
```suggestion
		se.PivotId = &params.Pivot.Id
		se.PivotValue = pivotValue
```

balzdur (approved):

balzdur:
@pascaldelange, I merge this for now, you can continue your work + take comments into account in a second PR

repositories/dbmodels/db_decision.go
```diff
@@ -97,15 +97,3 @@ func AdaptDecisionWithRank(db DbDecision, decisionCase *models.Case, rankNumber,
...
-	Id             string      `db:"id"`
-	OrganizationId string      `db:"org_id"`
-	DecisionId     string      `db:"decision_id"`
-	Name           string      `db:"name"`
-	Description    string      `db:"description"`
-	ScoreModifier  int         `db:"score_modifier"`
-	Result         bool        `db:"result"`
-	ErrorCode      int         `db:"error_code"`
-	DeletedAt      pgtype.Time `db:"deleted_at"`
-}
```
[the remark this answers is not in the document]
pascaldelange: yes 😬  this one was ugly

repositories/dbmodels/db_decision_rule.go
```diff
@@ -32,6 +33,16 @@ func AdaptRuleExecution(db DbDecisionRule) (models.RuleExecution, error) {
...
+	outcome := db.Outcome
+	if outcome == "" {
+		if db.ErrorCode != 0 {
+			outcome = "error"
+		} else if db.Result {
+			outcome = "hit"
+		} else {
+			outcome = "no_hit"
+		}
+	}
```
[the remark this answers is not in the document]
pascaldelange: yep that's right

usecases/evaluate_scenario/evaluate_scenario.go
pascaldelange: so instead of line 183 you mean ? Sure, can do that

usecases/evaluate_scenario/evaluate_scenario.go
pascaldelange: Thank you, your suggestion was totally my intent, I failed at writing line 138 (and didn't properly test this case locally)

PR #645, opened 2024-08-09
Test rule snoozing

pascaldelange:
- rework integration test suite: stop test if an error is met (instead of continuing & generating lots of noise) + some cleanup
- the whole test suite is starting to be a bit messy, but then again it's pretty efficient in terms of code coverage tested / code written (especially targetting the critical workflows)

--- 

## Update 12/08
It's not perfect yet, but it's starting to look not bad. I think we could rewrite it a bit to make it more modular and avoid the "test all the things in one big function and it becomes intractable", instead testing feature by feature. But I'm rather happy with the level of core logic coverage this gives us (excluding the api layer and TC, that is), so let's first roll with this.

### DIFF
--- a/integration_test/scenario_flow_test.go
+++ b/integration_test/scenario_flow_test.go
@@ -3,8 +3,6 @@ package integration
 import (
 	"context"
 	"fmt"
-	"log/slog"
-	"os"
 	"slices"
 	"testing"
 	"time"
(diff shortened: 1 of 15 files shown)


integration_test/scenario_flow_test.go
```diff
@@ -88,27 +68,35 @@ func setupApiCreds(ctx context.Context, t *testing.T, usecasesWithCreds usecases
 		Description:    "Test API key",
 		Role:           models.API_CLIENT,
 	})
-	assert.NoError(t, err, "Could not create api key")
+	if err != nil {
+		assert.FailNow(t, "Could not create api key", err)
```
pascaldelange, line 72: If this fails, nothing else after this will work in the test suite.
With assert.NoError, the test will continue to run, generating tons of noise and making it hard to find where the initial error was.

pascaldelange:
Good for testing of snoozes.
Next thing I want to add before merging this is a test on data ingestion (batch) and scheduled execution, while I'm at it.

integration_test/scenario_flow_test.go
```diff
@@ -130,98 +128,151 @@ func setupOrgAndCreds(ctx context.Context, t *testing.T) (models.Credentials, mo
...
+	})
+	if err != nil {
+		assert.FailNow(t, "could not create inbox", err)
+	}
+	fmt.Printf("Created inbox %s successfully\n", inbox.Id)
+
+	// dm, err = usecase.GetDataModel(ctx, organizationId)
+	if err != nil {
+		assert.FailNow(t, "Could not get data model", err)
+	}
```
balzdur, line 254: something to uncomment or delete here

balzdur (approved):

integration_test/scenario_flow_test.go
pascaldelange: right, good catch

PR #650, opened 2024-08-26
feat(license): use dev license in dev environment

balzdur:
# Issue

Since the backend start to use internally the license to block some premium features, we may need to align the behaviour between frontend and backend to ensure consistency (only divergent for dev mode)

> Note: we still have some divergence for SAAS env. But since whitelisted env and SAAS licenses return the same entitlements, we ensure consistency

# Proposition

I propose to introduce the same specific logic for dev special case (more or less) :

In the frontend, when we detect a "dev env" (= Firebase auth emulator is used, we do not relly on the explicit env var) we bypass the API call to return the hardcoded dev license.

Here, I propose to do the same in the backend. The main difference is :
- `KillIfReadLicenseError` prevent on this conditionnal logic.
- if a license key is provided, use it
- if the projectid is whitelisted, use the full license instead

### DIFF
--- a/cmd/batch_ingestion.go
+++ b/cmd/batch_ingestion.go
@@ -38,6 +38,7 @@ func RunBatchIngestion() error {
 	licenseConfig := models.LicenseConfiguration{
 		LicenseKey:             utils.GetEnv("LICENSE_KEY", ""),
 		KillIfReadLicenseError: utils.GetEnv("KILL_IF_READ_LICENSE_ERROR", false),
+		IsDevEnvironment:       utils.GetEnv("FIREBASE_AUTH_EMULATOR_HOST", "") != "",
 	}
 	jobConfig := struct {
 		env           string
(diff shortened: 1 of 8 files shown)


cmd/batch_ingestion.go
```diff
@@ -38,6 +38,7 @@ func RunBatchIngestion() error {
 	licenseConfig := models.LicenseConfiguration{
 		LicenseKey:             utils.GetEnv("LICENSE_KEY", ""),
 		KillIfReadLicenseError: utils.GetEnv("KILL_IF_READ_LICENSE_ERROR", false),
+		IsDevEnvironment:       utils.GetEnv("FIREBASE_AUTH_EMULATOR_HOST", "") != "",
```
Pascal-Delange, line 41: Hmm wait, this doesn't quite work.
The FIREBASE_AUTH_EMULATOR_HOST env variable is not needed for the batch ingestion jobs (or other batch jobs). As a consequence, there is no point in setting it for this job, and conversely it can be set to any non-empty value without side effects (which would bypass license checks).
This shortcut works for the server only, because setting the variable actually changes the behavior (on authentication).
Now, considering the feature entitlements currently added in NewDevLicense, this could just about work (in practice, you'd need the entitlement on the app to create the workflows, roles etc), but I'm not sure it's the best way to move forward.

balzdur:
Ok, so let's close this for now. ATM, this should'nt be a problem since :
1. IIUC, only the `webhook` entitlement is used on the server
2. this is initialised to `false` in dev (as no license is provided)

So for now, it's consistent (more or less by chance)

PR #655, opened 2024-08-27
first minimal working end to end test of API server

Pascal-Delange:
Use [httpexpect](https://github.com/gavv/httpexpect?tab=readme-ov-file) framework to easily write end to end tests of the API, all running as go tests (so without any additional docker image etc setup steps, as we would have if we used something like [venom](https://github.com/ovh/venom).

It's all pretty much equivalent to a locally running server (with faked GCS buckets though), only the firebase client is stubbed (we use the firebase admin client, which requires authentication with service acc credentials, even though we only use it to verify id tokens, which should not require auth). 

---- 

For now, I limited myself to only testing the API for:
- create users (org, org admin, org viewer)
- create api keys
- edit data model
- create, validate, publish scenario (with simple rule)
- ingest data (API), create decision (API)

I could go further, and honestly I think that in the long run we might as well only keep this suite of end to end tests rather than the more cumbersome "integration tests of the usecases". However, it's still a bit painful to write so for this week I think I'll only push so far. A future thing to add would be to replicate in this format all the tests for workflows, case manager, inboxes, batch ingestion & decisions, etc (which exist on the old integration test suite).

---- 

It all runs in ~2-3 seconds in CI, so that should be pretty good to use.


infra/firebase_client_mock.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+
+	"firebase.google.com/go/v4/auth"
+)
+
+type mockedTokenVerifier struct{}
+
+func (m mockedTokenVerifier) VerifyIDToken(_ context.Context, email string) (*auth.Token, error) {
```
Pascal-Delange, line 11: Simple stubbed firebase api client (in fact: token verifier interface, because that is what we need) for usage in tests. I simplified things a bit by making it take an email string as input, rather than rewriting a lot of "parse a jwt token" logic.

balzdur (approved):

PR #657, opened 2024-08-28
refactor how dependencies are injected to api

Pascal-Delange:


api/usecases_with_context_credentials.go
```diff
@@ -3,49 +3,13 @@ package api
...
 	"github.com/checkmarble/marble-backend/utils"
 )
 
-type API struct {
-	router        *gin.Engine
-	usecases      usecases.Usecases
-	marbleAppHost string
-}
-
-func New(
```
Pascal-Delange, line 23: This was horrible spaghetti code with unclear dependencies, and I was left befuddled several times when I tried who of api, router, server did what.
I cleaned it up to delete the API struct. Dependencies (in particular usecases) are now injected at the route definition level. See routes.go. The really relevant changes are in there.

balzdur (approved):

api/dependencies.go
```diff
@@ -30,15 +30,15 @@ func InitDependencies(
...
 	signingKey *rsa.PrivateKey,
-	tokenVerifier tokenCookieVerifier,
+	optTokenVerifier ...tokenVerifier,
 ) dependencies {
 	database := postgres.New(dbPool)
 
-	if tokenVerifier == nil {
-		tokenVerifier = infra.InitializeFirebase(ctx)
+	if len(optTokenVerifier) == 0 {
+		optTokenVerifier = append(optTokenVerifier, infra.InitializeFirebase(ctx))
```
balzdur, line 38: I find this a litle bit too complicate for what it really does. IMO, either :
1. transform for an optionnal prop with default value (= no array)
```go
var tokenVerifier ...
if len(optTokenVerifier) >=  1 {
  tokenVerifier = optTokenVerifier[0
} else {
  tokenVerifier = infra.InitializeFirebase(ctx)
}
```
2. force the parameter to be defined, and use `infra.InitializeFirebase(ctx)` everywhere we use the "default" value (no more implicit default inside the implem)

PR #658, opened 2024-08-28
Pascal/refactor org id injection

Pascal-Delange:
Retrospectively this was more painful than I hoped for...
Anyway: the way dependencies, in particular trivial ones like the org id, are injected as a helper function into the usecases was a pain in the ass, I wanted to get rid of it for a while.

### DIFF
--- a/usecases/case_usecase.go
+++ b/usecases/case_usecase.go
@@ -94,7 +94,7 @@ func (usecase *CaseUseCase) ListCases(
 		ctx,
 		usecase.transactionFactory,
 		func(tx repositories.Executor) ([]models.CaseWithRank, error) {
-			availableInboxIds, err := usecase.getAvailableInboxIds(ctx, tx)
+			availableInboxIds, err := usecase.getAvailableInboxIds(ctx, tx, organizationId)
 			if err != nil {
 				return []models.CaseWithRank{}, err
 			}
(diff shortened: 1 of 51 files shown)


usecases/case_usecase.go
```diff
@@ -152,7 +152,7 @@ func (usecase *CaseUseCase) GetCase(ctx context.Context, caseId string) (models.
 		return models.Case{}, err
 	}
 
-	availableInboxIds, err := usecase.getAvailableInboxIds(ctx, exec)
+	availableInboxIds, err := usecase.getAvailableInboxIds(ctx, exec, c.OrganizationId)
```
Pascal-Delange, line 155: I may need to review this - tomorrow

Pascal-Delange: Ok, I had a doubt about using the orgId from the case itself to get the available inbox ids and then do the permission check based on that. 
After verification, it's ok because getAvailableInboxIds calls inboxReader.ListInboxes which itself enforces the permission check on the inboxes it returns.

utils/context_credentials.go
```diff
@@ -14,41 +15,43 @@ func CredentialsFromCtx(ctx context.Context) (models.Credentials, bool) {
 }
 
 func OrganizationIdFromRequest(request *http.Request) (organizationId string, err error) {
+	if request == nil {
```
Pascal-Delange, line 18: @balzdur

balzdur (approved):

PR #659, opened 2024-08-29
set status of batch ingestion job to failed if errors occur (including 1h timeout)

Pascal-Delange:
Goes with https://github.com/checkmarble/marble-frontend/compare/pascal/mar-621-batch-job-data-ingestion-correctly-sets-status-in-error-case?expand=1

--- 

The motivation is here https://checkmarble.slack.com/archives/C05VAKSTL02/p1724747857435739
I ended up manually correcting upload log statuses in the db when there were errors.

In a next step, we could add mailing or other alerting to the users - but this is a good first improvement.
I'll do something similar for the scheduled execution (or at least make sure it doesn't remain in "processing" when the cloud run job times out).

---
I set a 1h timeout on the ingestion job (the cloud run job itself has a 2h timeout) which should be super sufficient (we'd be limited by the timeout of the file upload well before that).

--- 
I also added a field with the number of rows actually ingested on the upload log. Event if it fails before the end, because rows are ingested by batches of 1000, a part of the rows may have been ingested even if the final status of the job is "failure". See the front PR.

### DIFF
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -253,17 +253,20 @@ func (usecase *IngestionUseCase) processUploadLog(ctx context.Context, uploadLog
 	logger := utils.LoggerFromContext(ctx)
 	logger.InfoContext(ctx, fmt.Sprintf("Start processing UploadLog %s", uploadLog.Id))
 
-	err := usecase.uploadLogRepository.UpdateUploadLogStatus(ctx, exec, models.UpdateUploadLogStatusInput{
+	done, err := usecase.uploadLogRepository.UpdateUploadLogStatus(ctx, exec, models.UpdateUploadLogStatusInput{
 		Id:                           uploadLog.Id,
 		CurrentUploadStatusCondition: models.UploadPending,
 		UploadStatus:                 models.UploadProcessing,
 	})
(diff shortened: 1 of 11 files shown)


usecases/ingestion_usecase.go
```diff
@@ -240,121 +248,168 @@ func (usecase *IngestionUseCase) IngestDataFromCsv(ctx context.Context, logger *
...
-		Id: uploadLog.Id, UploadStatus: models.UploadProcessing,
+	done, err := usecase.uploadLogRepository.UpdateUploadLogStatus(ctx, exec, models.UpdateUploadLogStatusInput{
+		Id:                           uploadLog.Id,
+		CurrentUploadStatusCondition: models.UploadPending,
+		UploadStatus:                 models.UploadProcessing,
 	})
 	if err != nil {
 		return err
+	} else if !done {
+		logger.InfoContext(ctx, fmt.Sprintf("UploadLog %s is no longed in pending status", uploadLog.Id))
```
balzdur, line 264:
```suggestion
		logger.InfoContext(ctx, fmt.Sprintf("UploadLog %s is no longer in pending status", uploadLog.Id))
```

balzdur (approved):

PR #661, opened 2024-08-30
rate limiting on sending events to convoy

Pascal-Delange:


repositories/convoy_repository.go
```diff
@@ -28,16 +29,26 @@ func (noOpConvoyClientProvider) GetProjectID() string {
...
 
-	return ConvoyRepository{convoyClientProvider: convoyClientProvider}
-}
-
-type ConvoyRepository struct {
-	convoyClientProvider ConvoyClientProvider
+	// other instances may be running in parallel. If we want to do it more precisely, we'll need a distributed limiter
+	instanceLimit := rate.Limit(limit) / 2
+	burst := limit / 3
+	if burst < 1 {
```
Pascal-Delange, line 45: burst = 0 means no events sent (unless limit = infinity by convention)

usecases/webhook_events_usecase.go
```diff
@@ -112,9 +113,9 @@ func (usecase WebhookEventsUsecase) SendWebhookEventAsync(ctx context.Context, w
 	logger := utils.LoggerFromContext(ctx).With("webhook_event_id", webhookEventId)
 	ctx = utils.StoreLoggerInContext(ctx, logger)
 
-	// go routine to send the webhook event asynchronously, with a new context and timeout and a child span
+	// goroutine to send the webhook event asynchronously, with a new context and timeout and a child span
 	go func() {
-		ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), time.Second*3)
+		ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), ASYNC_WEBHOOKS_SEND_TIMEOUT)
```
Pascal-Delange, line 118: A remark about this (I struggled a bit with this and unexpected errors at first).
The rate limiter will return an error if we call Wait() and the expected wait time (considering other requests on the rate limiter instance) is longer than the remaining context duration, considering the configured refresh rate.
With a 5 minute timeout and a 25/sec refresh rate (corresponding to a 50/res rate limit on convoy configured in env variable), that's 25*60*5=7500 events max that can be queued before the internal rate limiter starts dropping some (the initial commit would have been only 25*3=75 events, so that batch retry jobs would still not quite have worked).

repositories/convoy_repository.go
balzdur: Can you change for explicit equal 0 condition instead of inequality ?

balzdur (approved):

PR #662, opened 2024-09-02
fix: api only returns active snoozes (& ignores old ones on POST)

Pascal-Delange:


repositories/rule_snoozes.go
```diff
@@ -158,13 +158,21 @@ func (repo *MarbleDbRepository) ListRuleSnoozesForDecision(
 		return nil, err
 	}
 
+	if len(snoozeGroupIds) == 0 {
```
Pascal-Delange, line 161: fix of a possible bug: no filter would be applied if an empty slice is passed (AFAIK)

balzdur (approved):

PR #671, opened 2024-09-09
new review status field on decisions

Pascal-Delange:
- 1st commit contains the actual changes
- 2nd commit is only types renaming
- 3rd commits makes nullable a field I want to remove

Can be reviewed, pending discussion with Arnaud on the actual values we use


Pascal-Delange:
@balzdur I actually did it with pending|decline|approve yesterday, so nothing to change

balzdur (approved):

PR #679, opened 2024-09-16
Pascal/mar 635 use gocloud to use provider agnostic blob storage

Pascal-Delange:
This PR allows for choosing buckets from different cloud providers (S3, GCS, azure blob, or S3-compatible APIs), simply by using the right bucket url in the env variables.

Copied from the .env.example:
```shell
# Configure the document storage backend with optional fake backends.
# The options are
# - "file://{path}?create_dir=true"
#    - for LOCAL TEST only, not fully compatible with all features - create_dir not required if the folder exists)
#    - see https://pkg.go.dev/gocloud.dev/blob/fileblob#URLOpener for details on the format
# - "gs://{bucket_name}"
# - "s3://{bucket_name}"
# - "azblob://{bucket_name}"
# See https://gocloud.dev/howto/blob/ for details.
# Using a Google Cloud Storage (gs://) bucket backend additionally requires that a service account key be present in the file pointed 
# to by GOOGLE_APPLICATION_CREDENTIALS. This is because with GCS, a private key from a service account key must be present to return signed urls.
# In all cases, credentials discovery is done automatically according to each provider's usual practices (env variables, 
# credentials file, metadata server...)
INGESTION_BUCKET_URL="file://./tempFiles/data-ingestion-bucket?create_dir=true"
CASE_MANAGER_BUCKET_URL="file://./tempFiles/case-manager-bucket?create_dir=true"
TRANSFER_CHECK_ENRICHMENT_BUCKET_URL="file://./tempFiles/transfercheck-bucket?create_dir=true"
```

------ 

Changes from before (here for internal use - I'll write a separate post in the changelog and in the doc):
- for local usage with a file-based backend:
    - drop the "FAKE_GCS" env variable
    - instead use "file://{path}?create_dir=true" as the bucket URL. See the .env.example file
    - notice that the env variable name has changed to `XXX_BUCKET_URL` from `GCS_XXX_BUCKET`
- for cloud deployment:
    - use a bucket url adapted for the provider (starts with gs://, s3://, azblob://)
    - for S3 or Azure, the standard ways of authenticating to the SDKs are used (as should be idiomatic for users of those providers). I tested the feature locally with AWS (using AWS_SECRET_KEY & AWS_ACCESS_KEY & AWS_REGION env variables) and with Azure (using AZURE_STORAGE_ACCOUNT & AZURE_STORAGE_CONNECTION_STRING env variables)
    - for GCS, it is necessary to inject a json service account key file to the container (or VM), and to refer to it in the GOOGLE_APPLICATION_CREDENTIALS env variable. This is because a private key from a service account should be used to sign URLs. Before using gocloud, the SDK would fall back to using the service account's default signing method, but this is not great in the "client on-premise" context because I'm not sure how it behaves outside of GCP infra. Anyway, doing this setup is already necessary to make the firebase SDK work in the backend outside of a GCP infra - it is now required for everybody, to reduce edge cases in the setup steps.
    - as before, permissions should be set up properly so that the service has permissions to write/read/delete files as needed on the buckets
    - more specific documentation on the expected format of urls for the different providers can be found at https://pkg.go.dev/gocloud.dev@v0.39.0/blob, in the different sub-directories for each provider

Note that for an on-premise setup, it may be fine to use just one bucket for everything - separation of concerns is more important for our multi-tenant, managed setup.

----- 

Prerequisite Terraform PR: https://github.com/checkmarble/terraform/pull/31


Pascal-Delange:
Code works. Remains to do: write the doc (internal & external) on how to set it up.

Pascal-Delange:
+ update the env variables in terraform for our managed environment

.vscode/settings.json
```diff
@@ -16,4 +16,13 @@
             ".git/logs": true,
       },
       "editor.tabSize": 4,
+      "files.associations": {
+            // This is in order to deactivate github copilot on the .env file. As a result syntax coloring will be disabled because
+            // VSCode will treat it as plain text, you can choose your language mode manually on the file
+            ".env": "env"
+      },
+      "github.copilot.enable": {
```
Pascal-Delange, line 24: avoid that env variables configured in .env containing secrets are leaked to github copilot

Pascal-Delange:
I'm going to consider that the doc I put in the .env.example is sufficient for our internal doc on "how to run marble locally".
I'll replicate it (or something similar) in the example env file in our marble repo for open-source users, and point to it in our changelog message.

A more complete explanation will be added in the "self-hosted onboarding guide" whenever we write it.

repositories/blob_repository.go
```diff
@@ -0,0 +1,284 @@
...
+	_ "gocloud.dev/blob/azureblob"
+	_ "gocloud.dev/blob/fileblob"
+	"gocloud.dev/blob/gcsblob"
+	_ "gocloud.dev/blob/s3blob"
+	"gocloud.dev/gcp"
+)
+
+const (
+	signedUrlExpiry    = 1 * time.Hour
+	placeholderFileUrl = "https://storage.googleapis.com/case-manager-tokyo-country-381508/54624b1f-aac3-4d3c-8fee-75db36436e12/5fd7bbee-b2df-4bc1-9ce8-e218e607c352/d9077dda-7836-45bd-bfde-0c8a3a0c0ad9?Expires=1821107332&GoogleAccessId=marble-backend-cloud-run%40tokyo-country-381508.iam.gserviceaccount.com&Signature=YYGF0msoL%2FmiIb3BKroFDRP0DzZrHQF3pS5VudT0OymeNnmxoIZS5DOycPaCcRa%2FMbRh454YEpAQGT%2F6Xf5dGWo%2FEj7UfmoKmPPyRGZ82qo9lr1ZMdvveBtBmSdzepgk6EBkWxWX3Ov0ZOguD58pKVy4Q0WzaMl5aD8dN8jv2ExuCfGRvNCpfvP43eONEtox6ilPkVkq4Bqhq9BHo4OBQj%2FuU8BLfnge35Db3IIy2f69CJ0wagLPiYkWfu5GODgoXMsjL0JtNEryeCJMH2ocXrTlV0XD00bx%2F8vhaHCHY9o%2Ft1V0sHmd7CoIa3bUsx4gixCuxivqvc5xLeDkgeTf%2FA%3D%3D"
```
balzdur, line 33: Not really impactfull, but I think the comment on `// placeholder file, url valid for 3 years from 2024/09/16` should go there

balzdur (approved):

repositories/blob_repository.go
Pascal-Delange: ah, yes of course

usecases/ingestion_usecase.go
```diff
@@ -280,14 +281,16 @@ func (usecase *IngestionUseCase) processUploadLog(ctx context.Context, uploadLog
 		}
 	}
 
-	file, err := usecase.gcsRepository.GetFile(ctx, usecase.GcsIngestionBucket, uploadLog.FileName)
+	file, err := usecase.blobRepository.GetBlob(ctx, usecase.GcsIngestionBucket, uploadLog.FileName)
+	if file.ReadCloser != nil {
+		defer file.ReadCloser.Close()
+	}
```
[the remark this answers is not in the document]
Pascal-Delange: Ah, yes I did them in both orders.
Honestly it's equivalent, writer is expected to be nil if err is not, and vice versa.
The important thing with writers is to:
- especially close them when you are done writing and check the err response, because it may still finish writing stuff from the previous call to Write (there is a buffer)
- preferably still defer Close() it to be sure you don't leave dangling writers hanging

.vscode/settings.json
Pascal-Delange: by default VSCode seems to be coloring .env files (or anything that is not typed?) like shell files, which I had to manually set again after changing this setting

PR #680, opened 2024-09-16
Pascal/kill export s3 feature

Pascal-Delange:
To do:
- update the interface in the front (S3 export is mentioned somewhere)
- update the doc
- update the backoffice org update

To do next week:
- remove the field from DB (currently it's not nullable with a default value `''`


dto/organization_dto.go
```diff
@@ -3,16 +3,14 @@ package dto
 import "github.com/checkmarble/marble-backend/models"
 
 type APIOrganization struct {
-	Id                         string `json:"id"`
-	Name                       string `json:"name"`
-	ExportScheduledExecutionS3 string `json:"export_scheduled_execution_s3"`
```
balzdur, line 8: Be mindfull, if you merge this, `export_scheduled_execution_s3` won't be available in org schema anymore (with get request).
This will break the BO due to this :
```ts
const OrganizationSchema = yup.object({
  id: yup.string().required(),
  name: yup.string().defined(),
  export_scheduled_execution_s3: yup.string().defined(),
});
```

balzdur (approved):

dto/organization_dto.go
Pascal-Delange: done in https://github.com/checkmarble/marble-backoffice/pull/26

PR #686, opened 2024-09-25
set TOAST_TUPLE_TARGET for decision_rules + no exec detail on TC

Pascal-Delange:
- Better setup of TOAST for decision_rules: storage of decision rule exec detail is dominating and it's bloat (see [postgres doc](https://www.postgresql.org/docs/current/storage-toast.html), [slack discussion](https://checkmarble.slack.com/archives/C04JB5ZADBN/p1727257822435979))
- Do not store the rule execution details for TC scenario executions (in the longer term, move to a blob storage, see [slack discussion](https://checkmarble.slack.com/archives/C04JB5ZADBN/p1727257822435979))
- Normalize decision_rule name & description - we are never filtering by them, they have no business being denormalized
- I also modified the sql questions in Metabase accordingly - though right now they're broken anyway because of the big decision_rules table

### DIFF
--- a/repositories/dbmodels/db_decision_rule.go
+++ b/repositories/dbmodels/db_decision_rule.go
@@ -3,29 +3,23 @@ package dbmodels
 import (
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/models/ast"
-	"github.com/checkmarble/marble-backend/utils"
-
-	"github.com/jackc/pgx/v5/pgtype"
 )
 
 type DbDecisionRule struct {
-	Id             string             `db:"id"`
-	OrganizationId string             `db:"org_id"`
-	DecisionId     string             `db:"decision_id"`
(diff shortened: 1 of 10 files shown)


repositories/dbmodels/db_decision_rule.go
```diff
@@ -3,30 +3,24 @@ package dbmodels
 import (
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/models/ast"
-	"github.com/checkmarble/marble-backend/utils"
-
-	"github.com/jackc/pgx/v5/pgtype"
 )
 
 type DbDecisionRule struct {
-	Id             string             `db:"id"`
```
Pascal-Delange, line 12: Those tags are used with the method `SqlToListOfModels` - it's kind of convenient when reading from a single table, but its convenience really breaks down as soon as we do joins

balzdur (approved):

PR #687, opened 2024-09-26
Pascal/mar 650 apply initial filter computed from trigger condition ast for

Pascal-Delange:
Parse the trigger condition AST node, to apply filters that can be applied on the query that returns the rows on which we want to run a batch execution of a scenario.

A "where" filter on the query will be applied in the following cases, for every row in the list of "AND" clause in the trigger condition where the operator is of the form "X OPS Y", with:
- OPS is one of the following operators: >,>=,<,<=,=,!=,is in, is not in, contains, does not contain. The "contains any of", "does not contain any of" operators are not handled.
- the left hand (X) and right hand (Y) operands are one of: a payload field, a constant, the now() operator, the date shift operator applied on a payload field or the now() operator. "is in list" operator only works with constant lists of strings (we could perhaps add custom list later, but I don't think it makes much of a difference).
- at least one of X and Y is a payload field, or derived from a payload field (date shift operator), this is what the actual filter in the query will act on

Some care had to be taken to make sure that we avoid any kind of sql injection (sanitization of the values we pass into the final query).
I do assume that the AST is in the form where we expect it (since it's presumably been validated), though I don't go to the point where I reconstruct a second AST validation. So there is some level of dependency between this feature and the AST.

To do at the point of writing:
- [x] works for (in)equality conditions
- [x] works for list conditions
- [x] works for string contains matching
- [x] run more extensive tests
- [x] handle nested parts in the conditions

---

One thing I'm not too happy about is that it kind of creates entanglement between pure logic on the AST and models on one side, and the SQL queries on the DB on the other hand.
But TBH I don't see a good way around it for now, at least not without going through excessive layers of abstraction that I'm not comfortable with right now, so I'll live with it.

--- 
## Local test example
Capture d’écran 2024-09-27 à 16 02 11

gives the following query:
```sql
SELECT
    account_name,
    after_migration,
    bool,
    not_uniq,
    num,
    object_id,
    uniq,
    updated_at
FROM
    "org-xpollens"."Accounts"
WHERE
    "org-xpollens"."Accounts".valid_until = $1
    AND now() + interval 'PT0S' < "org-xpollens"."Accounts"."updated_at"
    AND $2 < ($3 / NULLIF("org-xpollens"."Accounts"."num", 0))
    AND $4 = ANY ($5)
    AND $6 ILIKE CONCAT('%', $7, '%')
    AND "org-xpollens"."Accounts"."updated_at" < now()
    AND "org-xpollens"."Accounts"."after_migration" = ("org-xpollens"."Accounts"."bool" != $8)
```

and the following arguments:
Capture d’écran 2024-09-27 à 16 02 40
Lines 6, 7, 8 are not used to construct filters, as desired.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
         "--server",
         "--migrations"
       ],
-      // "console": "integratedTerminal"
+      "console": "integratedTerminal"
     },
     {
       "name": "Launch scenario scheduling job (.env)",
@@ -27,7 +27,7 @@
       "args": [
         "--scheduler"
       ],
-      // "console": "integratedTerminal"
+      "console": "integratedTerminal"
     },
     {
       "name": "Launch execute scheduled scenario job (.env)",
@@ -39,7 +39,7 @@
       "args": [
         "--scheduled-executer"
       ],
-      // "console": "integratedTerminal"
+      "console": "integratedTerminal"
     },
     {
       "name": "Launch data ingestion job (.env)",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
         "--server",
         "--migrations"
       ],
-      "console": "integratedTerminal"
+      // "console": "integratedTerminal"
     },
     {
       "name": "Launch scenario scheduling job (.env)",
@@ -27,7 +27,7 @@
       "args": [
         "--scheduler"
       ],
-      "console": "integratedTerminal"
+      // "console": "integratedTerminal"
     },
     {
       "name": "Launch execute scheduled scenario job (.env)",
@@ -39,7 +39,7 @@
       "args": [
         "--scheduled-executer"
       ],
-      "console": "integratedTerminal"
+      // "console": "integratedTerminal"
     },
     {
       "name": "Launch data ingestion job (.env)",
--- a/usecases/scheduled_execution/batch_filtering.go
+++ b/usecases/scheduled_execution/batch_filtering.go
@@ -1,0 +1,176 @@
+package scheduled_execution
+
+import (
+	"slices"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/models/ast"
+)
+
+func selectFiltersFromTriggerAstRootAnd(node ast.Node) ([]models.Filter, error) {
+	if node.Function != ast.FUNC_AND {
+		return nil, nil
+	}
+
+	filters := make([]models.Filter, 0, 10)
+	for _, node := range node.Children {
+		filter, valid := astConditionToFilter(node)
+		if valid {
+			filters = append(filters, filter)
+		}
+	}
+	return filters, nil
+}
+
+func astConditionToFilter(node ast.Node) (models.Filter, bool) {
+	switch node.Function {
+	case ast.FUNC_GREATER,
+		ast.FUNC_GREATER_OR_EQUAL,
+		ast.FUNC_LESS,
+		ast.FUNC_LESS_OR_EQUAL,
+		ast.FUNC_EQUAL,
+		ast.FUNC_NOT_EQUAL:
+		return models.Filter{}, false
+	case ast.FUNC_IS_IN_LIST,
+		ast.FUNC_IS_NOT_IN_LIST,
+		ast.FUNC_CONTAINS_ANY,
+		ast.FUNC_CONTAINS_NONE:
+		return models.Filter{}, false
+	case ast.FUNC_STRING_CONTAINS,
+		ast.FUNC_STRING_NOT_CONTAIN:
+		return models.Filter{}, false
+	}
+	return models.Filter{}, false
+}
+
+func filterFromComparisonNode(node ast.Node) (models.Filter, bool) {
+	if !slices.Contains([]ast.Function{
+		ast.FUNC_GREATER,
+		ast.FUNC_GREATER_OR_EQUAL,
+		ast.FUNC_LESS,
+		ast.FUNC_LESS_OR_EQUAL,
+		ast.FUNC_EQUAL,
+		ast.FUNC_NOT_EQUAL,
+	}, node.Function) {
+		return models.Filter{}, false
+	}
+
+	// left options:
+	// - constant (string, number, boolean allowed. String can be timestamp)
+	// - payload (field name)
+	// - time_add operator with something else inside
+	// - "now" operator
+
+	// right options:
+	// same things actually, the comparison is "symmetric"
+
+	// one at least must involve a "payload" field, otherwise nothing to filter on the table
+
+	fieldName, err := node.ReadConstantNamedChildString("fieldName")
+	if err != nil {
+		return models.Filter{}, false
+	}
+
+	value, err := node.ReadConstantNamedChildString("value")
+	if err != nil {
+		return models.Filter{}, false
+	}
+
+	return models.Filter{
+		LeftFieldOrValue:  nil,
+		Operator:          node.Function,
+		RightFieldOrValue: nil,
+	}, true
+}
+
+func comparisonValueFromNode(node ast.Node) (value any, involvesPayload bool, valid bool) {
+	// case 1: payload field
+
+	// case 2: date add operator
+
+	// case 3: now operator
+
+	// case 4: custom list accessor
+
+	// case 5: constant value
+	//   - string
+	//   - number
+	//   - boolean
+	//   - timestamp, but it's a string (we can pass the string to the sql query)
+	//   - list of strings => should not happen here because it's a comparison operator
+
+	return 0, false, false
+}
+
+// func aggregationNodeToQueryFamily(node ast.Node) (models.AggregateQueryFamily, error) {
+// 	if node.Function != ast.FUNC_AGGREGATOR {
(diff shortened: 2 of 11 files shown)


.vscode/launch.json
```diff
@@ -15,7 +15,7 @@
         "--server",
         "--migrations"
       ],
-      "console": "integratedTerminal"
+      // "console": "integratedTerminal"
```
balzdur, line 18: To remove

usecases/scheduled_execution/batch_filtering.go
```diff
@@ -0,0 +1,185 @@
...
+	"slices"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/models/ast"
+	"github.com/checkmarble/marble-backend/pure_utils/duration"
+
+	"github.com/jackc/pgx/v5"
+)
+
+func selectFiltersFromTriggerAstRootAnd(node ast.Node, table models.TableIdentifier) []models.Filter {
```
balzdur, line 14: Good anticipation to the possibility of OR trigger condition

usecases/scheduled_execution/batch_filtering.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func astConditionToFilter(node ast.Node, table models.TableIdentifier) (models.Filter, bool) {
+	switch node.Function {
+	case ast.FUNC_GREATER,
+		ast.FUNC_GREATER_OR_EQUAL,
+		ast.FUNC_LESS,
+		ast.FUNC_LESS_OR_EQUAL,
+		ast.FUNC_EQUAL,
+		ast.FUNC_NOT_EQUAL:
```
balzdur, line 36: Worst formatting. Not a usefull comment but I needed to tell it

As a side note, I think we should extract this subset of functions as a "function". So something like :
```go
if isMathOperator(....)

if isListOperator(...)
````

It's something we already have in the frontend and it helps to share behaviours when adding new functions.

I must admit, if the one we add is not compatible with this logic it will end bad ^^'
But if it does (= the group is usefull) it helps

File: usecases/scheduled_execution/batch_filtering.go [hunk 4]
```diff
@@ -0,0 +1,185 @@
...
+		return filterFromComparisonNode(node, table)
+	case ast.FUNC_IS_IN_LIST,
+		ast.FUNC_IS_NOT_IN_LIST,
+		ast.FUNC_CONTAINS_ANY,
+		ast.FUNC_CONTAINS_NONE:
+		return models.Filter{}, false
```
balzdur, line 42: IIUC, this is something you will add tomorrow ?

usecases/scheduled_execution/batch_filtering.go
```diff
@@ -0,0 +1,185 @@
...
+	case ast.FUNC_STRING_CONTAINS,
+		ast.FUNC_STRING_NOT_CONTAIN:
+		return models.Filter{}, false
+	}
+	return models.Filter{}, false
+}
+
+func filterFromComparisonNode(node ast.Node, table models.TableIdentifier) (models.Filter, bool) {
```
balzdur, line 50: You have the name of the sub set here 😛

usecases/scheduled_execution/batch_filtering.go
```diff
@@ -0,0 +1,185 @@
...
+		switch node.Constant.(type) {
+		case string, float64, bool:
+			return parsedFilter{value: node.Constant, valid: true}
+		default:
+			return parsedFilter{}
+		}
+	}
+
+	// return no filter in all other cases, in particular aggregates etc.
+	return parsedFilter{}
```
balzdur, line 184: And also, nesting !

usecases/scheduled_execution/batch_filtering.go
```diff
@@ -0,0 +1,185 @@
...
+	involvesPayload bool
+	valid           bool
+}
+
+func payloadFieldNodeToSql(node ast.Node, table models.TableIdentifier) (string, bool) {
+	field, ok := node.Children[0].Constant.(string)
+	return pgx.Identifier.Sanitize([]string{table.Schema, table.Table, field}), ok
+}
+
+func comparisonValueFromNode(node ast.Node, table models.TableIdentifier) parsedFilter {
```
balzdur, line 116: This function is really complicate, and i think I do not understand the need for `involvesPayload`

balzdur (commented):
IIUC, the main goal is to reduce the amount of rows returned by the batch exec, so it "speed up" the process (in the end, the trigger is still evaluated, but we do this on a reduced number of rows)

.vscode/launch.json
Pascal-Delange: yea, I'm still having issues with my VSCode randomly... it works when I first start it, then starts failing randomly after a few script restarts

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: haha yea, for this one I actually took inspiration from the front end !

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: yea exactly, I'm not done with those (but the part I did was the one with most branches AKA the most complex one)

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: sorry I did not understand this comment 😅

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: Yea formatting was not kind here. I still like it better like this than together in two lines.

I get the idea of extracting, though in this case it would break the usage of switch (or I must implement a function that maps functions to a type (implying only one)... maybe there is an elegant way around this but I think I'll pass it for now.

Pascal-Delange: (I also think it's unlikely that we add more math comparison operators)

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: The idea behind `involvesPayload` is: I don't care about any "lines" in the list of AND conditions that don't involve the payload, as by definition they won't lead to any filters on the table being read.
Think of things like parent (linked) fields (since we don't do a join for the filter), aggregates, or such.

Pascal-Delange: I'll see when I'm done if I see a good refactoring of the function - though TBH right now it's already split into 4 well separated cases

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: Daaamn, that's a good catch and one thing I forgot.
Ok, if I understand correctly what I should worry about is mostly the nested math operations: `+ - * /` (technically there are also all the comparison operators returning booleans, but perhaps I can ignore those for now since nested bool conditions seem unlikely).
Well, guess I'll have to handle those - protecting against any division by zero errors with NULLIF)

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: About `involvesPayload`, I guess in hindsight it's perhaps not strictly necessary if I already reject any condition that uses an aggregation or "linked field" lookup. That would just mean ignoring vs not ignoring conditions that are "constant" - like constant to constant comparison. I suppose I could just remove the condition on `involvesPayload`.

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: ok I did something like this in models/scheduled_scenario_executions.go

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: Ok I have something that works with nested - cut at depth 3 max for safety

usecases/scheduled_execution/batch_filtering.go
Pascal-Delange: I removed `involvesPayload`

usecases/scheduled_execution/batch_filtering_test.go
```diff
@@ -0,0 +1,186 @@
...
+		ruleJson string
+		expected models.Filter
+		valid    bool
+	}
+
+	testCases := []testCase{
+		{
+			name: "timestamp comparison with date shift",
+			ruleJson: `{
+				"name": "\u003c",
```
balzdur, line 31:
```suggestion
				"name": "<",
```

usecases/scheduled_execution/batch_filtering_test.go
```diff
@@ -0,0 +1,186 @@
...
+				LeftSql:  "now() + interval 'PT0S'",
+				Operator: ast.FUNC_LESS,
+				RightSql: "\"schema\".\"table\".\"updated_at\"",
+			},
+			valid: true,
+		},
+		{
+			name: "number comparison with nested division",
+			ruleJson: `{
+				"name": "\u003c",
```
balzdur, line 54:
```suggestion
				"name": "<",
```

balzdur (approved):

PR #690, opened 2024-10-02
WIP: rework batch execution

Pascal-Delange:
## Content
In this PR:
- the ids of (ingested) objects on which we want to take decisions are written to the DB ahead of creating the decisions
- the details of every object is only read when we actually create the decision on it
- decisions are then created one by one, each one in its own transaction, until an error encountered or all decisions have been created
- at the end, the scheduled execution is updated with its final status (can now be `"partial_failure"`) and with:
    - the number of decisions that were planned to evaluate
    - the number of decisions that have actually been evaluated
    - the number of decisions that have been created (trigger condition match & no error)
- intermediate "decisions waiting to be created" are written to their own table, and their status is updated when they are completed. **_However_**, I'm keeping open the option to not conserve them indefinitely, and I'd rather no application logic relies on them too much

----
## Front end side
Add the display of the different decision numbers, remove the decision download button.
See https://github.com/checkmarble/marble-frontend/pull/556

---- 

Notes: currently we write `the number of decisions that have actually been evaluated` to the field named `number_of_created_decisions`, which is obviously misleading, and is fixed in this PR

### DIFF
--- a/repositories/scheduled_executions.go
+++ b/repositories/scheduled_executions.go
@@ -19,7 +19,6 @@ type dbJoinScheduledExecutionAndScenario struct {
 
 func adaptJoinScheduledExecutionWithScenario(row pgx.CollectableRow) (models.ScheduledExecution, error) {
 	db, err := pgx.RowToStructByPos[dbJoinScheduledExecutionAndScenario](row)
-	// TODO: how the hell does it know where to put values ????
 	if err != nil {
 		return models.ScheduledExecution{}, err
 	}
--- a/repositories/scheduled_executions.go
+++ b/repositories/scheduled_executions.go
@@ -19,6 +19,7 @@ type dbJoinScheduledExecutionAndScenario struct {
 
 func adaptJoinScheduledExecutionWithScenario(row pgx.CollectableRow) (models.ScheduledExecution, error) {
 	db, err := pgx.RowToStructByPos[dbJoinScheduledExecutionAndScenario](row)
+	// TODO: how the hell does it know where to put values ????
 	if err != nil {
 		return models.ScheduledExecution{}, err
 	}
--- a/repositories/scheduled_executions.go
+++ b/repositories/scheduled_executions.go
@@ -140,7 +140,11 @@ func (repo *MarbleDbRepository) UpdateScheduledExecutionStatus(
 			*updateScheduledEx.NumberOfCreatedDecisions)
 	}
 
-	// return ExecBuilder(ctx, exec, query)
+	if updateScheduledEx.NumberOfEvaluatedDecisions != nil {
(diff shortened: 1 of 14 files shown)


usecases/scheduled_execution/run_scheduled_execution.go
```diff
@@ -110,54 +121,6 @@ func NewRunScheduledExecution(
 	}
 }
 
-func (usecase *RunScheduledExecution) ScheduleScenarioIfDue(ctx context.Context, organizationId string, scenarioId string) error {
```
Pascal-Delange, line 113: Moved to another file along with other methods used in it, no other changes (see schedule_scenarios.go)

usecases/scheduled_execution/run_scheduled_execution.go
```diff
@@ -261,231 +231,239 @@ func (usecase *RunScheduledExecution) executeScheduledScenario(ctx context.Conte
...
-					"object", object)
-				return nil
-			} else if err != nil {
-				return errors.Wrapf(err, "error evaluating scenario in executeScheduledScenario %s", scenario.Id)
-			}
-
-			decision := models.AdaptScenarExecToDecision(scenarioExecution, object, &scheduledExecutionId)
-			err = usecase.decisionRepository.StoreDecision(
+		// Update the number of created decisions on the scheduled execution, but not at every decision
+		if i%100 == 0 {
```
Pascal-Delange, line 332: This means we'll have an intermediate estimate of the number of decisions evaluated/created during the batch execution. The final number will be written at the end of the job.
Problem: this logic is not directly reusable once I move to an async worker implementation of batch decision creation (I can then no longer rely on the sequence of decisions being created).
I'll think of an evolution, worst case I'll drop this "feature"

repositories/scheduled_executions.go
```diff
@@ -155,3 +160,78 @@ func (repo *MarbleDbRepository) UpdateScheduledExecutionStatus(
 	}
 	return true, nil
 }
+
+func (repo *MarbleDbRepository) UpdateScheduledExecution(
+	ctx context.Context,
+	exec Executor,
+	input models.UpdateScheduledExecutionInput,
+) error {
+	// uses optimistic locking based on the current status to avoid overwriting the status incorrectly
```
balzdur, line 169: I don't understand the comment regarding this function implem

balzdur (approved):

repositories/scheduled_executions.go
Pascal-Delange: my bad, this must be leftover from a copy paste from the function above, thanks

Pascal-Delange: I accidentally corrected it in the other PR instead

PR #696, opened 2024-10-03
merge the functionality of the "schedule scenarios" ...

Pascal-Delange:
merge the functionality of the "schedule scenarios" entrypoint into the entrypoint that executes them

First step in an effort to start removing or greatly reduce the cron job entry points


jobs/scheduled_scenarios.go
```diff
@@ -3,24 +3,43 @@ package jobs
...
 		func(
 			ctx context.Context, usecases usecases.Usecases,
 		) error {
-			return scheduledScenarios(ctx, usecases)
+			usecasesWithCreds := GenerateUsecaseWithCredForMarbleAdmin(ctx, usecases)
+			runScheduledExecution := usecasesWithCreds.NewRunScheduledExecution()
+			ctx, cancel := context.WithTimeout(ctx, batchScenarioExecutionTimeout)
+			defer cancel()
+
+			// First, schedule all due scenarios
```
Pascal-Delange, line 30: N.B.
A possible worry was we may create several pending batch executions for the same scenario.
This should not happen because business logic in the usecase should avoid it, and I further created a unique index per scenario_id (only for the 2 pending statuses) to be double sure

balzdur (approved):

PR #697, opened 2024-10-07
ingest by batch via API

Pascal-Delange:
### Content
We can now ingest objects by batch. The number of max items in a batch is set to 100 by default and in our SaaS offering, but can be customized by on-premise customers if it makes sense for them, by setting the `"BATCH_INGESTION_MAX_SIZE"` env variable.

---
### Fixes
Renamed some variables/object properties that still contained "GCS" instead of "URL"

---
### QA
Tested with https://github.com/checkmarble/marble-backoffice/pull/29 and with a test case in the end to end test.

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -49,6 +49,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/:decision_id/snooze", handleSnoozeDecision(uc))
 
 	router.POST("/ingestion/:object_type", handleIngestion(uc))
+	router.POST("/ingestion/:object_type/multiple", handleIngestionMultiple(uc))
 	router.POST("/ingestion/:object_type/batch", timeoutMiddleware(batchIngestionTimeout), handlePostCsvIngestion(uc))
 	router.GET("/ingestion/:object_type/upload-logs", handleListUploadLogs(uc))
 
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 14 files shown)


api/routes.go
```diff
@@ -49,6 +49,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/:decision_id/snooze", handleSnoozeDecision(uc))
 
 	router.POST("/ingestion/:object_type", handleIngestion(uc))
+	router.POST("/ingestion/:object_type/multiple", handleIngestionMultiple(uc))
```
Pascal-Delange, line 52: that was dumb, but /batch was already taken 🤦

usecases/ingestion_usecase.go
```diff
@@ -27,21 +27,23 @@ import (
...
+	ingestionRepository   repositories.IngestionRepository
+	blobRepository        repositories.BlobRepository
+	dataModelRepository   repositories.DataModelRepository
+	uploadLogRepository   repositories.UploadLogRepository
+	ingestionBucketUrl    string
+	batchIngestionMaxSize int
 }
 
-func (usecase *IngestionUseCase) IngestObjects(
+func (usecase *IngestionUseCase) IngestObject(
```
Pascal-Delange, line 46: just a comment to myself, but it's funny that git/github use a different "diff algorithm" than VSCode does

balzdur (approved):
Since the same repository method is used under the hood, why do you need to use two distinct usecase ? It seems the "single" usecase is a special usecase from "multiple"

Pascal-Delange:
> Since the same repository method is used under the hood, why do you need to use two distinct usecase ? It seems the "single" usecase is a special usecase from "multiple"

well, it's pretty similar, but because I'm doing the body decoding in the usecase the decoding part is slightly different.
Anyway, probably slightly more code, but this is not expected to be touched too much anytime soon (or if we do, I'll refactor if needed) so I didn't want to spend too much effort in code quantity minimization

PR #703, opened 2024-10-10
chore: distinguish between transactions and generic executor passed b…

Pascal-Delange:
…etween usecases and repos


repositories/ingestion_repository.go
```diff
@@ -13,24 +13,24 @@ import (
 )
 
 type IngestionRepository interface {
-	IngestObjects(ctx context.Context, exec Executor, payloads []models.ClientObject, table models.Table) (int, error)
+	IngestObjects(ctx context.Context, tx Transaction, payloads []models.ClientObject, table models.Table) (int, error)
```
Pascal-Delange, line 16: This is something I wanted to do for a while: this repository method _expects_ a transaction as executor, to make sure the operations are done atomically. In other cases, it's up to the caller how he uses it.

repositories/db_executor_getter.go
```diff
@@ -20,10 +21,21 @@ type databaseSchemaGetter interface {
...
 
+type Transaction interface {
+	databaseSchemaGetter
+	Exec(ctx context.Context, sql string, arguments ...any) (pgconn.CommandTag, error)
+	Query(ctx context.Context, sql string, args ...any) (pgx.Rows, error)
+	QueryRow(ctx context.Context, sql string, args ...any) pgx.Row
+	// Begin(ctx context.Context) (pgx.Tx, error)
+	Commit(ctx context.Context) error
+	Rollback(ctx context.Context) error
+	RawTx() pgx.Tx
```
Pascal-Delange, line 36: Adding this was really the main initial objective of this PR. However, to be able to do this I needed to distinguish between just passing a generic pool/tx, or specifically passing a transaction (and the check should be at compile time, I don't want runtime errors "oh I was expecting a transaction").
It just made sense to take the time and rework it so we have finer control on what usecase/repo methods that _receive_ an executor, actually expect a transaction.

repositories/db_executor.go
```diff
@@ -9,30 +9,69 @@ import (
...
-func (e ExecutorPostgres) QueryRow(ctx context.Context, sql string, args ...any) pgx.Row {
+func (e PgExecutor) QueryRow(ctx context.Context, sql string, args ...any) pgx.Row {
 	return e.exec.QueryRow(ctx, sql, args...)
 }
+
+////////////////////////////////////
+// Transaction
+////////////////////////////////////
+
+type PgTx struct {
```
Pascal-Delange, line 46: we could, as an extension and when we need it, make this PgTx struct (and the interface it implements in `repositories/db_executor_getter.go`) also expose the other methods from pgx.Tx (or something slighly more adapted to our use case), so that we could for instance manually create checkpoints, rollback to checkpoints, etc.
But for now just having the additional `RawTx()` method allows to distinguish the interfaces.

usecases/decision_workflows/decision_workflows.go
```diff
@@ -43,7 +43,7 @@ type caseAndDecisionRepository interface {
 type webhookEventCreator interface {
 	CreateWebhookEvent(
 		ctx context.Context,
-		tx repositories.Executor,
+		tx repositories.Transaction,
```
Pascal-Delange, line 46: "CreateWebhookEvent" should be used with a transaction.
Then, by transitivity so should "AutomaticDecisionToCase"

balzdur (approved):

PR #704, opened 2024-10-10
Task queue with river for async decisions

Pascal-Delange:
## Doc
Doc page is [WIP on notion](https://www.notion.so/checkmarble/Job-queue-for-async-tasks-120d1b7bc7c48028a099e52c27c20a5a#120d1b7bc7c480849d66f189a7f0b19c).

## Prerequisites
This PR builds on https://github.com/checkmarble/marble-backend/pull/703, which I needed to extract raw pgx transactions (deployed to staging last week and planned to release in prod this week).

## Deployment
This changes somewhat how Marble is deployed, most importantly for our managed offering.
The execution of async tasks in 4 distinct cloud run jobs should go, replaced by one cloud run service with "always provision cpu" option on.
I'll start by manually pausing the scheduling of the existing cloud run jobs while I deploy this on staging.

For our self-hosted users too, they are invited to run the worker (now common for task queue and cron jobs) rather than schedule the jobs separately. This corresponds to what the docker-compose shared in the marble repo suggests, so it should not be a problem (I'm leaving the "--cron-scheduler" entrypoint for backward compatibility, though usage of "--worker" is preferred).

## Next up
I prefer to let this run on staging this week (and take time to run proper tests) before aiming for a release next week.

## Workflow
See [notion](https://www.notion.so/checkmarble/Job-queue-for-async-tasks-120d1b7bc7c48028a099e52c27c20a5a?pvs=4#120d1b7bc7c480638151fea4dfe42e9d) for a flowchart of the lifetime of a scheduled execution.

A tricky part, which was much easier when all decisions where created at once sequentially, was to keep some level of udpating the scheduled execution's progress, without polling the db too aggressively (in a context where we easily expect 10ks to 100ks of decisions created in a batch). I proposed something in `func (w *AsyncDecisionWorker) possiblyUpdateScheduledExecNumbers(`, but I'm open to suggestions.

----

Sorry I did not manage to more elegantly cut this into smaller changes - but I'm happy to run you through the changes if it's useful.


Pascal-Delange:
Note: I had some transient errors on the tests, I wasn't able to replicate them systematically. (related to the liveness check that now, since last weeks, calls the db)

repositories/dbmodels/db_decisions_to_create.go
```diff
@@ -20,13 +20,13 @@ const TABLE_DECISIONS_TO_CREATE = "decisions_to_create"
 
 var DecisionToCreateFields = utils.ColumnList[DecisionToCreate]()
 
-func AdaptDecisionToCrate(db DecisionToCreate) models.DecisionToCreate {
+func AdaptDecisionToCrate(db DecisionToCreate) (models.DecisionToCreate, error) {
```
balzdur, line 23: Typo
```suggestion
func AdaptDecisionToCreate(db DecisionToCreate) (models.DecisionToCreate, error) {
```

And I'm not a fan of adding a "nil" error. It feels like the function is failable from the signature, but it do not happen.

balzdur (approved):
IIUC, the "data source" for what is pending for async exec is the same than before (= with cron tasks) so no need for a specific migration task, is that right?

PR #715, opened 2024-10-21
middlewares for task queue jobs: logger, recover, tracer

Pascal-Delange:


usecases/webhook_events_usecase.go
```diff
@@ -117,12 +115,6 @@ func (usecase WebhookEventsUsecase) SendWebhookEventAsync(ctx context.Context, w
 	go func() {
 		ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), ASYNC_WEBHOOKS_SEND_TIMEOUT)
 		defer cancel()
-		tracer := utils.OpenTelemetryTracerFromContext(ctx)
```
Pascal-Delange, line 120: I'm removing this because, as it's run asynchronously, it sometimes results in super long spans (waiting for the "internal" rate limiting for convoy), which make the spans UI unreadable

jobs/middlewares.go
```diff
@@ -0,0 +1,178 @@
...
+func (m LoggerMiddleware) aggregateAndLogError(ctx context.Context, job *rivertype.JobRow, err error) {
+	m.errorCountLock.Lock()
+	defer m.errorCountLock.Unlock()
+
+	errorKey := fmt.Sprintf("%s:%s", job.Kind, err.Error())
+	m.errorCount[errorKey]++
+
+	if m.errorCount[errorKey] == 1 {
+		go func() {
+			time.Sleep(sentryErrorGroupingTime)
```
Pascal-Delange, line 70: Try to do some level of aggregation before sending errors to sentry, if errors occur on batch jobs. I don't know that the current value of 30sec is enough (in fact, it probably isn't) but let's see.

balzdur (approved):

PR #716, opened 2024-10-21
set timeout on all routes

Pascal-Delange:
- set a shared timeout (5 sec by default) on all routes, except those where we explicitly set a longer one
- change the signatures of presentError and usecasesWithCreds (see comment below, it's related to gin and its middlewares and the details of how `c.Request.Context()` works
- do not log errors (to sentry) with `LogAndReportSentryError` that are caused by a deadlined/canceled context

--- 

you may want to read the 3rd commit separately (you'll understand when you see it)


api/handle_analytics.go
```diff
@@ -13,14 +13,15 @@ import (
 
 func handleListAnalytics(uc usecases.Usecases) func(c *gin.Context) {
 	return func(c *gin.Context) {
+		ctx := c.Request.Context()
```
Pascal-Delange, line 16: There is a lot of this change in this PR.
The motivation is the following: if code is executed in the handler _after gin moved on to the next middleware_ (which may happen, for instance because of the timeout middleware) then c.Request.Context() will no longer return the same context (for instance trace and credential keys will be missing).
So the request middleware should be extract as the first thing in the handlers, before taking any possibly long-lived actions.

thynquest (approved):

PR #721, opened 2024-10-23
try uuid v7 for better data locality in index

Pascal-Delange:
Inserts of decisions and decision rules are still slower than I'd like them to be.
I tried some things already but that didn't quite work.
I'd like to try this, using (ordered by time) v7 uuids for better data locality in indexes. See [link](https://www.percona.com/blog/uuids-are-popular-but-bad-for-performance-lets-discuss/) for discussion.

NB: this means dropping the assumption, that we have anyway not applied on all tables since a while, that we do uuids with a shared prefix for all ids in a given org. But that's fine as this wasn't really useful in the end.

### DIFF
--- a/models/decision.go
+++ b/models/decision.go
@@ -5,7 +5,7 @@ import (
 	"time"
 
 	"github.com/checkmarble/marble-backend/models/ast"
-	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/google/uuid"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


balzdur (approved):
Should we migrate existing ones ? or it's pointless
Should we use it everywhere in the end (and kill `pure_utils.NewPrimaryKey` ?)

Pascal-Delange:
> Should we migrate existing ones ? or it's pointless
No I think that's pointless - it's relevant for writes IIUC, not for reads (the point is to make index writing faster, no to speed up index serving). And anyway the amount of new data is going to be greater than the amount of old data soon enough

> Should we use it everywhere in the end
perhaps for the sake of uniformity, though it's probably mainly going to make a difference on "large" tables. So I first want to try it here

> (and kill `pure_utils.NewPrimaryKey` ?)
yep probably

PR #722, opened 2024-10-25
Franck/implement backtest mode

thynquest:
This is a first step of implementing backtest mode on marble. it is focused on the main steps rather than the details; there are still modifications to add in order to have the feature completely implemented.


utils/context_credentials.go
```diff
@@ -39,17 +39,28 @@ func OrganizationIdFromRequest(request *http.Request) (organizationId string, er
...
+			return "", errors.Wrap(
+				models.ForbiddenError,
+				"An organizationId must be passed in the request query params for MARBLE_ADMIN to use this endpoint")
+		}
 		return "", errors.Wrap(
 			models.ForbiddenError,
 			"Unexpected error: credentials does not grant access to any organization")
 	}
 
+	// if creds.OrganizationId == "" && creds.Role == models.MARBLE_ADMIN {
```
Pascal-Delange, line 53: what going on here ? 👀

api/routes.go
```diff
@@ -86,6 +86,8 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 
 	router.GET("/scenario-publications", tom, handleListScenarioPublications(uc))
 	router.POST("/scenario-publications", tom, handleCreateScenarioPublication(uc))
+	router.POST("/scenarios-publications-backtest", tom,
```
Pascal-Delange, line 89: Hmm, ok here's an open though about publications.
Without discussing the details of how the publications log is currently stored (not necessarily in the most elegant way), the intention behind it is that we store a full history of what version was active from when to when, for audit purposes.
While that serves a purpose for "real live versions", I don't think that's necessary for "shadow test versions".
So perhaps we can skip the whole procedure of "storing a scenario publication", and as a consequence just use a usecase/endpoint pair that would be essentially called "start test mode"
What do you think ?

Pascal-Delange:
Dispo en fin de matinée si tu veux en reparler

utils/context_credentials.go
thynquest: Yes I have forgotten to remove the comments ..

Pascal-Delange: un petit coup de cleanup :)

repositories/migrations/20241029013218_scenario_testrun.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE scenario_testrun(
```
Pascal-Delange, line 3: I'm going to sound like a super annoying guy, but can we please make it `scenario_test_run`? (here and in the other places) 🙏

models/scenario_testrun.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	Up TestrunStatus = iota
+	Down
+	Unknown
+)
+
+func (t TestrunStatus) String() string {
+	switch t {
+	case Up:
+		return "Up"
```
Pascal-Delange, line 16: I'm fine with "up" and "down" as statuses, but let's please keep them lower case (we have a convention, in practice even if not in documentation, that statuses in DB/code are represented in lower snake case)

api/handle_scenario_testrun.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"net/http"
+
+	"github.com/checkmarble/marble-backend/dto"
+	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/checkmarble/marble-backend/utils"
+	"github.com/gin-gonic/gin"
+)
+
+func handleScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
```
Pascal-Delange, line 12:
```suggestion
func handleCreateScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
```

dto/scenario_testrun.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/checkmarble/marble-backend/models"
+)
+
+type CreateScenarioTestRunBody struct {
+	ScenarioIterationId string `json:"scenario_iteration_id"`
+	ScenarioId          string `json:"scenario_id"`
+	Period              string `json:"period"`
+}
+
+type ScenarionResp struct {
```
Pascal-Delange, line 15: I'm pretty sure we're going to a need a more complete DTO for test runs (at the latest when we want to display them in the app with their details), so I'd say we create a complete dto mapping for them in this PR already (and return it in the api handler)

api/handle_scenario_testrun.go
```diff
@@ -0,0 +1,37 @@
...
+		input, err := dto.AdaptCreateScenarioTestRunBody(data)
+		if presentError(ctx, c, err) {
+			return
+		}
+		scenarioTestRun, err := usecase.ActivateScenarioTestRun(ctx, organizationId, input)
+		if presentError(ctx, c, err) {
+			return
+		}
+		result := dto.AdaptScenarioTestRunDto(scenarioTestRun)
+		c.JSON(http.StatusOK, result)
```
Pascal-Delange, line 35:
```suggestion
		c.JSON(http.StatusOK, gin.H{
			"scenario_test_run": result,
		})
```
(a convention we've been trying to follow, even if it's not applied everywhere)

repositories/scenario_testrun.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type ScenarioTestRunRepository interface {
+	CreateTestRun(ctx context.Context, tx Transaction, testrunID string,
+		input models.ScenarioTestRunInput) error
+	GetByScenarioIterationID(ctx context.Context, exec Executor, scenarioID string) (models.ScenarioTestRun, error)
+	GetByID(ctx context.Context, exec Executor, testrunID string) (models.ScenarioTestRun, error)
+}
+
+type ScenarioTestRunRepositoryPostgresql struct{}
```
Pascal-Delange, line 19: here as well we have a bit of a mix of two conventions in the code, but I'd rather push you in the direction of: define new methods on the struct `MarbleDbRepository` (see any one of the more "recent" repositories, for instance the case_repository.
This way we have one concrete implementation of the repository, and we can just use minimal interfaces at the usecase level

Pascal-Delange: (then you may have to make the method names more complete however)

repositories/scenario_testrun.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func (repo *ScenarioTestRunRepositoryPostgresql) GetByScenarioIterationID(ctx context.Context,
+	exec Executor, scenarioIterationID string,
+) (models.ScenarioTestRun, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return models.ScenarioTestRun{}, err
+	}
+	return SqlToModel(
```
Pascal-Delange, line 63: Using the helper method SqlToModel will return an error wrapping `models.NotFoundError` if no row is scanned (see https://github.com/checkmarble/marble-backend/blob/main/repositories/sql_to_row.go#L85), seeing how you're planning to use it in the usecase you may want to use `SqlToOptionalModel` or `SqlToListOfModels` (and adapt the signature accordingly)

usecases/scenario_testrun_usecase.go
```diff
@@ -0,0 +1,69 @@
...
+	input models.ScenarioTestRunInput,
+) (models.ScenarioTestRun, error) {
+	exec := usecase.executorFactory.NewExecutor()
+	// we should not have any existing testrun for this scenario
+	existingTestrun, err := usecase.repository.GetByScenarioIterationID(ctx, exec, input.ScenarioIterationId)
+	if err != nil {
+		return models.ScenarioTestRun{}, errors.Wrap(err,
+			"error while fecthing entries to find an existing testrun")
+	}
+	if existingTestrun.ScenarioIterationId != "" && existingTestrun.Status == models.Up {
```
Pascal-Delange, line 35: ok for the idea, but see comment in the repo for the actual implem of GetByScenarioIterationID

usecases/scenario_testrun_usecase.go
```diff
@@ -0,0 +1,69 @@
...
+		return models.ScenarioTestRun{}, errors.Wrap(models.ErrTestRunAlreadyExist,
+			fmt.Sprintf("the scenario %s has a running testrun", input.ScenarioId))
+	}
+
+	// we should have a live version running
+	scenarioPublications, errScenarioPubs := usecase.scenarioPublicationsRepository.ListScenarioPublicationsOfOrganization(
```
Pascal-Delange, line 41: Sorry, in fact the "scenario publications" model serves more as an "audit log" of sorts of the past activations & deactivations of scenarios. You should more simply just check the liveVersionId field on the scenario.

Pascal-Delange: other than that, the logic is ok for me

usecases/scenario_testrun_usecase.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/checkmarble/marble-backend/usecases/security"
+	"github.com/pkg/errors"
+)
+
+type ScenarioTestRunUsecase struct {
+	transactionFactory             executor_factory.TransactionFactory
+	executorFactory                executor_factory.ExecutorFactory
+	scenarioPublicationsRepository repositories.ScenarioPublicationRepository
+	enforceSecurity                security.EnforceSecurityScenario
+	clientDbIndexEditor            clientDbIndexEditor
```
Pascal-Delange, line 20: we're not using this in the end :-) (you probably have a linter warning displayed)
we should check similarly to usecases/scenario_publications_usecase.go lines 89 to 100 if a "data preparation" (AKA index creation) is required.
Then, we have two options for the actual creation:
- either we trigger an index creation from the app, and only then allow to click on "start a test run" (that's what we currently have when we publish a scenario)
- or we go a step further and instrument this within the test run lifecycle, using a task queue for the async update.
We can have a chat about this in the afternoon if you'd like. If it makes things easier, it's ok for me to add this in a separate PR but it should be taken into account before the feature is released in prod.

Pascal-Delange (approved):
Some comments left in the PR, a bit about formatting and one concerning more work to do (possibly in another PR).
We can chat about it after 2pm.

repositories/migrations/20241029013218_scenario_testrun.sql
thynquest: ok no problem.

usecases/security/enforce_security_testrun.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+type EnforceSecurotyTestRunImpl struct {
+	EnforceSecurity
+	Credentials models.Credentials
+}
+
+func (e *EnforceSecurotyTestRunImpl) CreateTestRun(organizationId string) error {
+	return errors.Join(
+		e.Permissions(models.BUILDER_PERMISSIONS),
```
Pascal-Delange, line 20: Ok this works, but I'm wondering if it's necessary 🤔 
I might have misled you by saying "you need to be builder to have the permission". Perhaps just something like `e.Permission(models.SCENARIOS_CREATE),`  would suffice (which is contained in the builder role)

usecases/scenario_testrun_usecase.go
```diff
@@ -13,46 +13,41 @@ import (
...
 	organizationId string,
 	input models.ScenarioTestRunInput,
 ) (models.ScenarioTestRun, error) {
+	if err := usecase.enforceSecurity.CreateTestRun(organizationId); err != nil {
+		return models.ScenarioTestRun{}, err
+	}
 	exec := usecase.executorFactory.NewExecutor()
 	// we should not have any existing testrun for this scenario
-	existingTestrun, err := usecase.repository.GetByScenarioIterationID(ctx, exec, input.ScenarioIterationId)
+	existingTestrun, err := usecase.repository.GetTestRunByScenarioIterationID(ctx, exec, input.ScenarioIterationId)
```
Pascal-Delange, line 33: just thinking about this again: considering how the repository method is implemented, you have no guarantee here that _if an existing test run is found_, it's the most recent one (considering the condition you then apply line 38).
The safest thing would be to also pass a filter on "up" status test runs to the repo.

Pascal-Delange (approved):
two last comments, looks good

usecases/scenario_testrun_usecase.go
thynquest: good point.. which means there are two choices
- the one is about to apply what you just suggested, which means the method name should change (something like `GetActiveTestRunByScenarioID` )
- the other choice raise another question: is it possible to have many testruns for a given scenario_iteration_id ? if it is, then we could return all the testruns for a given scenario_iteration_id and check if there is any which has the status `up` ; in which case the method name would be `GetListTestrunScenarioByIterationID` 
what do you think ?

Pascal-Delange: Yes, that is correct.
Both work for me, so go for the one that seems simplest to you.
For what it's worth: we'll need some sort of "list test runs" repo method anyway soon.

usecases/scenario_testrun_usecase.go
thynquest: for the moment, I'd prefer to add this in another PR.. I will keep this in mind for the release of the feature

Pascal-Delange: yes, as discussed

repositories/scenario_testrun.go
```diff
@@ -54,25 +54,48 @@ func (repo *ScenarioTestRunRepositoryPostgresql) CreateTestRun(ctx context.Conte
...
 	if err := validateMarbleDbExecutor(exec); err != nil {
-		return models.ScenarioTestRun{}, err
+		return nil, err
 	}
-	return SqlToModel(
+	query := NewQueryBuilder().
+		Select(dbmodels.SelectScenarioTestRunColumns...).
+		From(dbmodels.TABLE_SCENARIO_TESTRUN + " AS tr").
+		Join(dbmodels.TABLE_SCENARIO_ITERATIONS + " AS scit ON scit.id = tr.scenario_iteration_id").
+		Join(dbmodels.TABLE_SCENARIOS + " AS sc ON sc.id = scit.scenario_id").
```
Pascal-Delange, line 84: you could also just filter on scenario_iterations.scenario_id directly

Pascal-Delange (approved):
cool, thanks 👍

usecases/decision_phantom/decision_phantom.go
```diff
@@ -54,9 +68,9 @@ func (usecase *PhantomDecisionUsecase) CreatePhantomDecision(ctx context.Context
 	if testRunScenarioExecution.ScenarioId == "" {
 		return models.PhantomDecision{}, err
 	}
-	decision := models.AdaptScenarExecToPhantomDecision(testRunScenarioExecution)
-	for i := range decision.RuleExecutions {
-		decision.RuleExecutions[i].Evaluation = nil
+	phantom_decision := models.AdaptScenarExecToPhantomDecision(testRunScenarioExecution)
```
Pascal-Delange, line 71: no big issue but let's stick with snake case variable names :)

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -194,16 +186,16 @@ func EvalTestRunScenario(ctx context.Context,
...
+	testRunIteration, err := repositories.EvalTestRunScenarioRepository.GetTestRunIterationByScenarioId(
 		ctx, exec, params.Scenario.Id)
 	if err != nil {
 		return models.ScenarioExecution{}, errors.Wrap(err,
 			"error getting testrun scenario iteration in EvalTestRunScenario")
 	}
-	if testRunIteration.Id == "" {
-		return models.ScenarioExecution{}, nil
+	if testRunIteration == nil {
+		return models.ScenarioExecution{}, models.ErrNoTestRunFound
```
Pascal-Delange, line 196: ok, but considering how you're using this method in the decisions usecase, this will result in error logs all the time, no ? Since you decided to have all the logic of "checking if there even is a test run to execute" for a decision in this usecase, I think the "nothing to do" case should be handled differently.

usecases/scenario_testrun_usecase.go
```diff
@@ -27,6 +28,27 @@ func (usecase *ScenarioTestRunUsecase) ActivateScenarioTestRun(ctx context.Conte
...
+			"Error while fetching indexes to create in ActivateScenarioTestRun")
+	}
+	if len(indexesToCreate) > 0 {
+		return models.ScenarioTestRun{}, errors.Wrap(
+			models.ErrScenarioIterationRequiresPreparation,
+			fmt.Sprintf("Cannot activate the scenario in test mode: it requires data preparation to be run first for %d indexes", len(indexesToCreate)),
+		)
+	}
+
+	errIdx := usecase.clientDbIndexEditor.CreateIndexesAsync(ctx, organizationId, indexesToCreate)
```
Pascal-Delange, line 47: We have a bit of a problem here, with this implementation: the method is async (as its name suggests) so it doesn't wait for the index creation to be done to return.
However, we don't wont the test run execution to start before the indexes are finished creating.
2-3 options:
- all of the test run creation becomes async, and there is no immediate user feedback (other than "ok taken into account) when he clicks "start test run" => definitely bad
- we create the test run in an intermediate status, and update its status to "active" after the index creation is finished => much better IMO
- we take inspiration from the way it's currently done in the builder: we require the indexes to be present before the user can click "activate" => better than 1 but I don't think it's the solution here. Pretty sure we should go for 2.

usecases/scheduled_execution/run_scheduled_execution.go
```diff
@@ -148,56 +150,117 @@ func (usecase *RunScheduledExecution) executeScheduledScenario(ctx context.Conte
...
 		models.TableIdentifier{Table: scenario.TriggerObjectType, Schema: db.DatabaseSchema().Schema},
 	)
-
 	objectIds, err := usecase.ingestedDataReadRepository.ListAllObjectIdsFromTable(ctx, db, scenario.TriggerObjectType, filters...)
 	if err != nil {
 		return err
 	}
-
 	nbPlannedDecisions := len(objectIds)
-	err = usecase.repository.UpdateScheduledExecution(ctx, exec, models.UpdateScheduledExecutionInput{
```
Pascal-Delange, line 193: I don't know if it was voluntary, but you removed the call to this repo update

usecases/scheduled_execution/run_scheduled_execution.go
```diff
@@ -148,56 +150,117 @@ func (usecase *RunScheduledExecution) executeScheduledScenario(ctx context.Conte
...
 	}
-
 	return usecase.insertAsyncDecisionTasks(
 		ctx,
 		scheduledExecution.Id,
 		scheduledExecution.Scenario,
 	)
 }
 
+func (usecase *RunScheduledExecution) processDecisionTask(ctx context.Context,
```
Pascal-Delange, line 160: is this an unfinished refactor ?

usecases/scenario_testrun_usecase.go
thynquest: if we do not want to execute the test run before the indexes has been created, we should make it blocking.

thynquest: ideally we should make it blocking but given the way it has been written, I do not see how we can make sure the indexes have been created correctly before doing anything else. unless we add a new feature to do so

usecases/scheduled_execution/run_scheduled_execution.go
thynquest: Yes it is not used anymore

usecases/scenario_testrun_usecase.go
Pascal-Delange: IMO, we need call a method that will: set the test run to an intermediate "preparation pending" (or such) status, starts the index creation, rolls back the status if there is an error or updates the status to "running" if the index creation succeeds. The easiest way of doing that would probable be to have another method like "createIndexesAsync" that accepts a pair of callbacks. But keep it simple, we'll touch this again very soon anyway.

Making it blocking isn't much of a solution because we can't just wait for the endpoint to respond for more than seconds.

integration_test/batch_ingestion_and_execution_test.go
```diff
@@ -151,7 +151,7 @@ func createDecisionsBatch(
 			assert.FailNow(t, "Failed to get scheduled execution", err)
 		}
 	}
-	if se.Status != models.ScheduledExecutionSuccess {
+	if se.Status != models.ScheduledExecutionProcessing {
```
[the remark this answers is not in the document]
thynquest: my bad .. I corrected it

repositories/dbmodels/db_scenario_testrun.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/checkmarble/marble-backend/utils"
+)
+
+type DBScenarioTestRun struct {
+	Id                  string        `db:"id"`
+	ScenarioIterationId string        `db:"scenario_iteration_id"`
+	CreatedAt           time.Time     `db:"created_at"`
+	Period              time.Duration `db:"period"`
+	Status              string        `db:"status"`
+}
```
balzdur, line 16: I think we need to save the couple of iteration ids used on each backtest (name TBD):
- the `liveVersionID`
- the `phantomVersionID`

This is used for at least to good reasons :
1. Easilly display the couple of versions when we list past testruns
2. Handle special cases where a testrun is "up" but the live version has changed (still to define, but IMO we should "down" the testrun when the associated live version id deactivated)

Bonnus: I would also store a "reason" to explain why the testrun goes from `up` -> `down` (if we do automatically close some testruns, it could help users understand the internal logic). But it's not mandatory in a V0.

Pascal-Delange (approved):
works for me 👍  (minor comments left in the PR)

Left to do, but can be in a separate PR (actually, let's please do it in another PR):
- logic to stop a test run if any (un)publish action happens on the scenario
- discuss Thomas' suggestion also left on this PR (I'll make a separate comment elsewhere)
- in this current status, a test run can remain stuck in "pending" - but that's ok as we plan to go back to this soon (as soon as the V0 of back-front exchange is finished) to make the index creation controlled by a task queue

repositories/dbmodels/db_scenario_testrun.go
Pascal-Delange: Though, on second thought: I think I'm against storing the live version id on the test run, as this is nasty data denormalization, and if anything bad happens it's bound to create a form of data corruption (not the absolutely worst type of it, since this is "analytics", but still).
We can discuss this one tomorrow.

Pascal-Delange: (nb: or we store the "reference iteration id" - which is almost the same thing 99% of the time, but has another interpretation)

PR #723, opened 2024-10-28
fix: typehint to let postgres prepare statement

Pascal-Delange:
I learned something new on how pgx works, it first runs a prepare statement on all queries (by default) where the query is parsed but the parameter values are not used. Hence, pg is not able to parse `CONCAT('str_a', $1, 'str_b')` because it can't know the type of $1 which will only be known at query runtime.
In other situations, pg is able to infer the type, for instance if operators like `=` are used.
The recommended fix is to add type hints where query is possibly ambiguous.

In practice, this error was happening if the right-hand side of the equation contained a constant (it works with a field identifier).

### DIFF
--- a/.github/workflows/deploy_production.yaml
+++ b/.github/workflows/deploy_production.yaml
@@ -6,7 +6,7 @@ on:
       - "v[0-9]+.[0-9]+.[0-9]+"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy_production.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - "v[0-9]+.[0-9]+.[0-9]+"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
Pascal-Delange, line 9: recommended by github actions extension as equivalent and safer

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -76,6 +76,12 @@ func EvalScenario(
 	logger.InfoContext(ctx, "Evaluating scenario", "scenarioId", params.Scenario.Id)
 	exec := repositories.ExecutorFactory.NewExecutor()
 
+	// If the scenario has no live version, don't try to Eval() it, return early
```
Pascal-Delange, line 79: The check was happening too late, causing a panic at line 90 (new line numbers) if a decision was called on the scenario and it was not active.

balzdur (approved):

PR #724, opened 2024-10-28
fix: better error handling on transfercheck

Pascal-Delange:
Goal of this PR: make those error messages go away.
Capture d’écran 2024-10-28 à 12 46 15
To be clear they do deserve monitoring, but not in this format.

### DIFF
--- a/api/present_error.go
+++ b/api/present_error.go
@@ -14,6 +14,8 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+const timeoutMst = "Sorry, the API timed out. Please try again later."
+
 func presentError(ctx context.Context, c *gin.Context, err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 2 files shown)


api/present_error.go
```diff
@@ -43,10 +45,10 @@ func presentError(ctx context.Context, c *gin.Context, err error) bool {
 
 	case errors.Is(err, context.DeadlineExceeded):
 		logger.WarnContext(ctx, fmt.Sprintf("Deadline exceeded: %v", err))
-		c.JSON(http.StatusRequestTimeout, errorResponse)
```
Pascal-Delange, line 46: If the API times out, I don't want to leak logic of where it happened exactly in the API response (though TBH our code is public so it's not as bad as it could be).

usecases/transfer_check_usecase.go
```diff
@@ -515,9 +513,7 @@ func (usecase *TransferCheckUsecase) ScoreTransfer(
 		},
 	)
 	if err != nil {
-		return models.Transfer{}, errors.Wrapf(
```
Pascal-Delange, line 518: The idea of this was that I don't want to leak any special errors coming from the decision usecase that could be interpreted as a distinct type of HTTP status in present_error to the transfercheck usecase.
However, I realize that context canceled errors are a counter-example for this.

balzdur (approved):

PR #726, opened 2024-11-01
first batch of work: separate DBs per customer (optional)

Pascal-Delange:


thynquest (approved):

repositories/migrations/20241105131500_drop_constraint_enum_values copy.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE organizations
+ADD COLUMN IF NOT EXISTS use_marble_db_schema_as_default BOOLEAN NOT NULL DEFAULT FALSE;
```
Pascal-Delange, line 4: The trick here is that I'll create the column and backfill it to true for existing orgs, before deploying the service.

PR #727, opened 2024-11-04
fix bug when a field named "id" is created on the data model

Pascal-Delange:


repositories/migrations/20241104103800_drop_constraint_enum_values.sql
```diff
@@ -0,0 +1,18 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE data_model_enum_values
```
Pascal-Delange, line 3: The constraint was causing problems with replication, because the data_model_enum_values table doesn't have a primary key (see https://www.postgresql.org/docs/current/logical-replication-publication.html)

thynquest (approved):

PR #737, opened 2024-11-06
fix bug: batch execution stuck if no rows match the trigger condition

Pascal-Delange:


usecases/scheduled_execution/run_scheduled_execution.go
```diff
@@ -198,10 +198,6 @@ func (usecase *RunScheduledExecution) insertAsyncDecisionTasks(
 		return err
 	}
 
-	if len(objectIds) == 0 {
```
Pascal-Delange, line 201: Turns out this was too smart: early exit skipped the insertion of the job that does the batch exec status handling, at no gain (because `len()==0` is not really a special case anyway)

thynquest (approved):

PR #741, opened 2024-11-13
fix: because PG_CONNECTION_STRING is an option, PG_PASSWORD no longer…

Pascal-Delange:
### Part 1 (commit (7c1af89))
A previous PR allowed to connect to postgres by passing a connection string rather than host/db/password/user separately, but the latter environment variables were still required, even if they were not used.

### Part 2: RSA private key handling, Goal and context
Copy pasted from the discussion on community slack (in DM) with the guy from Reyts:

Ok, I think I managed to replicate and fix your issue.
So, what is happening is that the code, as it is currently, expects a RSA private key in .pem format and beginning with `-----BEGIN RSA PRIVATE KEY-----....`, AKA PKCS#1 format.
Depending on how you generated your private/public key pari with openssl, you may have instead a file starting with `-----BEGIN PRIVATE KEY-----...` , AKA PKCS#8 format, with the algorithm being set in the key body as RSA (if it contains anything else than a RSA key, we are not treating it currently).

The problem that you probably encounter is that you are in the second case (`BEGIN PRIVATE KEY`), and this case was until now not handled well.

I am releasing a fix for this issue, which will be released in the next prod release next week.
In the meantime, your options are:
- keep the variable empty, this really should work
- generate a key in PKCS#1 format (AKA "BEGIN RSA PRIVATE KEY") and use it
- or, this should work now but I don't recommend to do it systematically, use the :latest tag on our backend docker image, as soon as I have merged the PR that I'm working on (I'll ping you to let you know)

### Others
- increase the timeout for async enum value insertion (let's debug later if it's taking longer than it should)
- decrease logging level if logger not found in context, super noisy in tests otherwise

### DIFF
--- a/infra/parse_signing_key.go
+++ b/infra/parse_signing_key.go
@@ -7,6 +7,8 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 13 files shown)


infra/parse_signing_key.go
```diff
@@ -48,13 +64,33 @@ func parsePrivateKey(privateKeyString string) *rsa.PrivateKey {
 	// when a multi-line env variable is passed to the docker container by docker-compose, it escapes the newlines
 	privateKeyString = strings.Replace(privateKeyString, "\\n", "\n", -1)
 	block, _ := pem.Decode([]byte(privateKeyString))
-	if block == nil || block.Type != "RSA PRIVATE KEY" {
+	// if block == nil || block.Type != "RSA PRIVATE KEY" {
+	// 	panic(errors.New("failed to decode PEM block containing RSA private key"))
+	// }
```
balzdur, line 69:
```suggestion
```

balzdur (approved):

infra/parse_signing_key.go
```diff
@@ -7,28 +7,44 @@ import (
...
 		logger.InfoContext(ctx, "No AUTHENTICATION_JWT_SIGNING_KEY received, generating a new key pair. This is not recommended for production environments as the key will not be stable in time")
-		privateKey, err := generatePrivateKey(2048)
+		privateKey, err := generateRsaPrivateKey(2048)
 		if err != nil {
 			panic("Error generating RSA private key")
 		}
 		return privateKey
 	}
 
+	if keyFile != "" {
```
thynquest, line 30: this means that keyfile had precedence over privatekey, is it what we want ?

thynquest (commented):
I have only one question .. otherwise it looks good to me.

infra/parse_signing_key.go
Pascal-Delange: Yea that's fine I think - anyway there is no reason at all to provide both, so I'm comfortable letting this at the customer's responsibility.

PR #742, opened 2024-11-14
first batch: allow to use null values in the AST, and return them as an error at the top level if null bubbles up

Pascal-Delange:
## Behavior
- null values "bubble up" the AST of rule execution
- basically any operator that gets a nil value in one of its expected children, returns nil to its parent
- by exception, a series of `OR` (resp `AND`) returns `true` (resp. `false`) if it is "the obvious thing to return" (`true OR NULL = true`... see code)
- `isEmpty` operator coming soon in another PR
- a rule returns a `models.ErrNullFieldRead` error if a null value reaches the "top" of the AST

> ⚠️  An aggregation operator now returns nil if at least one filter value is nil, where it previously returned an error. As a consequence, depending on the other conditions in OR above, a rule may now return a hit where it returned an error previously. This is technically a breaking change but seems closer to what people should expect.

--- 

Paired with https://github.com/checkmarble/marble-frontend/pull/595

### DIFF
--- a/dto/decision_dto.go
+++ b/dto/decision_dto.go
@@ -48,6 +48,7 @@ type DecisionRule struct {
 	Result        bool      `json:"result"`
 	RuleId        string    `json:"rule_id"`
 	ScoreModifier int       `json:"score_modifier"`
+	ErrorCode     *int      `json:"error_code"`
 
 	// RuleEvaluation is not returned by default, it only is for endpoints consumed by the frontend
 	RuleEvaluation *ast.NodeEvaluationDto `json:"rule_evaluation,omitempty"`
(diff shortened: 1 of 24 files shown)


usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -274,6 +274,9 @@ func evalScenarioRule(
 	dataModel models.DataModel,
 	snoozes []models.RuleSnooze,
 ) (int, models.RuleExecution, error) {
+	if rule.Name == "Check on account name" {
+		fmt.Println("ok")
+	}
```
balzdur, line 279: Probably to remove

balzdur (approved):

usecases/evaluate_scenario/evaluate_scenario.go
Pascal-Delange: oupsies, yes

PR #743, opened 2024-11-19
implement isEmpty, isNotEmpty operators

Pascal-Delange:
Goes with PR https://github.com/checkmarble/marble-frontend/pull/602 on the frontend.
Implements the IsEmpty and IsNotEmpty operators, but only on the main AST (not for aggregator filters).
The operator returns true if the value is NULL or empty string, else false.

Also remove the unused and badly defined `NumberOfArguments` property of FuncAttributes (previously returned by the API used by the front for the list of operators). The source of truth for this is in the implementation of `Evaluate` for each function in `ast_eval`.

Tested on unit decisions with API, also tested that the batch run still works with isEmpty operators in the trigger condition (to test the SQL pre-filter)

Nb: the same filter for aggregates has voluntarily not been implemented now (mostly because I think it's less urgent and it's also just late for today)

### DIFF
--- a/dto/func_attributes_dto.go
+++ b/dto/func_attributes_dto.go
@@ -5,15 +5,13 @@ import (
 )
 
 type FuncAttributesDto struct {
-	Name              string   `json:"name"`
-	NumberOfArguments int      `json:"number_of_arguments,omitempty"`
-	NamedArguments    []string `json:"named_arguments,omitempty"`
+	Name           string   `json:"name"`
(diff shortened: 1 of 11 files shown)


balzdur (approved):
LGTM, one remark: to stick with existing convention on operators, I would add some tests on evaluate

Pascal-Delange:
> LGTM, one remark: to stick with existing convention on operators, I would add some tests on evaluate

Allez, j'avoue hier soir il était tard j'ai eu la flemme, mais je fais ça

PR #744, opened 2024-11-20
fix: batch exec should use the schedueld exec's version

Pascal-Delange:


usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -252,13 +253,19 @@ func EvalScenario(
 	)
 	defer span.End()
 
-	liveVersion, err := repositories.EvalScenarioRepository.GetScenarioIteration(ctx, exec, *params.Scenario.LiveVersionID)
+	var targetVersionId string
+	if params.ScenarioIterationId != nil {
+		targetVersionId = *params.ScenarioIterationId
+	} else {
+		targetVersionId = *params.Scenario.LiveVersionID
+	}
```
balzdur, line 261: I would move this up and check that `targetVersionId` is not nil on 240 (+ call it `versionToRunId` based on the above naming)

balzdur (approved):

usecases/evaluate_scenario/evaluate_scenario.go
Pascal-Delange: yep that makes sense 👍

PR #746, opened 2024-11-22
Franck/decision data

thynquest:
this is a first implementation of the endpoint(s) we are going to use in order to display the distribution of decisions for a given scenario

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -57,7 +57,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
 		handlePostAllDecisions(uc, marbleAppHost))
-	router.GET("/decisions/data", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsData(uc, marbleAppHost))
+	router.GET("/decisions/data", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsData(uc))
 	router.GET("/decisions/:decision_id", tom, handleGetDecision(uc, marbleAppHost))
 	router.GET("/decisions/:decision_id/active-snoozes", tom, handleSnoozesOfDecision(uc))
 	router.POST("/decisions/:decision_id/snooze", tom, handleSnoozeDecision(uc))
--- a/api/routes.go
+++ b/api/routes.go
@@ -30,7 +30,7 @@ const (
 
 	// TODO: temporary workaround, what we need to do is to make the endpoint reliably respond quickly even
 	// without warmup
-	LIST_DECISION_TIMEOUT = 10 * time.Second
+	LIST_DECISION_TIMEOUT = 30 * time.Second
 
 	DEFAULT_TIMEOUT = 5 * time.Second
 )
--- a/api/routes.go
+++ b/api/routes.go
@@ -53,9 +53,6 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
 		handlePostAllDecisions(uc, marbleAppHost))
-	// query in the form localhost:8085/decisions/databyoutcome?scenario_id='123-abc'&created_ad=1992-03-15&expires_at=1992-03-15"
-	router.GET("/decisions/databyoutcome", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsDataByOutcome(uc))
-	router.GET("/decisions/databyscore", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsDataByScore(uc))
 	router.GET("/decisions/:decision_id", tom, handleGetDecision(uc, marbleAppHost))
 	router.GET("/decisions/:decision_id/active-snoozes", tom, handleSnoozesOfDecision(uc))
 	router.POST("/decisions/:decision_id/snooze", tom, handleSnoozeDecision(uc))
@@ -69,6 +66,9 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/scenarios", tom, createScenario(uc))
 	router.GET("/scenarios/:scenario_id", tom, getScenario(uc))
 	router.PATCH("/scenarios/:scenario_id", tom, updateScenario(uc))
+
+	router.GET("/testrun/:testrun_id/databyscore", timeoutMiddleware(LIST_DECISION_TIMEOUT),
+		handleDecisionsDataByOutcomeAndScore(uc))
 
 	router.GET("/scenario-iterations", tom, handleListScenarioIterations(uc))
 	router.POST("/scenario-iterations", tom, handleCreateScenarioIteration(uc))
--- a/api/routes.go
+++ b/api/routes.go
@@ -57,6 +57,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
 		handlePostAllDecisions(uc, marbleAppHost))
+	router.GET("/decisions/data", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsData(uc, marbleAppHost))
 	router.GET("/decisions/:decision_id", tom, handleGetDecision(uc, marbleAppHost))
 	router.GET("/decisions/:decision_id/active-snoozes", tom, handleSnoozesOfDecision(uc))
 	router.POST("/decisions/:decision_id/snooze", tom, handleSnoozeDecision(uc))
--- a/api/routes.go
+++ b/api/routes.go
@@ -57,7 +57,9 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
 		handlePostAllDecisions(uc, marbleAppHost))
-	router.GET("/decisions/data", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsData(uc))
+	// query in the form localhost:8085/decisions/databyoutcome?scenario_id='123-abc'&created_ad=1992-03-15&expires_at=1992-03-15"
+	router.GET("/decisions/databyoutcome", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsDataByOutcome(uc))
+	router.GET("/decisions/databyscore", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsDataByScore(uc))
 	router.GET("/decisions/:decision_id", tom, handleGetDecision(uc, marbleAppHost))
 	router.GET("/decisions/:decision_id/active-snoozes", tom, handleSnoozesOfDecision(uc))
 	router.POST("/decisions/:decision_id/snooze", tom, handleSnoozeDecision(uc))
--- a/dto/decision_dto.go
+++ b/dto/decision_dto.go
@@ -142,13 +142,10 @@ func ProcessDecisionDataDtoFromModels(inputs []models.DecisionsByVersionByOutcom
 	result := make([]DecisionData, len(inputs))
 	for i, input := range inputs {
 		item := DecisionData{
-			Version:        input.Version,
-			Outcome:        input.Outcome,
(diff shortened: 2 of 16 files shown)


dto/decision_dto.go
```diff
@@ -72,6 +72,12 @@ type PivotValue struct {
 	PivotId    null.String `json:"pivot_id"`
 }
 
+type DecisionData struct {
```
Pascal-Delange, line 75: This format is sufficient for the first graph described in figma, but not sufficient for the next one (histogram of decisions by score). IMO we should expose the more granular data on the backend, and then the frontend can perform an additional aggregation to display the right data ?

repositories/decisions_repository.go
```diff
@@ -71,6 +71,29 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
 	return decisions[0], nil
 }
 
+func (repo *MarbleDbRepository) DecisionsByOutcome(ctx context.Context, exec Executor,
+	scenarioID string,
+) ([]models.DecisionsByVersionByOutcoume, error) {
+	query := NewQueryBuilder().
+		Select("d.outcome, d.scenario_version, Count(d.outcome) as total").
+		From(fmt.Sprintf("%s AS d", dbmodels.TABLE_DECISIONS)).
+		Join(fmt.Sprintf("%s as sp on sp.scenario_iteration_id = d.scenario_iteration_id",
```
Pascal-Delange, line 80: why the join with scenario publications ?
I think we just need to join decisions and scenario_iterations, and filter by decision created_at

Pascal-Delange (commented):
I left a couple of comments, other than that it's broadly going in the direction that I had in mind.
What do you think about caching of the values on this ?

repositories/decisions_repository.go
```diff
@@ -72,16 +73,41 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
...
-		Join(fmt.Sprintf("%s as sp on sp.scenario_iteration_id = d.scenario_iteration_id",
-			dbmodels.TABLE_SCENARIOS_PUBLICATIONS)).
-		Where(fmt.Sprintf("sp.publication_action = %s", models.Publish)).Where(squirrel.Eq{
-		"sp.scenario_id": scenarioID,
-	}).GroupBy("d.outcome, d.scenario_version")
+		Select("d.outcome, d.scenario_version, d.score, Count(d.outcome) as total").
+		From(fmt.Sprintf("%s AS d", dbmodels.TABLE_DECISIONS)).Where(squirrel.Eq{
+		"d.scenario_id": scenarioID,
+	}).Where(squirrel.Eq{
+		"d.created_at": begin.String(),
```
Pascal-Delange, line 83: If I'm not mistaken, you want to use squirrel.Gt (or squirrel.GtOrEq), and vice versa below

Pascal-Delange: (or simply write out the inequalities in raw sql with placeholders in `.Where(...)`)

repositories/decisions_repository.go
```diff
@@ -71,6 +72,54 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
...
+		func(row pgx.CollectableRow) (models.DecisionsByVersionByOutcoume, error) {
+			db, err := pgx.RowToStructByPos[dbmodels.DbDecisionsByOutcome](row)
+			if err != nil {
+				return models.DecisionsByVersionByOutcoume{}, err
+			}
+			return dbmodels.AdaptDecisionByOutcome(db), nil
+		})
+}
+
+func (repo *MarbleDbRepository) DecisionsByScore(ctx context.Context, exec Executor,
```
Pascal-Delange, line 99: As far as I understand, those two sql queries return exactly the same results (in fact, the only difference is `count(d.score)` vs `count(d.outcome)`, which is equivalent here seeing as it counts rows)
Either way, do we need the two ?
(if we do want a version that groups only by outcome, which is to be discussed with Thomas and Kevin tomorrow, it should have a less granular GROUP BY condition)

api/routes.go
```diff
@@ -57,6 +57,9 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
 		handlePostAllDecisions(uc, marbleAppHost))
+	// query in the form localhost:8085/decisions/databyoutcome?scenario_id='123-abc'&created_ad=1992-03-15&expires_at=1992-03-15"
+	router.GET("/decisions/databyoutcome", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleDecisionsDataByOutcome(uc))
```
Pascal-Delange, line 61: two questions here on the API:
- would it make sense to have one endpoint that returns all the stats, both for real decisions and phantom decisions related to a given test run ? I'm open to both approaches and we can discuss it tomorrow, but I'm a bit afraid of API endpoint inflation (and they would always be used together)
- either way, and this is more of a detail, if the API exposes the before/after filters on decisions, my feeling is that they should be called accordingly (maybe "created_before"/"created_after"): created_at and expires_at, relating to the name of fields on the test run, do not by themselves have meaning in the context of an endpoint on decisions

models/decision.go
```diff
@@ -52,6 +52,13 @@ type DecisionWithRuleExecutions struct {
 	RuleExecutions []RuleExecution
 }
 
+type DecisionsByVersionByOutcoume struct {
```
Pascal-Delange, line 55:
```suggestion
type DecisionsByVersionByOutcome struct {
```
small typo here also

api/routes.go
```diff
@@ -67,6 +67,9 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.GET("/scenarios/:scenario_id", tom, getScenario(uc))
 	router.PATCH("/scenarios/:scenario_id", tom, updateScenario(uc))
 
+	router.GET("/testrun/:testrun_id/databyscore", timeoutMiddleware(LIST_DECISION_TIMEOUT),
```
Pascal-Delange, line 70: let's put a higher timeout (say 30 sec, at least to give us some buffer while we test and release the feature)

usecases/decision_usecase.go
```diff
@@ -103,6 +108,22 @@ func (usecase *DecisionUsecase) GetDecision(ctx context.Context, decisionId stri
...
 
+func (usecase *DecisionUsecase) GetDecisionsByOutcomeAndScore(ctx context.Context,
+	testrunId string,
+) ([]models.DecisionsByVersionByOutcome, error) {
+	testrun, errTestRun := usecase.scenarioTestRunRepository.GetTestRunByID(ctx,
+		usecase.executorFactory.NewExecutor(), testrunId)
+	if errTestRun != nil {
+		return nil, errTestRun
+	}
+	decisions, err := usecase.repository.DecisionsByOutcomeAndScore(ctx,
```
Pascal-Delange, line 119: reading this, there is one last thing we have forgotten I think (or perhaps it's more a remark that concerns https://github.com/checkmarble/marble-backend/pull/756): the test run may have stopped before its initially set expiry date.
So two options:
- either we update testrun.Expires when we effectively stop a test run (change to do in the other PR)
- or we use a distinct field to keep a trace of the initially planned stop date and the actual stop date
(I think option 1 is fine, and in this case we don't need to change anything here)

models/scenario_testrun.go
```diff
@@ -40,7 +40,8 @@ func ScenarioTestStatusFrom(s string) TestrunStatus {
 type ScenarioTestRun struct {
 	ScenarioIterationId string
 	ScenarioId          string
-	Period              time.Duration
+	CreatedAt           time.Time
+	Expires             time.Time
```
Pascal-Delange, line 44:
```suggestion
	ExpiresAt             time.Time
```

repositories/decisions_repository.go
```diff
@@ -71,6 +72,31 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
 	return decisions[0], nil
 }
 
+func (repo *MarbleDbRepository) DecisionsByOutcomeAndScore(ctx context.Context, exec Executor,
+	scenarioID string, begin, end time.Time,
+) ([]models.DecisionsByVersionByOutcome, error) {
+	query := NewQueryBuilder().
```
Pascal-Delange, line 78: really sorry, but this sql query is not correct 🙈 
if you do 
```
FROM decisions
JOIN phantom_decisions USING(scenario_id)
```
you will get millions of rows because it's basically a cartesian product (assuming thousands of decisions and phantom decisions).

What you really want to do is a `UNION ALL`, though really it would be much simpler to just have two separate queries (though they would look almost the same) for the real decisions and the phantom decisions.
Then we get a list of aggregates for the real decisions, a list of aggregates for the phantom decisions, and we can return them together in the dto slice.

models/decision.go
```diff
@@ -52,6 +52,17 @@ type DecisionWithRuleExecutions struct {
 	RuleExecutions []RuleExecution
 }
 
+type DecisionsByVersionByOutcome struct {
```
Pascal-Delange, line 55: see the remark on the repository: this is actually a regression, the initial model (and dto) with version,outcome,score,total count was the correct one - and the results for both versions of the scenario should be unioned rather than joined.

repositories/decisions_repository.go
```diff
@@ -75,19 +75,30 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
...
-		From(fmt.Sprintf("%s AS d", dbmodels.TABLE_DECISIONS)).Join(
-		dbmodels.TABLE_PHANTOM_DECISIONS + " AS pd on pd.scenario_id = d.scenario_id").Where(squirrel.Eq{
-		"d.scenario_id": scenarioID,
-	}).Where(squirrel.GtOrEq{
+	decisionQuery := NewQueryBuilder().
+		Select("outcome, scenario_version, score").
+		From(dbmodels.TABLE_DECISIONS).Where(squirrel.GtOrEq{
+		"d.created_at": begin.String(),
+	}).Where(squirrel.LtOrEq{
+		"d.deleted_at": end.String(),
```
Pascal-Delange, line 83:
```suggestion
		"d.created_at": end.String(),
```

Pascal-Delange: same on the one below

repositories/decisions_repository.go
```diff
@@ -71,6 +72,42 @@ func (repo *MarbleDbRepository) DecisionWithRuleExecutionsById(ctx context.Conte
...
+		From(dbmodels.TABLE_PHANTOM_DECISIONS).Where(squirrel.GtOrEq{
+		"d.created_at": begin.String(),
+	}).Where(squirrel.LtOrEq{
+		"d.deleted_at": end.String(),
+	})
+	query, err := WithUnionAll(decisionQuery, phantomDecisionQuery)
+	if err != nil {
+		return nil, err
+	}
+	unionQuery, _, _ := query.ToSql()
```
Pascal-Delange, line 96: By doing this and reusing unionQuery directly below, you're losing the params passed as arguments (begin.String() and end.string(), twice).
Instead, you can:
- do the aggregation (group by) in each subquery and then UNION ALL them
- or kill line 96 and replace `From(unionQuery + " AS q")` below by `FromSelect(query)` (see https://github.com/checkmarble/marble-backend/blob/main/repositories/case_repository.go#L31 for instance)

Pascal-Delange: => gotta run this endpoint/query for real locally before we merge it, because I think as it is it returns an error

Pascal-Delange (commented):
Last comment on the SQL query side

repositories/decisions_repository.go
thynquest: this is the corresponding raw sql query I tested locally without error
````sql
select q.outcome,q.scenario_version, q.score, Count(q.outcome) as total from (select outcome, scenario_version, score from decisions where created_at>='2011-01-01' and created_at<='2012-01-01' union all select outcome, scenario_version, score from phantom_decisions where created_at>='2011-01-01' and created_at<='2012-01-01') as q group by scenario_version,outcome,score;
````

usecases/decision_usecase.go
thynquest: noted.. let's put this one on the testrun, on another pr..

repositories/decisions_repository.go
Pascal-Delange: ok, apologies I understand.
Then however that means that the begin/end parameters end up as raw sql in the query, which is not great for another reason. In this case there is no risk of SQL injection of course (which is the n°1 argument for passing params as args rather than code them into the query), but doing so still helps us get proper statistics on queries usage, execution plans, runtime stats etc. (it also helps the DB reuse query plans, but that's perhaps less impactful).

Pascal-Delange: Hmm as a matter of fact I'm not able to reproduce your output with `where created_at>='2011-01-01'` 🤔  (using either version)

PR #747, opened 2024-11-25
fix permissions: create transfercheck user

Pascal-Delange:
Transfercheck user creation was broken in the last refactor of the create user usecase permissions.

### DIFF
--- a/usecases/security/enforce_security_user.go
+++ b/usecases/security/enforce_security_user.go
@@ -37,6 +37,13 @@ func (e *EnforceSecurityUserImpl) CreateUser(input models.CreateUser) error {
 		)
 	}
 
+	if input.Role == models.TRANSFER_CHECK_USER && e.Credentials.Role != models.MARBLE_ADMIN {
+		return errors.Wrap(
+			models.ForbiddenError,
+			"only marble admins can create transfer check users",
+		)
+	}
+
 	// should already be handled by the fact that only the ADMIN & MARBLE_ADMIN roles have the
 	// MARBLE_USER_CREATE permission, but make double sure
 	if input.Role == models.ADMIN &&
@@ -44,15 +51,6 @@ func (e *EnforceSecurityUserImpl) CreateUser(input models.CreateUser) error {
 		return errors.Wrap(
 			models.ForbiddenError,
 			"only org admins and marble admins can create org admins",
-		)
-	}
-
-	if input.PartnerId != nil && (e.Credentials.Role != models.MARBLE_ADMIN ||
-		e.Credentials.PartnerId == nil ||
-		*e.Credentials.PartnerId != *input.PartnerId) {
-		return errors.Wrap(
-			models.ForbiddenError,
-			"only marble admins can create users with partner_id",
 		)
 	}
 
@@ -70,6 +68,13 @@ func (e *EnforceSecurityUserImpl) UpdateUser(targetUser models.User, updateUser 
 		return errors.Wrap(
 			models.BadParameterError,
 			"only marble admins can create marble admins")
+	}
+
+	if updateUser.Role != nil &&
+		*updateUser.Role == models.TRANSFER_CHECK_USER {
+		return errors.Wrap(
+			models.BadParameterError,
+			"cannot update an existing user to TRANSFER_CHECK_USER")
 	}
 
 	// Only org admins and marble admins can create org admins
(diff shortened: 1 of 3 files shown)


usecases/user_usecase.go
```diff
@@ -26,6 +26,16 @@ func (usecase *UserUseCase) AddUser(ctx context.Context, createUser models.Creat
...
 
+	if createUser.Role == models.TRANSFER_CHECK_USER && createUser.PartnerId == nil {
+		return models.User{}, errors.Wrap(models.BadParameterError,
+			"PartnerId is required for TRANSFER_CHECK_USER role")
+	}
+
+	if createUser.PartnerId != nil && createUser.Role != models.TRANSFER_CHECK_USER {
+		return models.User{}, errors.Wrap(models.BadParameterError,
+			"Partner Id is only allowed for TRANSFER_CHECK_USER role")
+	}
```
balzdur, line 37: I think splitting user roles per app and check on an array here, will be better.
It will allow to create differnet role types for TC (if the product ask for it) easilly

balzdur (approved):

usecases/user_usecase.go
Pascal-Delange: what do you mean exactly, have a different usecase for the two apps ?

balzdur: More or less something like that, so if we create role management in transfercheck, it will not be a burdon (ATM, there is only user and api roles)
```suggestion
	if slices.Contains(models.GetValidTransfercheckUserRoles(), createUser.Role) && createUser.PartnerId == nil {
		return models.User{}, errors.Wrap(models.BadParameterError,
			"PartnerId is required for TRANSFER_CHECK_USER role")
	}

	if slices.Contains(models.GetValidMarbleUserRoles(), createUser.Role) && createUser.PartnerId != nil {
		return models.User{}, errors.Wrap(models.BadParameterError,
			"Partner Id is only allowed for TRANSFER_CHECK_USER role")
	}
```

thynquest (approved):

usecases/user_usecase.go
```diff
@@ -51,13 +51,34 @@ func (usecase *UserUseCase) AddUser(ctx context.Context, createUser models.Creat
...
+func validateUserCreate(createUser models.CreateUser) error {
+	switch {
+	case slices.Contains(models.GetValidTransfercheckUserRoles(), createUser.Role):
+		if createUser.PartnerId == nil {
+			return errors.Wrap(models.BadParameterError,
+				"PartnerId is required for transfercheck users")
+		}
+
+		return nil
+	case slices.Contains(models.GetValidUserRoles(), createUser.Role):
```
balzdur, line 70: I would rename this for `models.GetValidMarbleCoreUserRoles()` (or alike)

usecases/user_usecase.go
```diff
@@ -51,13 +51,34 @@ func (usecase *UserUseCase) AddUser(ctx context.Context, createUser models.Creat
...
+		}
+
+		return nil
+	default:
+		return errors.Wrap(models.BadParameterError, "Invalid role received")
+	}
+}
+
 func (usecase *UserUseCase) UpdateUser(ctx context.Context, updateUser models.UpdateUser) (models.User, error) {
 	if updateUser.Role != nil && !slices.Contains(models.GetValidUserRoles(), *updateUser.Role) {
```
balzdur, line 83: I think you still need to do the same kind of function validator here

usecases/security/enforce_security_user.go
```diff
@@ -72,6 +70,13 @@ func (e *EnforceSecurityUserImpl) UpdateUser(targetUser models.User, updateUser
...
 	}
 
+	if updateUser.Role != nil &&
+		*updateUser.Role == models.TRANSFER_CHECK_USER {
+		return errors.Wrap(
+			models.BadParameterError,
+			"cannot update an existing user to TRANSFER_CHECK_USER")
+	}
+
 	// Only org admins and marble admins can create org admins
```
balzdur, line 80: The comment is missleading (based on the condition)

balzdur (commented):
IMO, you should also have the same kind of switch statement inside`enforce_security_user.go` to ease distinguish business logic for each set of roles but that sounds overkill for now.

Pascal-Delange:
> IMO, you should also have the same kind of switch statement inside`enforce_security_user.go` to ease distinguish business logic for each set of roles but that sounds overkill for now.

@balzdur TBH I think you're right about all those, I'm just a bit short on time to do this as well as I would ideally like :/ 
I'll come back to this later

PR #756, opened 2024-11-26
Franck/backtest mode handle publish unpublish

thynquest:
I prefered to make a separate branch for this one


Pascal-Delange:
👍  makes sense, I took the liberty to change the target branch to [franck/implement-backtest-mode](https://github.com/checkmarble/marble-backend/tree/franck/implement-backtest-mode)
I'll check it tonight, school time now

repositories/scenario_testrun.go
```diff
@@ -90,6 +116,21 @@ func (repo *MarbleDbRepository) GetActiveTestRunByScenarioIterationID(ctx contex
 	)
 }
 
+func (repo *MarbleDbRepository) GetTestRunByLiveVersionID(ctx context.Context,
```
Pascal-Delange, line 119: there can be more than one, so we need to be more specific here - I haven't finished reading, but probably the query should order by createdAt desc and take the most recent one / apply logic ?

usecases/scenario_testrun_usecase.go
```diff
@@ -82,6 +82,9 @@ func (usecase *ScenarioTestRunUsecase) ActivateScenarioTestRun(ctx context.Conte
 		return models.ScenarioTestRun{}, models.ErrWrongIterationForTestRun
 	}
 
+	// keep track of the live version associated to the current testrung
```
Pascal-Delange, line 85:
```suggestion
	// keep track of the live version associated to the current testrun
```

usecases/scenarios/scenario_publisher.go
```diff
@@ -99,6 +100,20 @@ func (publisher ScenarioPublisher) PublishOrUnpublishIteration(
 	return scenarioPublications, nil
 }
 
+func (publisher ScenarioPublisher) shutDownTestRunIfNeeded(ctx context.Context, tx repositories.Transaction, liveVersionId string) error {
+	testrun, errTestrun := publisher.ScenarioTestRunRepository.GetTestRunByLiveVersionID(ctx, tx, liveVersionId)
+	if errTestrun != nil {
+		return errTestrun
+	}
+	if testrun != nil && testrun.Status == models.Up {
```
Pascal-Delange, line 108: also include testrun.Status == pending I think (it probably won't happen really, just for completeness)

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -187,6 +189,24 @@ func EvalTestRunScenario(ctx context.Context,
 		),
 	)
 	defer span.End()
+	testrun, errTr := repositories.ScenarioTestRunRepository.GetTestRunByLiveVersionID(ctx, exec, *params.Scenario.LiveVersionID)
```
Pascal-Delange, line 192: IMO we can make things a bit simpler here (without looking up the scenario by its live version, as discussed there can be a situation where decisions are still being created on a previous version):
IF we receive a decision based on a version X
AND there is a test run in status "up" using "live version id" X
THEN we create a phantom decision on this test run
ELSE pass

Pascal-Delange: (that's an edge case though, broadly the logic is correct)

Pascal-Delange (approved):
A small comment on the repo, otherwise it's ok

models/scenario_testrun.go
```diff
@@ -38,14 +38,16 @@ func ScenarioTestStatusFrom(s string) TestrunStatus {
...
+	ScenarioIterationId     string
+	ScenarioId              string
+	ScenarioLiveIterationId string
+	ExpiresAt               time.Time
+	Status                  TestrunStatus
 }
 
 type ScenarioTestRunInput struct {
 	ScenarioIterationId string
+	LiveScenarioId      string
```
balzdur, line 50: IIUC
```suggestion
	ScenarioLiveIterationId      string
```

Pascal-Delange:
+ mentioned in the other PR (https://github.com/checkmarble/marble-backend/pull/746): update the expires at value if the test run stops before the end

usecases/evaluate_scenario/evaluate_scenario.go
thynquest: I think this case is already handled. I just completed by testing the status of the testrun 

````golang
if testrun == nil || testrun.Status != models.Up
````

PR #761, opened 2024-11-29
Franck/testrun data

thynquest:


api/routes.go
```diff
@@ -91,6 +91,8 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.GET("/scenario-publications", tom, handleListScenarioPublications(uc))
 	router.POST("/scenario-publications", tom, handleCreateScenarioPublication(uc))
 	router.POST("/scenarios-testrun", tom, handleCreateScenarioTestRun(uc))
+	router.GET("/scenario-testrun", tom, handleListScenarioTestRun(uc))
```
Pascal-Delange, line 94: cosmetic only request: please move the 3 endpoints related to test runs out of the bloc of endpoints related to scenario-publications (we can have another "bloc" just below)

dto/scenario_testrun.go
```diff
@@ -9,22 +9,33 @@ import (
...
 type ScenarioTestRunResp struct {
-	ScenarioId string    `json:"scenario_id"`
-	CreatedAt  time.Time `json:"created_at"`
-	ExpiresAt  time.Time `json:"expires_at"`
-	Status     string    `json:"status"`
+	Id              string `json:"id"`
+	ScenarioId      string `json:"scenario_id"`
+	RefIterationId  string `json:"ref_iteration_id"`
+	TestIterationId string `json:"test_iteration_id"`
+	StartDate       string `json:"start_date"`
```
Pascal-Delange, line 24: Hmm there is one thing which I don't exactly understand here, using time.Time for created_at/end_at seemed fine to me. 
Is there a specific goal behind it ?

api/handle_scenario_testrun.go
```diff
@@ -37,3 +38,38 @@ func handleCreateScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
...
+			return
+		}
+		scenarioID := c.Query("scenario_id")
+
+		usecase := usecasesWithCreds(ctx, uc).NewScenarioTestRunUseCase()
+		testruns, err := usecase.ListTestRunByScenarioId(ctx, organizationId, scenarioID)
+		if presentError(ctx, c, err) {
+			return
+		}
+		c.JSON(http.StatusOK, pure_utils.Map(testruns, dto.AdaptScenarioTestRunDto))
```
Pascal-Delange, line 56: just for the sake of following the convention in (most of) the codebase, please also wrap it in a `gin.H{"scenario_test_runs": ...}` to return

api/handle_scenario_testrun.go
```diff
@@ -37,3 +38,38 @@ func handleCreateScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
...
+	return func(c *gin.Context) {
+		ctx := c.Request.Context()
+		organizationId, err := utils.OrganizationIdFromRequest(c.Request)
+		if presentError(ctx, c, err) {
+			return
+		}
+		scenarioID := c.Query("scenario_id")
+
+		usecase := usecasesWithCreds(ctx, uc).NewScenarioTestRunUseCase()
+		testruns, err := usecase.ListTestRunByScenarioId(ctx, organizationId, scenarioID)
```
Pascal-Delange, line 52: We have a little pain point in most of the backend, that I have not had time to resolve yet, which is that we use "string" fields in models & dtos to contain UUIDs. 
Since the usecase here clearly expects a (non empty) value for scenarioID, let's please validate its presence and the format in the handler (especially here because scenario_id comes from the query, so could be seen as optional)

usecases/scenario_testrun_usecase.go
```diff
@@ -97,7 +97,41 @@ func (usecase *ScenarioTestRunUsecase) ActivateScenarioTestRun(ctx context.Conte
...
+) (models.ScenarioTestRun, error) {
+	testrun, err := usecase.repository.GetTestRunByID(ctx,
+		usecase.executorFactory.NewExecutor(), testRunId)
+	if err != nil {
+		return models.ScenarioTestRun{}, err
+	}
+	if testrun == nil {
+		return models.ScenarioTestRun{}, nil
+	}
+	return *testrun, nil
```
Pascal-Delange, line 136: here (and on every test run returned by the "list" method), let's run an `enforceSecurity.ReadTestRun(testRun)` (method to create) before returning them to the API (similarly to what is done in other usecases, see users for instance)

Pascal-Delange (commented):
I have a few small comments of things to complete, but the structure is ok for me :-)

dto/scenario_testrun.go
thynquest: it concerns the dto that will be sent back to the front.. we could live it with a time type but string seemed to be simpler

api/handle_scenario_testrun.go
thynquest: in this case the format is already specified in the oas specification.. which will perform a first step of validation.

usecases/scenario_testrun_usecase.go
```diff
@@ -104,26 +104,25 @@ func (usecase *ScenarioTestRunUsecase) ActivateScenarioTestRun(ctx context.Conte
...
+				return nil, err
+			}
 		}
 	}
 	return testruns, nil
 }
 
 func (usecase *ScenarioTestRunUsecase) GetTestRunById(ctx context.Context,
-	testRunId string,
+	testRunId, organizationId string,
```
Pascal-Delange, line 125: no longer needed with the change :-)

dto/scenario_testrun.go
```diff
@@ -6,14 +6,20 @@ import (
 	"github.com/checkmarble/marble-backend/models"
 )
 
+// type CreateScenarioTestRunBody struct {
```
Pascal-Delange, line 9: 🔥

dto/scenario_testrun.go
```diff
@@ -6,14 +6,20 @@ import (
...
+
 type CreateScenarioTestRunBody struct {
-	ScenarioIterationId string `json:"scenario_iteration_id"`
-	ScenarioId          string `json:"scenario_id"`
-	RefIterationId      string `json:"ref_iteration_id"`
-	PhantomIterationId  string `json:"phantom_iteration_id"`
-	StartDate           string `json:"start_date"`
-	EndDate             string `json:"end_date"`
-	Period              string `json:"period"`
+	PhantomIterationId string `json:"test_iteration_id"`
```
Pascal-Delange, line 20: this is just a lint blocker, you can ignore it if you want (but in the end I think it would make sense to align the model and the DTO)

api/handle_scenario_testrun.go
```diff
@@ -42,28 +42,29 @@ func handleCreateScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
...
-		c.JSON(http.StatusOK, pure_utils.Map(testruns, dto.AdaptScenarioTestRunDto))
+		c.JSON(http.StatusOK, gin.H{"list scenario testrun by scenario id": pure_utils.Map(
+			testruns, dto.AdaptScenarioTestRunDto)})
 	}
 }
 
 func handleGetScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
 	return func(c *gin.Context) {
 		ctx := c.Request.Context()
+		organizationId, err := utils.OrganizationIdFromRequest(c.Request)
```
Pascal-Delange, line 60: can be removed if we remove the orgId from the usecase input again

Pascal-Delange (approved):
LGTM, a few last cleanup comments

PR #762, opened 2024-12-03
implementing rule execution data

thynquest:
this is a first implementation of the retrieval of rule executions. the final sql request remains; still need the `stable` identifier field in order to finalize it


models/decision.go
```diff
@@ -79,6 +79,12 @@ type ScenarioExecution struct {
 	OrganizationId      string
 }
 
+type RuleExecutionStat struct {
```
Pascal-Delange, line 82: Here, I think we also need to include, at least, the rule id (or perhaps more conveniently the stable id, that will be introduced)

Pascal-Delange: (and in the dto, of course)

usecases/rules_usecase.go
```diff
@@ -46,6 +50,26 @@ func (usecase *RuleUsecase) ListRules(ctx context.Context, iterationId string) (
 		})
 }
 
+func (usecase *RuleUsecase) ListRuleExecution(ctx context.Context, testrunId string) ([]models.RuleExecutionStat, error) {
```
Pascal-Delange, line 53: we'll need to pass in the time boundaries too as filters (presumably coming from the test run limits)

repositories/rules.go
```diff
@@ -46,6 +46,14 @@ func (repo *MarbleDbRepository) ListRulesByIterationId(ctx context.Context, exec
 	)
 }
 
+func (repo *MarbleDbRepository) ListRulesExecutionByIterationId(ctx context.Context, exec Executor, iterationId string) ([]models.RuleExecutionStat, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+	// TO BE DEFINED
```
Pascal-Delange, line 53: Just a note, let's please do 2 separate queries this time to get the numbers for the "live" and "tested" versions rather than a union all, it's simpler to read and no longer to do for the db :-)

Pascal-Delange (commented):
I'll send you the PR concerning the "stable id" field real soon

Pascal-Delange:
See https://github.com/checkmarble/marble-backend/pull/763

Pascal-Delange (approved):
@thynquest  I pushed some changes on the SQL queries (+ created an index), I'll let you have a look tomorrow and merge it

PR #764, opened 2024-12-05
removing total_count for a has_next_page boolean

ChibiBlasphem:
COUNT is a heavy operation so we want to get rid of it.

Instead with return a "has_next_page" boolean which is calculated by taking LIMIT + 1 item.


Pascal-Delange:
I'll close it, replaced by https://github.com/checkmarble/marble-backend/pull/765

PR #765, opened 2024-12-05
update cases & decisions list dto: no more count, return has_next_page

Pascal-Delange:
## Context
Loosely inspired from https://github.com/checkmarble/marble-backend/pull/764, and prerequisite for https://github.com/checkmarble/marble-frontend/pull/617.

## Contains
- refactoring of the pagination model & dto (to better enforce at a higher level "now raw user input makes it into a sql query"
- the cases, decision list endpoints no longer return a total count (**breaking change**)
- the cases, decision list endpoints pagination input no longer accepts a `previous` or `next` (boolean) query params input when passing an `offset_id`. Instead, "`next`" type pagination is used by default (**breaking change**)
- the public endpoint for decisions no longer returns the `start_index`,`end_index` for a given page of items (**breaking change**)
- the list case and decision endpoints return a `has_next_page` field (boolean), to indicate to the client if there is a next page to retrieve. This is implemented by retrieving `limit+1` items at the usecase level, returning `limit` items to the client and the flag depends on the presence of a `+1` item.

## Doc
- https://docs.checkmarble.com/reference/get_decisions has been updated

Sorry it's a bit massive and hard to read, as you can see below... I struggled with the github security bot.

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -49,6 +49,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.GET("/ast-expression/available-functions", tom, handleAvailableFunctions)
 
 	router.GET("/decisions", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleListDecisions(uc, marbleAppHost))
+	router.GET("/decisions/internal", tom, handleListDecisionsInternal(uc, marbleAppHost))
 	router.POST("/decisions", timeoutMiddleware(models.DECISION_TIMEOUT), handlePostDecision(uc, marbleAppHost))
 	router.POST("/decisions/all",
 		timeoutMiddleware(SEQUENTIAL_DECISION_TIMEOUT),
--- a/api/routes.go
+++ b/api/routes.go
@@ -49,7 +49,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.GET("/ast-expression/available-functions", tom, handleAvailableFunctions)
 
 	router.GET("/decisions", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleListDecisions(uc, marbleAppHost))
-	router.GET("/decisions/internal", tom, handleListDecisionsInternal(uc, marbleAppHost))
+	router.GET("/decisions/with-ranks", tom, handleListDecisionsInternal(uc, marbleAppHost))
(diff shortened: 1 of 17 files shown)


api/routes.go
```diff
@@ -49,6 +49,7 @@ func addRoutes(r *gin.Engine, auth Authentication, tokenHandler TokenHandler, uc
 	router.GET("/ast-expression/available-functions", tom, handleAvailableFunctions)
 
 	router.GET("/decisions", timeoutMiddleware(LIST_DECISION_TIMEOUT), handleListDecisions(uc, marbleAppHost))
+	router.GET("/decisions/internal", tom, handleListDecisionsInternal(uc, marbleAppHost))
```
Pascal-Delange, line 52: new: two different endpoints. This one should be used by the front and still returns the start/end indexes of the returned values.

dto/pagination.go
```diff
@@ -2,56 +2,18 @@ package dto
...
 import "github.com/checkmarble/marble-backend/models"
 
-type PaginationAndSortingInput struct {
-	OffsetId string              `form:"offset_id"`
-	Previous bool                `form:"previous"`
-	Next     bool                `form:"next"`
-	Sorting  models.SortingField `form:"sorting"`
-	Order    models.SortingOrder `form:"order"`
-	Limit    int                 `form:"limit" binding:"max=100"`
+type PaginationAndSorting struct {
```
Pascal-Delange, line 5: I reworked the DTO and models for pagination, mainly to avoid the endless security alerts from CodeQL (I don't think our implementation had actual security issues, but it definitely made it easier to introduce vulnerabilities accidentally)

models/pagination_and_sorting.go
```diff
@@ -1,58 +1,115 @@
...
+	case SortingFieldCreatedAt:
+		return "created_at"
+	case SortingFieldUpdatedAt:
+		return "updated_at"
+	default:
+		return "unknown"
+	}
+}
+
+func SortingFieldFrom(s string) SortingField {
```
Pascal-Delange, line 43: Disclaimer: a limitation is that this implementation only works as-is if the field on which we order in the DB has the same name as the field on the DTO. This is fine for now I think, and we'd just need to introduce a mapping table in the future if we need something more advanced.

balzdur (approved):
LGTM, to be honnest, I skipped a lot of code (mostly focus on Case pagination)

PR #771, opened 2024-12-10
add starts_with and ends_with operator validation and evaluation

ChibiBlasphem:
- starts with: takes 2 operands of string type
- ends with: takes 2 operands of string type


usecases/ast_eval/evaluate/eval_string_starts_with.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cockroachdb/errors"
+
+	"github.com/checkmarble/marble-backend/models/ast"
+)
+
+type StringStartsWith struct {
```
Pascal-Delange, line 13: As discussed orally: let's put the two in just one class, receiving the function as input

Pascal-Delange:
Nb: presently this does not implement the usecase of "batch filtering in SQL for batch execution" (see `usecases/scheduled_execution/batch_filtering.go`).
That's totally fine but I just wanted to mention it here.
Perhaps add a comment in that file for future reference.

usecases/ast_eval/evaluate/eval_string_starts_ends_with.go
```diff
@@ -0,0 +1,86 @@
...
+
+	rightString, errString := adaptArgumentToString(rightAny)
+	rightList, errList := adaptArgumentToListOfStrings(rightAny)
+
+	if errString == nil {
+		return startsEndsWithString(left, rightString, f.Function)
+	} else if errList == nil {
+		return startsEndsWithListOfStrings(left, rightList, f.Function)
+	} else {
+		return MakeEvaluateError(errors.Wrap(
```
Pascal-Delange, line 45: small syntax sugar: you can use "errors.Wrapf(..." and avoid the fmt.Sprintf inside

usecases/ast_eval/evaluate/eval_string_starts_ends_with.go
```diff
@@ -0,0 +1,86 @@
...
+func startsEndsWithListOfStrings(left string, right []string, f ast.Function) (any, []error) {
+	if f == ast.FUNC_STRING_STARTS_WITH {
+		var containsElement bool
+		for _, r := range right {
+			if strings.HasPrefix(strings.ToLower(left), strings.ToLower(r)) {
+				containsElement = true
+				break
+			}
+		}
+		return containsElement, nil
```
Pascal-Delange, line 72: you could also
```suggestion
		for _, r := range right {
			if strings.HasPrefix(strings.ToLower(left), strings.ToLower(r)) {
				return true, nil
			}
		}
		return false, nil
```

usecases/ast_eval/evaluate/eval_string_starts_ends_with.go
```diff
@@ -0,0 +1,86 @@
...
+	} else if f == ast.FUNC_STRING_ENDS_WITH {
+		return strings.HasSuffix(strings.ToLower(left), strings.ToLower(right)), nil
+	} else {
+		return MakeEvaluateError(errors.New(fmt.Sprintf(
+			"StringStartsWith does not support %s function", f.DebugString())))
+	}
+}
+
+func startsEndsWithListOfStrings(left string, right []string, f ast.Function) (any, []error) {
+	if f == ast.FUNC_STRING_STARTS_WITH {
```
Pascal-Delange, line 64: this is fine, but this `if f == ... else if f == ... else ` form also works well as a switch/case format ("if" is fine, just wanted to point out the possibility to you)

Pascal-Delange (approved):
👍

PR #775, opened 2024-12-11
Add is rounded ast validation and evaluation

ChibiBlasphem:
- The is rounded function returns a boolean if the value is a multiple of the specified threshold.
- It returns an error is the threshold is not an int and it return false if the value is not a multiple of the threshold or is not a int

### DIFF
--- a/usecases/ast_eval/evaluate/eval_is_rounded.go
+++ b/usecases/ast_eval/evaluate/eval_is_rounded.go
@@ -1,0 +1,59 @@
+package evaluate
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"github.com/cockroachdb/errors"
+
+	"github.com/checkmarble/marble-backend/models/ast"
+)
+
+type IsRounded struct{}
+
+func (f IsRounded) Evaluate(ctx context.Context, arguments ast.Arguments) (any, []error) {
+	value, valueErr := AdaptNamedArgument(arguments.NamedArgs, "value", promoteArgumentToFloat64)
+	threshold, thresholdErr := AdaptNamedArgument(arguments.NamedArgs, "threshold", promoteArgumentToFloat64)
+
+	errs := MakeAdaptedArgsErrors([]error{valueErr, thresholdErr})
+	if len(errs) > 0 {
+		return nil, errs
+	}
+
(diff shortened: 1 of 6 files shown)


usecases/ast_eval/evaluate/eval_is_rounded.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/cockroachdb/errors"
+
+	"github.com/checkmarble/marble-backend/models/ast"
+)
+
+// List of string cleaning steps applied:
+// - normalize
+// - remove diacritics
+// - set to lower case
+// - only letters and numbers
```
Pascal-Delange: 🔥

usecases/ast_eval/evaluate/eval_is_rounded.go
```diff
@@ -0,0 +1,64 @@
...
+type IsRounded struct{}
```
Pascal-Delange, line 13:
```suggestion
type IsRoundNumber struct{}
```
proposition ?

usecases/ast_eval/evaluate/eval_is_rounded.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func downcastToInt64(n float64) (int64, bool) {
+	if n < math.MinInt64 || n > math.MaxInt64 {
+		return 0, false
+	}
+	if n != float64(int64(n)) {
+		return 0, false
+	}
+	return int64(n), true
```
Pascal-Delange, line 44: I don't think this does exactly what we want, because int(9.99999999)=9:
see https://go.dev/play/p/qCP-hSxP5Tg

I'd do something like (syntax to adjust, I'm writing without IDE):
```
r := math.Round(n)
if math.Abs(r-n) > 1e-8 {
     return 0, false
}
return int64(r)
```
(à voir pour le threshold mais je pense que ça le fait)

Pascal-Delange (commented):
With the renaming to "is multiple of", I'm getting slightly uneasy about the "threshold" named param, but I don't have a much better idea so let's roll with it.

Pascal-Delange (approved):
LGTM 💪  sorry for the back and forth on naming

PR #776, opened 2024-12-11
fix logic error on multiple test run creation + various renamings etc

Pascal-Delange:


models/scenario_testrun.go
```diff
@@ -52,6 +52,21 @@ type ScenarioTestRun struct {
...
+func (i ScenarioTestRunInput) CreateDbInput(liveIterationId string) ScenarioTestRunCreateDbInput {
+	return ScenarioTestRunCreateDbInput{
+		ScenarioId:         i.ScenarioId,
+		PhantomIterationId: i.PhantomIterationId,
+		LiveScenarioId:     liveIterationId,
+		EndDate:            i.EndDate,
+	}
+}
+
+type ScenarioTestRunCreateDbInput struct {
```
Pascal-Delange, line 69: I'm creating another struct here, because the input to the DB is computed in the usecase using the liveIterationId, which is not known in the DTO. Just adding the field to the struct as it was done before was error prone.

carere (approved):

PR #788, opened 2024-12-19
fix: analytics page availabe to viewers in backend

Pascal-Delange:
@balzdur do I remember correctly that our credentials contain the list of permissions (not the role) attached to the user ?


balzdur (approved):

balzdur:
@Pascal-Delange, we store both on the user session
```ts
export interface CurrentUser {
  organizationId: string;
  partnerId?: string;
  role: string;
  actorIdentity: {
    userId?: string;
    email?: string;
    firstName?: string;
    lastName?: string;
  };
  permissions: UserPermissions;
}
```

But if the question concerns the visibility of the page on the client, it uses the permission under the hood (so it should be good with backend change).

It may takes time to apply (since permissions are cached in the user session cookie).

PR #796, opened 2025-01-08
Adding organization_feature_access table & functionality

carere:


repositories/dbmodels/db_organization_feature_access.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"time"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/utils"
+)
+
+const TABLE_ORGANIZATION_FEATURE_ACCESS = "organization_feature_access"
+
+var SelectOrganizationFeatureAccessColumn = utils.ColumnList[models.OrganizationFeatureAccess]()
```
Pascal-Delange, line 12:
```suggestion
var SelectOrganizationFeatureAccessColumn = utils.ColumnList[DBOrganizationFeatureAccess]()
```
should be this instead (the "db" tags are used to map the field names, and of course the exact list of fields may differ between the DB DTO and the model)

repositories/migrations/20250108102844_add_organization_feature_access_table.sql
```diff
@@ -0,0 +1,25 @@
...
+    test_run VARCHAR(255) NOT NULL DEFAULT 'allow',
+    workflows VARCHAR(255) NOT NULL DEFAULT 'allow',
+    webhooks VARCHAR(255) NOT NULL DEFAULT 'allow',
+    rule_snoozed VARCHAR(255) NOT NULL DEFAULT 'allow',
+    roles VARCHAR(255) NOT NULL DEFAULT 'allow',
+    analytics VARCHAR(255) NOT NULL DEFAULT 'allow',
+    sanctions VARCHAR(255) NOT NULL DEFAULT 'allow',
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    deleted_at TIMESTAMP WITH TIME ZONE,
```
Pascal-Delange, line 15: we can do without a deleted_at field, there is no reason today to ever delete them

repositories/migrations/20250108102844_add_organization_feature_access_table.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE organization_feature_access (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    org_id UUID NOT NULL,
+    test_run VARCHAR(255) NOT NULL DEFAULT 'allow',
```
Pascal-Delange, line 6: I know I left some VARCHAR(nb) in the DB as well, but the latest standard is: use VARCHAR or TEXT only (unless it's a user input and there is a really good reason to limit the size of it in the db)

repositories/organization_repository.go
```diff
@@ -76,6 +81,34 @@ func (repo *OrganizationRepositoryPostgresql) CreateOrganization(
 				name,
 			),
 	)
+
+	if err == nil {
```
Pascal-Delange, line 85: prefer: if err != nil {...} above, then no indented block (unless really necessary)

repositories/organization_repository.go
```diff
@@ -76,6 +81,34 @@ func (repo *OrganizationRepositoryPostgresql) CreateOrganization(
...
+		err := ExecBuilder(ctx, exec, NewQueryBuilder().
```
Pascal-Delange, line 86: In fact, let's let the CreateOrganization repository method accept a list of allowed features (in fact, just sanctions & test run probably, see main review message) and use them as an input (with sensible defaults written somewhere)

models/organization_feature_access.go
```diff
@@ -0,0 +1,28 @@
...
+	Roles          FeatureAccess
+	Analytics      FeatureAccess
+	Sanctions      FeatureAccess
+	CreatedAt      time.Time
+	UpdatedAt      time.Time
+}
+
+type UpdateOrganizationFeatureAccessInput struct {
+	OrganizationId string
+	TestRun        FeatureAccess
```
Pascal-Delange, line 21: This one is a matter of taste, but we tend to prefer partial update (PATCH) to full overwrite update (PUT) in the API. Correspondingly, this means for update inputs we typically use structs with pointers, and in the repo we only call Set for the non-nil attributes that are passed.

models/organization_feature_access.go
```diff
@@ -0,0 +1,28 @@
...
+	RuleSnoozed    FeatureAccess
+	Roles          FeatureAccess
+	Analytics      FeatureAccess
+	Sanctions      FeatureAccess
+	CreatedAt      time.Time
+	UpdatedAt      time.Time
+}
+
+type UpdateOrganizationFeatureAccessInput struct {
+	OrganizationId string
```
Pascal-Delange, line 20: I think the OrgId in the struct is never used (because the id is passed separately to the repo). It's easy to check by using vscode to see where the variable is used.

api/routes.go
```diff
@@ -129,6 +130,9 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth Aut
 	router.GET("/organizations/:organization_id", tom, handleGetOrganization(uc))
 	router.PATCH("/organizations/:organization_id", tom, handlePatchOrganization(uc))
 	router.DELETE("/organizations/:organization_id", tom, handleDeleteOrganization(uc))
+	router.GET("/organizations/:organization_id/feature_access", tom, handleGetOrganizationFeatureAccess(uc))
+	router.PATCH("/organizations/:organization_id/feature_access", tom,
```
Pascal-Delange, line 134: CF comment on the model, about PUT vs PATCH (let's keep "patch", but let's adapt the logic to match it)

repositories/organization_repository.go
```diff
@@ -124,3 +157,43 @@ func (repo *OrganizationRepositoryPostgresql) DeleteOrganizationDecisionRulesAsy
...
+}
+
+func (repo *OrganizationRepositoryPostgresql) UpdateOrganizationFeatureAccess(ctx context.Context, exec Executor,
+	organizationId string, updateFeatureAccess models.UpdateOrganizationFeatureAccessInput,
+) error {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return err
+	}
+
+	query := NewQueryBuilder().
```
Pascal-Delange, line 186: cf comment on put vs patch above. You can have a look at the case repository for instance to take inspiration.

usecases/organization_usecase.go
```diff
@@ -109,3 +109,23 @@ func (usecase *OrganizationUseCase) DeleteOrganization(ctx context.Context, orga
 	)
 	return nil
 }
+
+func (usecase *OrganizationUseCase) GetOrganizationFeatureAccess(ctx context.Context,
+	organizationId string,
+) (models.OrganizationFeatureAccess, error) {
+	if err := usecase.enforceSecurity.CreateOrganization(); err != nil {
```
Pascal-Delange, line 116: This one should probably rather be like this
```suggestion
	if err := usecase.enforceSecurity.ReadOrganization(); err != nil {
```
(maybe you need to change the enforceSecurity that's injected, I'm not sure while looking at it on github).
This is because every used in the org should be allowed to read the feature access (since the front will request it)

Pascal-Delange (commented):
Thanks ! 
I left you some comments in the PR body.
The two main points I have to add here:
- at a high level, the idea is that the front no longer stores the license key, instead calls this new endpoint on the backend that will: read feature accesses, merge them with the features available on the license (which the backend validates with our license server), and return the merged set of features to the frontend
- as a result, while the API return DTO should contain all the relevant features, we only need to specifically store in the DB at the org level, those that we want to set at the org level. That is: sanctions, test run (as of today). So you can simplify by reducing the nb of features we store.
- last point, for orgs we create internally, the API should accept a list of features to add when the org is created, so we should add this input in the org creation API DTO, and pass it over to the org creation repository (end customers who self-host are not expected to use this)

Pascal-Delange (approved):

PR #799, opened 2025-01-14
AST: Lazy evaluate boolean child nodes.

apognu:
This adds circuit breaking to boolean evaluation.

An AND boolean resolves to false if the first operand is false, and an OR boolean resolves to true if the first operand is true. With this, second operands and skipped if they can have no sway on the node's result.

Circuit breaking is disabled when trying to validate a scenario, so all cases are considered.

### DIFF
--- a/integration_test/scenario_flow_test.go
+++ b/integration_test/scenario_flow_test.go
@@ -529,11 +529,11 @@ func createDecisions(
 		assert.Nil(t, ruleExecution.Error, "Expected error to be nil, got \"%s\"", ruleExecution.Error)
 	}
 
-	// Create a decision [APPROVE] with a division by zero
+	// Create a decision [DECLINE] with a division by zero
 	transactionPayloadJson = []byte(`{
 		"object_id": "{transaction_id}",
 		"updated_at": "2020-01-01T00:00:00Z",
-		"account_id": "{account_id_approve}",
+		"account_id": "{account_id_decline}",
 		"amount": 0
 	}`)
 	approveDivisionByZeroDecision := createAndTestDecision(ctx, t, transactionPayloadJson, table,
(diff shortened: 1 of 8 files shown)


usecases/ast_eval/evaluate_ast.go
```diff
@@ -19,19 +19,29 @@ func EvaluateAst(ctx context.Context, environment AstEvaluationEnvironment, node
...
-	evalChild := func(child ast.Node) ast.NodeEvaluation {
+	// Only interested in lazy callback which will have default value if an error is returned
+	attrs, _ := node.Function.Attributes()
+
+	evalChild := func(child ast.Node) (childEval ast.NodeEvaluation, evalNext bool) {
 		childEval, ok := EvaluateAst(ctx, environment, child)
+
 		if !ok {
 			childEvaluationFail = true
+			return
```
apognu, line 30: Should we stop evaluating at the first child returning an error, like this? There may be a good reason to continue evaluating other children if a previous one fails.

integration_test/scenario_flow_test.go
```diff
@@ -533,7 +533,7 @@ func createDecisions(
 	transactionPayloadJson = []byte(`{
 		"object_id": "{transaction_id}",
 		"updated_at": "2020-01-01T00:00:00Z",
-		"account_id": "{account_id_approve}",
+		"account_id": "{account_id_decline}",
```
Pascal-Delange, line 536: As a short term fix, definitely good enough (though fix the comment line 532 maybe).
In a longer term perspective, this is another case (not the first time this happened to me over the last year) where this integration test fails because of unexpected side effects, I'd be interested to discuss options to improve on it. It's really gotten too big and unfocused and hard to follow.

Pascal-Delange (approved):
nice 👍

apognu:
Will amend tomorrow morning with the discussed edits.

PR #800, opened 2025-01-16
AST: Allow children of commutative nodes to be reordered cost-wise.

apognu:
This adds several things:

 * Commutative functions are marked as such.
 * AST functions now have an associated cost depending on their computational complexity.
 * Evaluation will now reorder the children of a commutative node to execute the cheapest first.

This would have the benefit, thanks to the lazy evaluation of AST nodes (circuit breaking), to improve overall performance for those scenarii.

### DIFF
--- a/models/ast/ast_node.go
+++ b/models/ast/ast_node.go
@@ -7,6 +7,8 @@ import (
 )
 
 type Node struct {
+	Index int
+
 	// A node is a constant xOR a function
 	Function Function
 	Constant any
@@ -48,4 +50,21 @@ func (node Node) ReadConstantNamedChildString(name string) (string, error) {
 		return "", errors.New(fmt.Sprintf("\"%s\" constant is not a string: takes value %v", name, child.Constant))
 	}
 	return value, nil
+}
+
+// Cost calculates the weights of an AST subtree to reorder, when the parent is commutative,
+// nodes to prioritize faster ones.
+func (node Node) Cost() int {
+	selfCost := 0
+	childCost := 0
+
+	if attrs, err := node.Function.Attributes(); err == nil {
+		selfCost = attrs.Cost
+	}
+
+	for _, ch := range node.Children {
+		childCost += ch.Cost()
+	}
+
+	return selfCost + childCost
 }
--- a/models/ast/ast_node.go
+++ b/models/ast/ast_node.go
@@ -65,6 +65,9 @@ func (node Node) Cost() int {
 	for _, ch := range node.Children {
 		childCost += ch.Cost()
 	}
+	for _, ch := range node.NamedChildren {
+		childCost += ch.Cost()
+	}
 
 	return selfCost + childCost
 }
--- a/models/ast/ast_node_evaluation.go
+++ b/models/ast/ast_node_evaluation.go
@@ -7,8 +7,9 @@ import (
 )
 
 type NodeEvaluation struct {
-	// Index of the initial node in the AST tree, used to reorder the results as they were.
-	// This should become obsolete when each node has a unique ID.
+	// Index of the initial node winhin its level of the AST tree, used to
+	// reorder the results as they were. This should become obsolete when each
+	// node has a unique ID.
 	Index int
 	// Skipped indicates whether this node was evaluated at all or not. A `true` values means the
 	// engine determined the result of this node would not impact the overall decision's outcome.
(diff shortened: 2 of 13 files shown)


models/ast/ast_node.go
```diff
@@ -49,3 +51,20 @@ func (node Node) ReadConstantNamedChildString(name string) (string, error) {
...
+// nodes to prioritize faster ones.
+func (node Node) Cost() int {
+	selfCost := 0
+	childCost := 0
+
+	if attrs, err := node.Function.Attributes(); err == nil {
+		selfCost = attrs.Cost
+	}
+
+	for _, ch := range node.Children {
```
Pascal-Delange, line 65: Also gotta do the same on name children (for keeping generality at least, not sure if the case can be seen currently)

usecases/ast_eval/evaluate_environment.go
```diff
@@ -28,12 +29,25 @@ func (environment *AstEvaluationEnvironment) GetEvaluator(function ast.Function)
...
+	return environment
+}
+
 func (environment AstEvaluationEnvironment) WithoutCircuitBreaking() AstEvaluationEnvironment {
 	environment.disableCircuitBreaking = true
 
 	return environment
 }
 
+func (environment AstEvaluationEnvironment) WithoutCostOptimizations() AstEvaluationEnvironment {
```
Pascal-Delange, line 45: Just checking my understanding here: this is necessary so we don't have to reorder error results that we return to the frontend, correct?

models/ast/ast_node_evaluation.go
```diff
@@ -7,6 +7,13 @@ import (
 )
 
 type NodeEvaluation struct {
+	// Index of the initial node in the AST tree, used to reorder the results as they were.
```
Pascal-Delange, line 10: just to clarify the comment: this is the index in the initial array of children, correct ? not a tree-global index of nodes

Pascal-Delange (approved):
looks good !

usecases/ast_eval/evaluate_environment.go
apognu: The `Without*` methods are mostly for testing (when you're testing one of the two optimizations, so the other one does not bother the test), and for validation of scenario (so optimizations do not change the given scenario).

models/ast/ast_node_evaluation.go
apognu: Indeed, this is the original index at that level of the tree. I will clarify the comment.

models/ast/ast_node.go
apognu: Aren't named children already order-agnostic? They're both stored in a `map`, which does not have a reproducible order, and I assume are used out-of-order?

Will look into it tomorrow.

usecases/ast_eval/evaluate_environment.go
apognu: Now that I think about it, was your comment on the right line?

If it was about the whole re-ordering / de-re-ordering instead, yes, it is done to ensure the tree results are sent in the order the frontend expects it to be, since it relies on that specific structure to match results with their node.

That would not be necessary if every node had a unique ID, but since that is not the case, we need this.

models/ast/ast_node_evaluation.go
```diff
@@ -7,6 +7,13 @@ import (
...
+	// This should become obsolete when each node has a unique ID.
+	Index int
+	// Skipped indicates whether this node was evaluated at all or not. A `true` values means the
+	// engine determined the result of this node would not impact the overall decision's outcome.
+	Skipped bool
```
Pascal-Delange, line 15: oh, I just remembered.
If the field is to be exploited on the frontend, the same field should be added on the NodeEvaluationDto as well (and set in the adapted function)

apognu: The field is not serialized for the frontend, it is used internally to restore the initial order.

Pascal-Delange: I thought index is, but skipped was to be returned to the frontend to display "this part was not computed..." ?

apognu: You are correct, it's on my local working copy, didn't push it yet, I'll do it in a few minutes.

Pascal-Delange: ok, I didn't view correctly, it's there alright

apognu:
This could benefit, before merging, of adding some new costs for other functions accessing the database.

PR #802, opened 2025-01-16
Added initial migration and dbmodels for sanction checks.

apognu:


apognu:
I will close this and rebase #803 on it, since work is mainly done there now.

PR #803, opened 2025-01-16
Sanction check base

apognu:


repositories/sanction_check_config_repository.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/repositories/dbmodels"
+	"github.com/checkmarble/marble-backend/utils"
+	"github.com/cockroachdb/errors"
+)
+
+func (repo *MarbleDbRepository) GetSanctionCheckConfig(ctx context.Context, exec Executor,
```
Pascal-Delange, line 15: question: since the model of scenario iterations contains a `*SanctionCheckConfig`, and we get the sanction check config by scenario iteration id, are we perhaps better served with a repo method that returns a `*SanctionCheckConfig` (or a []SanctionCheckConfig) rather than a SanctionCheckConfig with need to check on the error type for "not found" ?

repositories/sanction_check_repository.go
```diff
@@ -0,0 +1,192 @@
...
+		Where(squirrel.Eq{"id": match.Id}).
+		Suffix(fmt.Sprintf("RETURNING %s", strings.Join(dbmodels.SelectSanctionCheckMatchesColumn, ",")))
+
+	return SqlToModel(ctx, exec, sql, dbmodels.AdaptSanctionCheckMatch)
+}
+
+func (*MarbleDbRepository) InsertSanctionCheck(ctx context.Context, exec Executor, decisionId string,
+	sanctionCheck models.SanctionCheck,
+) (models.SanctionCheck, error) {
+	utils.LoggerFromContext(ctx).Debug("SANCTION CHECK: inserting matches in database")
```
Pascal-Delange, line 110: prefer debugContext so that we get traces in the log (if I'm not mistaken)

usecases/sanction_check_usecase.go
```diff
@@ -0,0 +1,294 @@
...
+	if decision[0].Case == nil {
+		return models.Decision{}, models.SanctionCheck{}, errors.Wrap(models.NotFoundError,
+			"this sanction check is not linked to a case")
+	}
+	if decision[0].Case.Status.IsFinalized() || sanctionCheck.Status.IsFinalized() {
+		return models.Decision{}, models.SanctionCheck{},
+			errors.Wrap(models.NotFoundError, "this sanction is not pending review")
+	}
+
+	inboxes, err := uc.inboxRepository.ListInboxes(ctx, uc.executorFactory.NewExecutor(), decision[0].OrganizationId, nil, false)
```
Pascal-Delange, line 238: have a look at `(i *InboxReader) ListInboxes(...)`, maybe we can reuse that here (the permission logic for "do I have access to this inbox" should already been handled in it)

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -217,6 +247,16 @@ func EvalTestRunScenario(ctx context.Context,
 		return models.ScenarioExecution{}, err
 	}
 
+	scc, err := repositories.EvalSanctionCheckConfigRepository.GetSanctionCheckConfig(ctx, exec, testRunIteration.Id)
```
Pascal-Delange, line 250: note for later: we said that we perhaps don't want to call the opensanctions api again in the context of a test run (to check with arnaud)

Pascal-Delange (commented):
thanks ! 
It's a bit massive so I didn't read every line in depth, I don't think I have major comments so far (a few minor comments in the PR body)
We can continue with this as a feature branch or merge it and take it up from there, I don't have a strong opinion on either solution really

repositories/sanction_check_config_repository.go
```diff
@@ -0,0 +1,92 @@
...
+			utils.Or(cfg.Enabled, true),
+			cfg.Outcome.ForceOutcome.MaybeString(),
+			utils.Or(cfg.Outcome.ScoreModifier, 0),
+			utils.Or(triggerRule, []byte(``)),
+		)
+
+	updateFields := make([]string, 0, 4)
+
+	if cfg.Enabled != nil {
+		updateFields = append(updateFields, "enabled = EXCLUDED.enabled")
```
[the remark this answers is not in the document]
apognu: Yes, it is the way to make an upsert in Postgres.

There is another tip to simplify the logic when we do not want to update on `NULL`s, at a slight cost in syntax legibility:

```sql
insert into sanction_check_configs as t (scenario_iteration_id, enabled, trigger_rule, forced_outcome, score_modifier, datasets)
values (
    'b0d7c8b9-16bc-42bf-80a0-1e2d609c0dc0',
    true,
    '{}',
    NULL,
    NULL,
    NULL
)
on conflict (scenario_iteration_id) do update
set
    forced_outcome = coalesce(excluded.forced_outcome, t.forced_outcome),
    score_modifier = coalesce(excluded.score_modifier, t.score_modifier),
    datasets = coalesce(excluded.datasets, t.datasets);

```

repositories/sanction_check_repository.go
apognu: Oh yeah, those were just for development, I forgot to remove them indeed.

repositories/migrations/20250120101100_sanction_check.sql
```diff
@@ -0,0 +1,84 @@
...
+);
+
+create index idx_sanction_check_match_comments_sanction_check_match_id on sanction_check_match_comments (sanction_check_match_id);
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+drop table sanction_checks;
```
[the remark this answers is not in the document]
apognu: Are you certain? Dropping a table drops indices directly associated with that table according to the doc.

> DROP TABLE always removes any indexes, rules, triggers, and constraints that exist for the target table.

`CASCADE` would be needed to delete another table or view depending on this one.

usecases/sanction_check_usecase.go
apognu: Does that mean we can remove the check below this line (`ReadOrUpdateCase`) because the call to `ListInboxes` will already check if the user has access to the inbox, and that is the only permission we are interested in here (if someone has access to an inbox, does that mean they are allowed to update the cases in that inbox?)?

Pascal-Delange: Actually sorry, my answer was a bit off. Using it will not allow to remove any lines of code that you already wrote - however it will avoid us having to add some more to take into account inbox permissions.

Pascal-Delange: => https://github.com/checkmarble/marble-backend/pull/810

repositories/sanction_check_config_repository.go
Pascal-Delange: => https://github.com/checkmarble/marble-backend/pull/817

usecases/evaluate_scenario/evaluate_scenario.go
Pascal-Delange: => done in https://github.com/checkmarble/marble-backend/pull/826

Pascal-Delange (approved):

PR #804, opened 2025-01-17
Adding ability to filter case by name

carere:


repositories/case_repository.go
```diff
@@ -244,6 +244,9 @@ func applyCaseFilters(query squirrel.SelectBuilder, filters models.CaseFilters)
 	if len(filters.InboxIds) > 0 {
 		query = query.Where(squirrel.Eq{"c.inbox_id": filters.InboxIds})
 	}
+	if filters.Name != "" {
+		query = query.Where(squirrel.ILike{"c.name": filters.Name})
```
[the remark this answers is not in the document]
Pascal-Delange: In fact, the main index that exists currently on cases is one on `(org_id, created_at desc)`. 
The table should also be large-ish but not too much, since the hypothesis is that cases are created to be human-reviewed (so unless errors are made on the scenario setup, that should mean cases will remain in the 10ks per org rather than millions).
So, I think we could create a new index (replacing the existing one) on `(org_id, created_at DESC) STORING(inbox_id, name, status)`.
And indeed, the ILIKE condition should take `%filters.Name%` rather than just `filters.Name` as value.

repositories/migrations/20250117095221_fuzzy_match_on_case_name.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE EXTENSION pg_trgm;
+ALTER DATABASE marble SET pg_trgm.similarity_threshold = 0.1;
+CREATE INDEX trgm_cases_on_name ON cases USING GIN (name gin_trgm_ops);
```
Pascal-Delange, line 5: As discussed, let's also create this `CREATE INDEX CONCURRENTLY case_org_id_idx_2 ON cases(org_id, created_at DESC) INCLUDE(inbox_id, status, name)`
and once it succeeds, drop the old case_org_id_idx.

In order to create it concurrently you'll need to run the goose migration outside of a transaction, by removing the `-- +goose StatementBegin/End` and prefixing `-- +goose NO TRANSACTION` in the first line

Pascal-Delange, line 5:
```suggestion
CREATE INDEX CONCURRENTLY trgm_cases_on_name ON cases USING GIN (name gin_trgm_ops);
```

carere: Ok donc cette migration se fera en dehors d'une migratrion, je fais la modif

Pascal-Delange: outside of a transaction 👌  (because you can't create index concurrently in a transaction)

Pascal-Delange (approved):
let's merge !

PR #806, opened 2025-01-27
Fix issue where an admin could not change a user's role except to ADMIN.

apognu:
There was an issue in the security enforcement for update user where an admin could not change a user's role except **to** `ADMIN`.


Pascal-Delange (approved):

apognu:
Will merge tomorrow morning.

PR #807, opened 2025-01-28
Payload node cost should be 0

apognu:
I misunderstood the payload node. Should its cost be set to 0?


Pascal-Delange (approved):
that's correct, sorry, it seems I wasn't very careful when I read this.
In fact, I think we may as well set the cost for custom list reading & db field reading to 0 (or a very low value anyway) because they should be really low compared to aggregates.

PR #808, opened 2025-01-29
Implement query AST, with a new function, StringConcat.

apognu:


models/scenario_iterations.go
```diff
@@ -70,9 +71,14 @@ type UpdateSanctionCheckConfigInput struct {
 	Enabled     *bool
 	Datasets    []string
 	TriggerRule *ast.Node
+	Query       *SanctionCheckConfigQuery
 	Outcome     UpdateSanctionCheckOutcomeInput
 }
 
+type SanctionCheckConfigQuery struct {
+	Name ast.Node
```
Pascal-Delange, line 79: if I understand correctly this is the "name" on which we filter on yente, would it make sense for future compatibility to also store the entity (here, default to "thing") on which we filter ?

Pascal-Delange (approved):

models/scenario_iterations.go
Pascal-Delange: ok, got it. So it's "name" as in the design where we filter on the "name" (and optionally other things)

apognu: Yes. It translates, on execution, to an OpenSanctions query which is stored in the `sanction_checks` table.

PR #811, opened 2025-01-30
Cache node evaluation to reuse results within a run

apognu:
The AST language is idempotent within a single run. We can assume that it is expected a node configured the same way returns the same result within a single run of a scenario.

This PR makes it so:
 * Two identical nodes executing at the same time would only be executed once (the result being shared with the second, waiting execution).
 * A node executing after an identical node finished executing will reuse its result without executing.


models/ast/ast_node_evaluation.go
```diff
@@ -63,3 +72,106 @@ func (root NodeEvaluation) GetStringReturnValue() (string, error) {
...
+	Took         time.Duration
+	Nodes        int
+	SkippedCount int
+	CachedCount  int
+	Skipped      bool
+	Cached       bool
+	Children     []EvaluationStats
+}
+
+func BuildEvaluationStats(root NodeEvaluation, parentCached bool) EvaluationStats {
```
Pascal-Delange, line 101: Seeing how the main source of execution delay by is linked to computing aggregates today, I think this may be at the same time a bit overkill for generic nodes and lack of detailed breakdown for aggregates nodes (and perhaps tomorrow, for sanction check/api call nodes)

Pascal-Delange: (we can keep it though, it's fine)

Pascal-Delange (approved):
nice work !

models/ast/ast_node_evaluation.go
apognu: We could filter some low-cost nodes, if necessaery. The issue being that costly functions could be nested under those "simple" nodes, which is why I walk the whole tree to gather statistics.

PR #815, opened 2025-01-31
Experiment with name recognition.

apognu:


usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -74,3 +77,81 @@ func (e ScenarioEvaluator) evaluateSanctionCheck(
...
+
+	matches, err := e.nameRecognizer.PerformNameRecognition(ctx, labelFilter)
+	if err != nil {
+		return queries, errors.New("could not perform name recognition on label")
+	}
+
+	var personQuery *models.OpenSanctionsCheckQuery = nil
+	var companyQuery *models.OpenSanctionsCheckQuery = nil
+
+	for _, match := range matches {
```
apognu, line 124: Here, I create a subquery for each type of match we get from name recognition, so if the name recognizer gives us two `Person`s and one `Company`, we will end up with a request containing the following subqueries:

* `Thing` with the initial query
* `Person` containing two values
* `Company` with one value

Is this what we want?

apognu (commented):
This assumes the name recognition API is published and configured. Interface to this API still needs to be formalized.

usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -41,27 +41,30 @@ func (e ScenarioEvaluator) evaluateSanctionCheck(
...
+	queries := []models.OpenSanctionsCheckQuery{mainQuery}
+
+	if e.nameRecognizer != nil && iteration.SanctionCheckConfig.Query.Label != nil {
+		queries, err = e.evaluateSanctionCheckLabel(ctx, queries, iteration, dataAccessor)
+		if err != nil {
+			return nil, true, err
+		}
+	}
+
 	// Then, actually perform the sanction check
```
Pascal-Delange, line 60: as a very minor comment, the control flow is a little bit hard to follow here (well at least as a git diff it is), if there is no particular reason to defer this to after the call to evaluateSanctionCheckLabel maybe it could be done right after creating mainQuery (if there is a specific reason and I just missed it, just disregard this comment)

usecases/evaluate_scenario/evaluate_sanction_check.go
Pascal-Delange: this seems correct to me, the one thing I'm not sure of is, do we want to really specify the type of entity that we filter by or "keep it safe" and search by "thing" => to check with Arnaud

Pascal-Delange (approved):
👍

Pascal-Delange (approved):

PR #817, opened 2025-02-02
Pascal/mar 725 enforce license keys feature access permissions to setup the

Pascal-Delange:
### small refactoring in the first commit
- GetSanctionCheckConfig by scenario iteration id returns a pointer to a sanction check config, because there may not be a sanction check on the iteration
- remove some function parameters that were effectively passed in a duplicate way
- de-indent some code in usecases/evaluate_scenario/evaluate_sanction_check.go (TBH this one is a matter of taste, I like to try and keep code not over indented to keep the flow more readable, but I'm happy to discuss it and revert it)

### Actual aim of the PR
- create a feature access class that we can reuse everywhere we need to check feature accesses before doing things in the backend

### DIFF
--- a/models/scenario_iterations.go
+++ b/models/scenario_iterations.go
@@ -55,6 +55,7 @@ type UpdateScenarioIterationBody struct {
 }
 
 type SanctionCheckConfig struct {
+	// TODO: Do we really need the "enabled" field ?
 	Enabled     bool
 	Datasets    []string
 	TriggerRule ast.Node
--- a/models/scenario_iterations.go
+++ b/models/scenario_iterations.go
@@ -55,8 +55,6 @@ type UpdateScenarioIterationBody struct {
 }
 
 type SanctionCheckConfig struct {
-	// TODO: Do we really need the "enabled" field ?
-	Enabled     bool
 	Datasets    []string
 	TriggerRule ast.Node
 	Query       SanctionCheckConfigQuery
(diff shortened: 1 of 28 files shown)


models/scenario_iterations.go
```diff
@@ -55,6 +55,7 @@ type UpdateScenarioIterationBody struct {
 }
 
 type SanctionCheckConfig struct {
+	// TODO: Do we really need the "enabled" field ?
```
Pascal-Delange, line 58: open question here. Was your intention to have a sanction_check_config row for every iteration (with backfill) and control if they're active with this ?

usecases/evaluate_scenario/evaluate_scenario.go
```diff
@@ -46,7 +45,16 @@ type SnoozesForDecisionReader interface {
...
 
+type ScenarioEvaluatorFeatureAccessReader interface {
+	GetOrganizationFeatureAccess(
+		ctx context.Context,
+		organizationId string,
+	) (models.OrganizationFeatureAccess, error)
+}
+
 type ScenarioEvaluationRepositories struct {
+	// TODO: I'll do it in a future PR to make review easier, but dependency injection has gotten out of hand
```
Pascal-Delange, line 56: next up, in a separate PR to keep it readable

apognu (approved):

models/scenario_iterations.go
Pascal-Delange: (I'll do it in this PR or in the one that follows then)

usecases/evaluate_scenario/evaluate_scenario.go
Pascal-Delange: https://github.com/checkmarble/marble-backend/pull/820

PR #818, opened 2025-02-03
feat: add endpoint to get ingested object data

ChibiBlasphem:
Add endpoint `/data/:tableName/:objectID` which returns ingested object data

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -51,6 +51,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth Aut
 	router.POST("/ingestion/:object_type/multiple", tom, handleIngestionMultiple(uc))
 	router.POST("/ingestion/:object_type/batch", timeoutMiddleware(conf.BatchTimeout), handlePostCsvIngestion(uc))
 	router.GET("/ingestion/:object_type/upload-logs", tom, handleListUploadLogs(uc))
+	router.GET("/ingestion/:object_type/:object_id", tom, handleGetIngestedObject(uc))
 
 	router.GET("/scenarios", tom, listScenarios(uc))
 	router.POST("/scenarios", tom, createScenario(uc))
@@ -175,7 +176,6 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth Aut
 	router.GET("/data-model/openapi", tom, handleGetOpenAPI(uc))
 	router.POST("/data-model/pivots", tom, handleCreateDataModelPivot(uc))
 	router.GET("/data-model/pivots", tom, handleListDataModelPivots(uc))
-	router.GET("/data-model/data/:tableName/:objectID", tom, handleGetDataModelObject(uc))
 
 	router.POST("/transfers", tom, handleCreateTransfer(uc))
 	router.GET("/transfers", tom, handleQueryTransfers(uc))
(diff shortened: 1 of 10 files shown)


api/routes.go
```diff
@@ -175,6 +175,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth Aut
 	router.GET("/data-model/openapi", tom, handleGetOpenAPI(uc))
 	router.POST("/data-model/pivots", tom, handleCreateDataModelPivot(uc))
 	router.GET("/data-model/pivots", tom, handleListDataModelPivots(uc))
+	router.GET("/data-model/data/:tableName/:objectID", tom, handleGetDataModelObject(uc))
```
Pascal-Delange, line 178: purely for good ordering of the file, I think this would be best placed with the "ingestion" routes (similar role, similar format)

repositories/ingested_data_read_repository.go
```diff
@@ -25,6 +25,7 @@ type IngestedDataReadRepository interface {
 		exec Executor,
 		table models.Table,
 		objectId string,
+		columnsToAdd []string,
```
Pascal-Delange, line 28: small note: what you can easily do to do the same effect without changing the function signature for all callers, is to use `columnsToAdd ...string,`, using a variadic (repeated & optional) parameter
That being said, see the other comment.

Pascal-Delange: (it still changes the signature of the interface, but not necessarily of calling code)

usecases/ingested_data_reader_usecase.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/checkmarble/marble-backend/usecases/executor_factory"
+)
+
+type IngestedDataReaderUsecase struct {
+	ingestedDataReadRepository repositories.IngestedDataReadRepository
+	dataModelRepository        repositories.DataModelRepository
+	executorFactory            executor_factory.ExecutorFactory
+}
+
+func (usecase *IngestedDataReaderUsecase) GetIngestedObject(ctx context.Context,
```
Pascal-Delange, line 16: Functionally, this does the job. 
However, I'd like to propose you to push the modelisation a bit further by doing (something like) the following:
- return a model containing data (the actual object data: fields from the data model) and metadata (valid from, valid until, id)
-  modify the signature & implementation of QueryIngestedObject so that it does the same
- don't need to pass "columnsToAdd"
- return a similar format on the API DTO

Motivation for doing this is twofold:
- it should make it easier on the frontend side to consume the metadata and the actual data separately
- we avoid passing a free content "columnsToAdd" parameter that tightly couples the implementation of the repository's interface to the usecase that consumes it

repositories/ingested_data_read_repository.go
ChibiBlasphem: Yeah I saw that version when I wanted to add the parameters but if we want to change the signature again it will cost more.

repositories/ingested_data_read_repository.go
```diff
@@ -259,7 +260,20 @@ func (repo *IngestedDataReadRepositoryImpl) QueryIngestedObject(
 		return nil, err
 	}
 
-	return objectsAsMap, nil
+	var ingestedObjects []models.DataModelObject
```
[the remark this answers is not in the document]
Pascal-Delange: Also, you still need to add the columns you want to read in addition (valid_from at least I guess) to the list of `columnNames`

Pascal-Delange (approved):
almost good, one minimal thing to change. Thanks !

Pascal-Delange (approved):
🙌  great

PR #821, opened 2025-02-04
Pascal/mar 734 return information on missing configuration in feature

Pascal-Delange:
- feature access has a new option "missing_configuration" for a feature that is accessible but that has missing dependencies (computed from checking if the minimal required environment variables are set, not by checking that the variables are valid end to end)
- return "test" instead of "restricted" for all features if test mode is enabled, AKA if the firebase emulator host is set

Related Frontend PR https://github.com/checkmarble/marble-frontend/pull/715

### DIFF
--- a/usecases/usecases.go
+++ b/usecases/usecases.go
@@ -24,6 +24,9 @@ type Usecases struct {
 	caseManagerBucketUrl        string
 	failedWebhooksRetryPageSize int
 	hasConvoyServerSetup        bool
+	hasMetabaseSetup            bool
+	hasOpensanctionsSetup       bool
+	hasTestMode                 bool
 	license                     models.LicenseValidation
 }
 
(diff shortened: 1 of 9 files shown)


usecases/usecases.go
```diff
@@ -86,6 +114,10 @@ func newUsecasesWithOptions(repositories repositories.Repositories, o *options)
 		caseManagerBucketUrl:        o.caseManagerBucketUrl,
 		failedWebhooksRetryPageSize: o.failedWebhooksRetryPageSize,
 		license:                     o.license,
+		hasConvoyServerSetup:        o.hasConvoyServerSetup,
```
Pascal-Delange, line 117: this one was missing (even though the option was already passed) so the fix where calling convoy was bypassed if not configured, probably never worked

carere (approved):

PR #822, opened 2025-02-04
Copy files when sanction check is refined

apognu:


repositories/sanction_check_repository.go
```diff
@@ -320,3 +320,24 @@ func (repo *MarbleDbRepository) GetSanctionCheckFile(ctx context.Context, exec E
...
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return err
+	}
+
+	sql := NewQueryBuilder().
+		Insert(dbmodels.TABLE_SANCTION_CHECK_FILES).
+		Columns("bucket_name", "file_reference", "file_name", "sanction_check_id").
+		Select(squirrel.
+			Select("bucket_name", "file_reference", "file_name").
+			Column("?", newSanctionCheckId).
```
Pascal-Delange, line 334:
```suggestion
			Column("sanction_check_id", newSanctionCheckId).
```
Should be this ? or I didn't understand

Pascal-Delange (approved):

repositories/sanction_check_repository.go
apognu: We are inserting a static value (the ID of the new sanction check into that column), so I believe what I initially did was correct. It amounts to:

```sql
INSERT INTO files (bucket_name, file_reference, file_name, sanction_check_id)
SELECT bucket_name, file_reference, file_name, '<actual uuid of new sanction check>'
FROM files
WHERE id = '<uuid of old sanction check>';
```

Pascal-Delange: aah yes of course, I didn't get it at the first read. Makes sense now

PR #832, opened 2025-02-06
Fixing cached evaluation

apognu:
Reverts checkmarble/marble-backend#830

### DIFF
--- a/usecases/ast_eval/evaluate_ast.go
+++ b/usecases/ast_eval/evaluate_ast.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/checkmarble/marble-backend/models/ast"
 	"github.com/checkmarble/marble-backend/pure_utils"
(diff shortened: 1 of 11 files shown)


Pascal-Delange (approved):
I haven't checked how mohae/deepcopy works, but in principle seems ok to me 👍 
what we can do is merge this tomorrow morning and do a new release of the backend, but ramp up the traffic on it progressively (so that we don't release it at the same time as sanctions, and also don't sit on it for 2 weeks)

usecases/ast_eval/evaluate_ast.go
```diff
@@ -38,53 +80,83 @@ func EvaluateAst(ctx context.Context, environment AstEvaluationEnvironment, node
...
+
+		if !ok {
+			// The evaluator is supposed to return nil ReturnValue when an error is present.
+			evaluation.ReturnValue = nil
+		}
+
+		evaluationResponse := nodeEvaluationResponse{evaluation, ok}
+
+		if cache != nil {
+			cache.Cache.Store(hash, deepcopy.Copy(evaluationResponse))
```
apognu, line 143: Arguably, this copy should not be required.

We copy in two other instances:

* When retrieving this from the cache (since it may be shared by others using the cache)
* Below if other nodes were waiting for our result.

Pascal-Delange (approved):

PR #834, opened 2025-02-07
Allow adding a comment when reviewing a match.

apognu:
For V1, comments can should only be added when reviewing a match, and there can only be one per match.

### DIFF
--- a/dto/sanction_check_dto.go
+++ b/dto/sanction_check_dto.go
@@ -93,7 +93,8 @@ func AdaptSanctionCheckMatchDto(m models.SanctionCheckMatch) SanctionCheckMatchD
 }
 
 type SanctionCheckMatchUpdateDto struct {
-	Status string `json:"status"`
+	Status  string  `json:"status"`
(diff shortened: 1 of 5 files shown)


dto/sanction_check_dto.go
```diff
@@ -93,7 +93,8 @@ func AdaptSanctionCheckMatchDto(m models.SanctionCheckMatch) SanctionCheckMatchD
 }
 
 type SanctionCheckMatchUpdateDto struct {
-	Status string `json:"status"`
+	Status  string  `json:"status"`
+	Comment *string `json:"comment,omitempty"`
```
apognu, line 97: Is adding a comment on reviews mandatory?

Pascal-Delange: I'd say technically yes, though there is no point in overdoing the validation since nothing stops the customer from putting a short non-empty string that doesn't say anything

Pascal-Delange (approved):

PR #837, opened 2025-02-07
Pascal/fix status update

Pascal-Delange:
This should fix the cascade of status updates

----

### thoughts on testing with pgxmock in usecases
Debrief "à chaud" on using testing to the sql layer from a usecase: I spent an unreasonable amount of time in the end of the afternoon and in the early evening getting the TestUpdateMatchStatus test to pass.
Issues I encountered:
- can't expect successive calls to the same sql query (even with different params) in a row with different outputs => I couldn't test the exact usecase (update to confirmed hit and mark others as skipped) I initially wanted to
- gets really cumbersome to work with if you want to test an execution path where several calls to the repository are chained and the output of one (with specific ids, status...) impacts the following ones.

All in all, I love the idea of using pgxmock at the usecase unit test level for simple usecase calls where there are just one or two calls to the "fully tested" usecase repo (with little logic coupling between their input/output), but in this case it got frustrating quickly. It's a fine line to tread. 

(oh and then I rebased the feature branch and spent another 15 min manually finding the diff in the queries. Today made me grumpy against go tooling)

=> all in all nothing critical, but gives me some food for thought on how to improve both our testing and testability


usecases/sanction_check_usecase.go
```diff
@@ -328,24 +328,30 @@ func (uc SanctionCheckUsecase) UpdateMatchStatus(
 				for _, m := range pendingMatchesExcludingThis {
 					_, err = uc.repository.UpdateSanctionCheckMatchStatus(ctx, tx, m, models.SanctionCheckMatchUpdate{
 						MatchId:    m.Id,
-						Status:     models.SanctionMatchStatusNoHit,
+						Status:     models.SanctionMatchStatusSkipped,
```
Pascal-Delange, line 331: this was an error

apognu (approved):

apognu:
Regardless of the status of the PR, I will have a quick stab at the testing issue.

Pascal-Delange:
> Regardless of the status of the PR, I will have a quick stab at the testing issue.

thanks for the changes !

PR #839, opened 2025-02-10
Implement sanction check whitelists.

apognu:
A sanction check config can now receive an AST to dynamically select the field uniquely identifying a record.

A sanction check match review can now include a boolean, `whitelist` that will permanently link an Open Sanctions entity ID to the unique identifier selected by the previously defined expression.

Any object/entity that has been whitelisted will not be excluded from a sanction check matches.

### DIFF
--- a/dto/sanction_check_dto.go
+++ b/dto/sanction_check_dto.go
@@ -73,6 +73,7 @@ type SanctionCheckMatchDto struct {
 	Status     string                         `json:"status"`
 	ReviewedBy *string                        `json:"reviewer_id,omitempty"` //nolint:tagliatelle
 	Datasets   []string                       `json:"datasets"`
+	ObjectId   *string                        `json:"object_id"`
 	Payload    json.RawMessage                `json:"payload"`
 	Comments   []SanctionCheckMatchCommentDto `json:"comments"`
 }
@@ -86,6 +87,7 @@ func AdaptSanctionCheckMatchDto(m models.SanctionCheckMatch) SanctionCheckMatchD
 		QueryIds:   m.QueryIds,
 		Datasets:   make([]string, 0),
 		Payload:    m.Payload,
+		ObjectId:   m.ObjectId,
 		Comments:   pure_utils.Map(m.Comments, AdaptSanctionCheckMatchCommentDto),
 	}
 
--- a/dto/sanction_check_dto.go
+++ b/dto/sanction_check_dto.go
@@ -67,28 +67,28 @@ func AdaptSanctionCheckRefineDto(dto SanctionCheckRefineDto) models.SanctionChec
 }
 
 type SanctionCheckMatchDto struct {
-	Id         string                         `json:"id"`
-	EntityId   string                         `json:"entity_id"`
-	QueryIds   []string                       `json:"query_ids"`
-	Status     string                         `json:"status"`
-	ReviewedBy *string                        `json:"reviewer_id,omitempty"` //nolint:tagliatelle
-	Datasets   []string                       `json:"datasets"`
-	ObjectId   *string                        `json:"object_id"`
-	Payload    json.RawMessage                `json:"payload"`
-	Comments   []SanctionCheckMatchCommentDto `json:"comments"`
+	Id                           string                         `json:"id"`
+	EntityId                     string                         `json:"entity_id"`
+	QueryIds                     []string                       `json:"query_ids"`
+	Status                       string                         `json:"status"`
+	ReviewedBy                   *string                        `json:"reviewer_id,omitempty"` //nolint:tagliatelle
+	Datasets                     []string                       `json:"datasets"`
+	UniqueCounterpartyIdentifier *string                        `json:"unique_counterparty_identifier"`
+	Payload                      json.RawMessage                `json:"payload"`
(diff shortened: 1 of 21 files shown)


dto/sanction_check_dto.go
```diff
@@ -86,15 +87,17 @@ func AdaptSanctionCheckMatchDto(m models.SanctionCheckMatch) SanctionCheckMatchD
 		QueryIds:   m.QueryIds,
 		Datasets:   make([]string, 0),
 		Payload:    m.Payload,
+		ObjectId:   m.ObjectId,
```
apognu, line 90: Should we use another term for this, to not be mixed up with our other "object IDs"?

repositories/migrations/20250210144400_create_sanctions_whitelist.sql
```diff
@@ -0,0 +1,45 @@
+-- +goose Up
+-- +goose StatementBegin
+
+alter table sanction_check_configs
+    add column whitelist_field_ast
```
apognu, line 5: We may want to harmonize this depending on how we end up calling those object IDs (see previous comment).

repositories/dbmodels/db_sanction_check.go
```diff
@@ -16,20 +16,21 @@ var (
...
+	SearchInput         json.RawMessage `db:"search_input"`
+	SearchDatasets      []string        `db:"search_datasets"`
+	MatchThreshold      int             `db:"match_threshold"`
+	MatchLimit          int             `db:"match_limit"`
+	IsManual            bool            `db:"is_manual"`
+	RequestedBy         *string         `db:"requested_by"`
+	IsPartial           bool            `db:"is_partial"`
+	IsArchived          bool            `db:"is_archived"`
+	InitialHasMatches   bool            `db:"initial_has_matches"`
+	WhitelistedEntities *[]string       `db:"whitelisted_entities"`
```
Pascal-Delange, line 31: any particular reason for a pointer to a slice ? 👀

usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -70,6 +70,37 @@ func (e ScenarioEvaluator) evaluateSanctionCheck(
...
+			params.Scenario.OrganizationId,
+			dataAccessor.ClientObject,
+			params.DataModel,
+		)
+		if err != nil {
+			sanctionCheckErr = errors.Wrap(err, "could not extract object field for whitelist check")
+			return
+		}
+
+		whitelistField, err := whitelistFieldResult.GetStringReturnValue()
```
Pascal-Delange, line 87: Here, IMO we should adapt logic if a null value is read (check on ErrNullFieldRead returned, or just rewrite the downcast logic here) as well as if an empty string is read.
(well: empty string perhaps takes care of itself in FilterOut, but at least avoid errors incase of null)

dto/sanction_check_dto.go
Pascal-Delange: Yes let's do that.
How about "UniqueCounterpartyIdentifier" or such ?

usecases/sanction_check_usecase.go
```diff
@@ -384,6 +428,15 @@ func (uc SanctionCheckUsecase) UpdateMatchStatus(
 					return err
 				}
 			}
+
+			if update.Status == models.SanctionMatchStatusNoHit && update.Whitelist && data.match.ObjectId != nil {
```
Pascal-Delange, line 432: and `!= ""`?

Pascal-Delange (commented):
We discussed something about querying "match_limit + nb_whitelisted" for the object when first calling OS/yente, I still think that made sense ?

usecases/sanction_check_usecase.go
apognu: I would rather we not store an empty string to begin with (see one of your previous comments) rather than storing empty strings and conditionally applying logic from that.

apognu:
> We discussed something about querying "match_limit + nb_whitelisted" for the object when first calling OS/yente, I still think that made sense ?

You are correct, that is an oversight on my part. Will amend.

usecases/evaluate_scenario/evaluate_sanction_check.go
apognu: I guess we only want to throw an error for unexpected `errs` and continue processing (but skipping registering the whitelist) on `ErrNullFieldRead` and `whitelistField == ""`?

repositories/dbmodels/db_sanction_check.go
Pascal-Delange: let's do that 👍

usecases/sanction_check_usecase.go
Pascal-Delange: yes indeed

usecases/evaluate_scenario/evaluate_sanction_check.go
Pascal-Delange: yes agreed with this

Pascal-Delange (approved):
👍

apognu:
@Pascal-Delange New review requested, I added the counterparty ID expression in the sanction check config API endpoint (latest 1 commit).

PR #841, opened 2025-02-11
Fix permissions on list sanction checks

apognu:
The only security enforcement we had on listing sanction checks was to check whether the user had permission to read or update the case. This fell flat when a decision was not yet added to a case.

This conditions the case permission checking on whether the decision is in a case, or otherwise checks that the user can read the decision.


Pascal-Delange (approved):
good catch

PR #842, opened 2025-02-11
Use uppercase fields on refine query search

apognu:
Frontend used uppercased field names for those.


apognu:
WIll also need to double-check which types of entities we allow the user to refine on, with which fields. Apparently, we are doing different things.

Pascal-Delange (approved):

PR #845, opened 2025-02-13
Fixed sanction check config label AST that cannot be set.

apognu:


dto/sanction_check_config_dto.go
```diff
@@ -113,8 +113,8 @@ func AdaptSanctionCheckConfigInputDto(dto SanctionCheckConfig) (models.UpdateSan
 }
 
 type SanctionCheckConfigQuery struct {
-	Name  NodeDto   `json:"name"`
-	Label *ast.Node `json:"label"`
+	Name  NodeDto  `json:"name"`
+	Label *NodeDto `json:"label"`
```
apognu, line 117: We will still need to handle that name is also optional.

dto/sanction_check_config_dto.go
```diff
@@ -141,5 +152,13 @@ func AdaptSanctionCheckConfigQueryDto(dto SanctionCheckConfigQuery) (models.Sanc
 		Name: nameAst,
 	}
 
+	if dto.Label != nil && dto.Label.Name != ast.FUNC_UNDEFINED.DebugString() {
```
Pascal-Delange, line 155:
```suggestion
	if dto.Label != nil && dto.Label.Name != ast.FUNC_UNDEFINED.DebugString() {
```
I think DebugString() is no the correct function to call here - at least it's likely not what would correspond to what the dto sends.
Perhaps we need another function ast.Function.Name() for this purpose

apognu: For `Undefined` it is, but yes, I just saw there is no clear mapping between the DTO string of a function and what this function returns, so it is a bad call. I would prefer to do this properly, so should I create a `Name()` function for all functions that return the AST identifier?

Pascal-Delange: I'd suggest this 👍 
there is a function that (perhaps clumsily) does the opposite thing in dto_ast_node.go I think

Pascal-Delange (approved):

PR #852, opened 2025-02-17
Create standard response for requirement misconfiguration.

apognu:
To be discussed, it might be interesting to settle on a common status code / error payload to inform the frontend that some preconditions failed on the backend, especially as far as third-party services are concerned.

Those errors could indicate that a specific class of error happened when trying to use a certain third-party service.

In this PR, the error looks like this:

```json
{
    "message": "A required configuration was missing or invalid",
    "details": {
        "error": "missing API key for SaaS Open Sanctions configuration",
        "reason": "missing_configuration",
        "requirement": "open_sanctions"
    },
    "error_code": "missing_requirement"
}
```

Where both `details.requirement` and `details.reason` are enum variants:

```go
type (
	RequirementError       string
	RequirementErrorReason string
)

const (
	REQUIREMENT_OPEN_SANCTIONS RequirementError = "open_sanctions"

	REQUIREMENT_REASON_MISSING_CONFIGURATION = "missing_configuration"
	REQUIREMENT_REASON_INVALID_CONFIGURATION = "invalid_configuration"
	REQUIREMENT_REASON_HEALTHCHECK_FAILED    = "healthcheck_failed"
)
```

Those will be extended to cover more third-parties are maybe more reason for the requirement failure, but they should be standardized so the frontend can generate legible error messages indicating the source of the failure.

All of this is to be discussed.

### DIFF
--- a/api/present_error.go
+++ b/api/present_error.go
@@ -42,6 +42,24 @@ func presentError(ctx context.Context, c *gin.Context, err error) bool {
 	case errors.Is(err, models.ConflictError):
 		logger.InfoContext(ctx, fmt.Sprintf("ConflictError: %v", err))
 		c.JSON(http.StatusConflict, errorResponse)
+	case errors.Is(err, models.MissingRequirementError{}):
+		var req models.MissingRequirementError
+
+		if errors.As(err, &req) {
+			logger.InfoContext(ctx, fmt.Sprintf("MissingRequirementError: %v", err))
+
+			errorResponse = dto.APIErrorResponse{
+				ErrorCode: dto.MissingRequirement,
+				Message:   "A required configuration was missing or invalid",
+				Details: dto.RequirementErrorDto{
+					Requirement: string(req.Requirement),
+					Reason:      string(req.Reason),
+					Error:       req.Err.Error(),
+				},
+			}
+
+			c.JSON(http.StatusInternalServerError, errorResponse)
+		}
 
 	case errors.Is(err, context.DeadlineExceeded):
 		logger.WarnContext(ctx, fmt.Sprintf("Deadline exceeded: %v", err))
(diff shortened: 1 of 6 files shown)


apognu:
Poke @Pascal-Delange for discussion.

api/present_error.go
```diff
@@ -42,6 +42,24 @@ func presentError(ctx context.Context, c *gin.Context, err error) bool {
...
+				ErrorCode: dto.MissingRequirement,
+				Message:   "A required configuration was missing or invalid",
+				Details: dto.RequirementErrorDto{
+					Requirement: string(req.Requirement),
+					Reason:      string(req.Reason),
+					Error:       req.Err.Error(),
+				},
+			}
+
+			c.JSON(http.StatusInternalServerError, errorResponse)
```
Pascal-Delange, line 61: from an API design point of view, should this perhaps be a 422 (unprocessable content) ?

Pascal-Delange (approved):
very much agreed on the design 👍   (with a little open question on status codes)

api/present_error.go
apognu: I thought about it, but could not come up with a status code that satisfied me.

422 is still supposed to be a client-side error, where it asks for something semantically incorrect, instead of syntactically (400). The difference is ambiguous because the codes come from different ecosystems. If I squint hard enough, I could say "the client asked for a feature that the server cannot support", but it's a stretch (as would be 406).

Semantically speaking, I think 501 might be the closest to the meaning we strive for here: "Not Implemented". It is a server error, which we want, and the definition seems to be very close to our meaning without being tied to HTTP protocol details.

Pascal-Delange: I tended to look at it from the angle of "the client is requesting something that he's not supposed to and should know not to call" (in the case of the frontend app in particular, because it received the feature access flags).
But anyway i'm nitpicking and I don't think there is a single truth on this, I'm totally fine with 500 or 501
Especially as long as it's only part of our internal rather than public API

PR #855, opened 2025-02-18
Move index creation to an asynchronous worker

apognu:
This is a prototype for using our workers and asynchronous tasks for various aspects of Marble maintenance or operational tasks. Examples of such tasks might include:

* Creation of database indices
* Cleanup of useless or invalid indices
* Notification sending

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -14,6 +14,8 @@ import (
 	"github.com/checkmarble/marble-backend/repositories"
 	"github.com/checkmarble/marble-backend/usecases"
 	"github.com/checkmarble/marble-backend/utils"
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/riverdriver/riverpgxv5"
 
 	"github.com/cockroachdb/errors"
 	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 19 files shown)


Pascal-Delange (commented):
No remarks on the structure (already discussed in the other PR), but I'm not sure at the first read if this implements all the logic that is needed, so to double check:
IIUC this creates a job in the queue that starts the "async create all indexes" repo method, and then starts a secondary task that monitors the creation to notify about errors.
However I don't think it helps us restart the index creation if for whatever reason it fails in the middle of the creation of one of of several indexes (db error, or more likely a backend container stopped/restarted).
Do you confirm ?

apognu:
Indeed, this is not yet reprensented in the PR.

If we wanted to do that, the error condition [here](https://github.com/checkmarble/marble-backend/pull/855/files#diff-464aed26bf0c7d287f2ccbd342286507666c3dc3deab8448d2fd945d138a9fc7R120) should probably return a success (we are done monitoring), log the error and enqueue a new creation task with the missing indices.

apognu:
It should now retry the creation of failed indices. It added a completely-arbitrary wait time of one minute before the next job starts.

Pascal-Delange (approved):
🚀

PR #856, opened 2025-02-18
Add sanction check run error codes

apognu:
Also fixes a panic introduced with the refactoring of forced outcome.

### DIFF
--- a/usecases/evaluate_scenario/evaluate_sanction_check.go
+++ b/usecases/evaluate_scenario/evaluate_sanction_check.go
@@ -9,6 +9,10 @@ import (
 	"github.com/cockroachdb/errors"
 )
 
+const (
+	ErrSanctionCheckAllFieldsNull = "all_fields_null"
+)
+
 func (e ScenarioEvaluator) evaluateSanctionCheck(
 	ctx context.Context,
 	iteration models.ScenarioIteration,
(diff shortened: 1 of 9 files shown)


usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -44,19 +48,42 @@ func (e ScenarioEvaluator) evaluateSanctionCheck(
...
+				nullFieldRead += 1
+			default:
+				return nil, true, err
+			}
+		}
+	}
+
+	if nullFieldRead >= 2 {
+		sanctionCheck = &models.SanctionCheckWithMatches{
+			SanctionCheck: models.SanctionCheck{
```
Pascal-Delange, line 79: one thing I wondered while reading this was, is there any chance that we're missing some of the other fields on the SanctionCheck model that may be used downstream => turns out no because all of them have reasonable default values when we insert to DB
(must be my OCD but those partial models are always stressing me out, and I don't have a great solution to get rid of them)

usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -157,6 +184,9 @@ func (e ScenarioEvaluator) evaluateSanctionCheckLabel(ctx context.Context, queri
 	if err != nil {
 		return queries, err
 	}
+	if labelFilterAny.ReturnValue == nil {
+		return queries, ast.ErrNullFieldRead
```
Pascal-Delange, line 188: oh oh, looks like I'll have conflicts on the branch I opened yesterday evening at home, trying to clean up usage of `ast.ErrNullFieldRead` as a part of the implicit function signature in different places post-ast evaluation 😅 
(which is fine of course!)

Pascal-Delange (approved):
👍

PR #863, opened 2025-02-19
proposed changes on empty value handling for opensanctions search

Pascal-Delange:
The diff got kind of big, but the premise here is basically:
- opensanctions errors out (400) if it returns a query with only empty search inputs
- so we want to handle a search with empty string in the same way as null value read, on our side
- so I'm also making the concatenation ast node return an empty string if it receives no inputs (or only nil inputs). There could be a point to be made to return nil if it only receives nils ? but I'm not sure if it's worth it

### DIFF
--- a/usecases/ast_eval/evaluate/eval_string_concat.go
+++ b/usecases/ast_eval/evaluate/eval_string_concat.go
@@ -43,9 +43,5 @@ func (f StringConcat) Evaluate(ctx context.Context, arguments ast.Arguments) (an
 		}
 	}
 
-	if sb.Len() == 0 {
-		return MakeEvaluateError(ast.ErrNullFieldRead)
-	}
-
 	return sb.String(), nil
 }
--- a/usecases/evaluate_scenario/evaluate_sanction_check.go
+++ b/usecases/evaluate_scenario/evaluate_sanction_check.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	ErrSanctionCheckAllFieldsNull = "all_fields_null"
+	ErrSanctionCheckAllFieldsNullOrEmpty = "all_fields_null_or_empty"
 )
 
 func (e ScenarioEvaluator) evaluateSanctionCheck(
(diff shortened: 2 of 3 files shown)


usecases/evaluate_scenario/evaluate_sanction_check.go
```diff
@@ -48,37 +48,35 @@ func (e ScenarioEvaluator) evaluateSanctionCheck(
...
-				return nil, true, err
-			}
+			return nil, true, errors.Wrap(err, "could not evaluate sanction check label")
+		} else if emptyInput {
+			emptyFieldRead += 1
 		}
 	}
 
-	if nullFieldRead >= 2 {
+	if emptyFieldRead == nbEvaluatedFields {
```
Pascal-Delange, line 75: this condition was incorrect if one of the two inputs was left empty

usecases/ast_eval/evaluate/eval_string_concat.go
```diff
@@ -43,9 +43,5 @@ func (f StringConcat) Evaluate(ctx context.Context, arguments ast.Arguments) (an
 		}
 	}
 
-	if sb.Len() == 0 {
-		return MakeEvaluateError(ast.ErrNullFieldRead)
```
Pascal-Delange, line 47: we actually used to have a lot of those, they have been rather deprecated in favor of "cleanly" returning nil in Evaluate if a nil value is "read" (AKA some nil received in a required input or actual nil value read in the logic of the function)

apognu:
I was not able to reproduce manually the 400 on empty inputs, do you have the query JSON handy?

Pascal-Delange:
> I was not able to reproduce manually the 400 on empty inputs, do you have the query JSON handy?

Hmm no, I'll check a bit later

Pascal-Delange:
> I was not able to reproduce manually the 400 on empty inputs, do you have the query JSON handy?

@apognu  in short what seems to happens in the main branch is that in some cases at least we are left with an empty slice of queries, for which yente returns a 400 code. 
When only empty strings were received, no proper queries for OS were generated

PR #867, opened 2025-02-20
Ordering of sanction check matches.

apognu:
They are now ordered according to the following criteria:

 - Confirmed hit first, then pending, then no match, then skipped
 - Within those groups, higher scores are returned first


apognu:
A question could be asked about performance, but the filtered set should always be only a few dozen matches at most and fit in memory, so the ordering and JSONB extraction should not have a noticeable impact on performance.

Open to discussion.

Pascal-Delange (approved):
I learned something here, I didn't know the array_position function
logic makes sense to me 👍

PR #868, opened 2025-02-21
Summarize test run rule evaluations in a separate table.

apognu:
This is a prototype of using the worker set up worked on in #855 to periodically save a test run rule execution statistics (for both the live and phantom versions) into a summary table, for faster querying.


repositories/scenario_testrun.go
```diff
@@ -182,3 +185,63 @@ func (repo *MarbleDbRepository) GetTestRunByID(ctx context.Context, exec Executo
...
+				squirrel.Expr("strs.watermark < str.expires_at"),
+				squirrel.Expr("strs.watermark < now()"),
+			},
+		}).
+		GroupBy("str.id")
+
+	return SqlToListOfModels(ctx, exec, sql, dbmodels.AdaptScenarioTestrunWithSummary)
+}
+
+func (repo *MarbleDbRepository) SaveTestRunSummary(ctx context.Context, exec Executor,
```
Pascal-Delange, line 216: so to be clear, the mapping from test run to test run summary should always be [1 to (0,1)], correct ?

models/scenario_testrun.go
```diff
@@ -49,6 +49,12 @@ type ScenarioTestRun struct {
 	CreatedAt               time.Time
 	ExpiresAt               time.Time
 	Status                  TestrunStatus
+	Summarized              bool
```
Pascal-Delange, line 52: This one would benefit from an accompanying comment I think - I head to look up the usecase where it's written to understand that it's useful only for test runs that are done running

usecases/scheduled_execution/test_run_summary_job.go
```diff
@@ -0,0 +1,135 @@
...
+	repository RulesRepository,
+) TestRunSummaryWorker {
+	return TestRunSummaryWorker{
+		executor_factory:    executor_factory,
+		transaction_factory: transaction_factory,
+		repository:          repository,
+	}
+}
+
+func (w *TestRunSummaryWorker) Work(ctx context.Context, job *river.Job[models.TestRunSummaryArgs]) error {
```
Pascal-Delange, line 75: do we also want a define timeout on the job ? I'd say something in the ballpark of 20s to a minute would be reasonable

repositories/scenario_testrun.go
```diff
@@ -182,3 +185,63 @@ func (repo *MarbleDbRepository) GetTestRunByID(ctx context.Context, exec Executo
...
+func (repo *MarbleDbRepository) GetRecentTestRunForOrg(ctx context.Context, exec Executor,
+	orgId string,
+) ([]models.ScenarioTestRunWithSummary, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+
+	sql := NewQueryBuilder().
+		Select(columnsNames("str", dbmodels.SelectScenarioTestRunColumns)...).
+		Column(fmt.Sprintf("ARRAY_AGG(ROW(%s)) filter (where strs.id is not null) AS summaries",
```
Pascal-Delange, line 198: I don't remember what the bug was, but you had this case on sanction check matches where you had to `FILTER (WHERE scm.id IS NOT NULL)` on the left joined part of the query using `ARRAY_AGG(ROW` - what was it for and why is it not needed here ? 🤔

models/scenario_test_run_summary.go
```diff
@@ -0,0 +1,14 @@
+package models
+
+import "time"
+
+type ScenarioTestRunSummary struct {
+	Id           string
+	RuleName     string
+	RuleStableId string
```
Pascal-Delange, line 8: A warning here about stable id if we also do the stats of sanction check rules: there is a bit of a dirty hack in the TestRunStatsByRuleExecution usecase method that computes a fake (uuid) "stable rule id" on the fly for sanction checks in live/test version. This is fine as long as we only compute them in real time, but we need to add a proper stable rule id on the sanction check if we also want to precompute them

usecases/scheduled_execution/test_run_summary_job.go
```diff
@@ -0,0 +1,135 @@
...
+		testRunId string,
+		begin, end time.Time,
+	) ([]models.RuleExecutionStat, error)
+	SaveTestRunSummary(ctx context.Context, exec repositories.Executor,
+		testRunId string, stat models.RuleExecutionStat, newWatermark time.Time,
+	) error
+	SetTestRunAsSummarized(ctx context.Context, exec repositories.Executor, testRunId string) error
+}
+
+func NewTestRunSummaryPeriodicJob(orgId string) *river.PeriodicJob {
```
Pascal-Delange, line 38: Beware that this is a first actual coupling between queue names and business logic - namely, currently queues are created with org ids as names, this function may start behaving strangely if for any reason we start creating queues with another pattern. Definitely not a blocking point, but it's worth documenting (perhaps in the cmd/worker file or so) that some functions rely on the fact that queue names = org ids.

Pascal-Delange (commented):
All clear and cool.
The only thing I have to say is, if we're doing this, we might as well complete it and also do the pre-accounting of:
- sanction check rule execution stats (goes inside the same stats table as the other rules, see a comment inline about the stable rule id)
- decisions by status stats (this one is less painful that the decisions/rules stats, but it's still liable to get significantly slow in the face of high decision volumes)
WDYT?

usecases/scheduled_execution/test_run_summary_job.go
apognu: What would be the next success condition when we hit the timeout?

Except for transient errors, it is fair to say that a run hitting the timeout will never succeed since the next run would have even more data to process, and so on and so forth.

That raises a question this PR did not address yet which is that of monitoring. How do we raise alerts if a test run summary lags too far behind?

repositories/scenario_testrun.go
apognu: The filter is present here, though. 🤓

The issue stems from how the query is resolved when there are no rows associated. The `LEFT JOIN` sets all the RHS columns to `NULL`, making the `array_agg` produce a single all-`NULL` row (e.g. `{,,,,,,,}`, which fails to be deserialized into a struct, since values are not nullable, hence filtering on a column that cannot be null to remove that "phantom" row.

usecases/scheduled_execution/test_run_summary_job.go
apognu: As a less error-prone way for now, we could include the periodic-per-org creation in the `QueuesFromOrgs` methods and append those to the global periodics bundle.

That way, it is clear those specific tasks are per-org, while letting us manage queues and periodics separately.

repositories/scenario_testrun.go
Pascal-Delange: got it, sorry it was late and I missed the "several versions/rules" 🤦

repositories/scenario_testrun.go
Pascal-Delange: ok, j'avais vraiment pas les yeux en face des trous 😅
And got it, it's clearer this morning

usecases/scheduled_execution/test_run_summary_job.go
apognu: I changed the way those are fetched according to my previous comment, is that better?

usecases/scheduled_execution/test_run_summary_job.go
Pascal-Delange: Good question.
By default what would happen now is, if the job consistently fails, we would get notified on sentry (at least I think so - it's worth double checking if a timeout specifically triggers a sentry alarm in the job logger middleware).
Your remark about the "doom loop" where if it fails, it is expected to fail consistently is valid, and a good reason to have a long-ish value for the timeout, at least something longer than a duration that seems realistic to wait for it (a few minutes may do the job)

apognu:
> All clear and cool. The only thing I have to say is, if we're doing this, we might as well complete it and also do the pre-accounting of:
> 
>     * sanction check rule execution stats (goes inside the same stats table as the other rules, see a comment inline about the stable rule id)
> 
>     * decisions by status stats (this one is less painful that the decisions/rules stats, but it's still liable to get significantly slow in the face of high decision volumes)
>       WDYT?

Totally agree, let's talk about it today, I'll start looking at how it works.

usecases/scheduled_execution/test_run_summary_job.go
Pascal-Delange: That being said, I still think it's good to explicitly set the timeout, as not doing so will only implicitly inherit the default job timeout that is (not, in this case) set at the client level.

usecases/scheduled_execution/test_run_summary_job.go
Pascal-Delange: yep I think that's a good workaround 👍

usecases/scheduled_execution/test_run_summary_job.go
apognu: I set a 2-minute timeout for now. All timings will have to be defined, though (intervals and timeouts).

apognu:
I implemented the decision stats in the latest commit. I will have a look at sanctions stats later today.

Pascal-Delange (approved):
Looks good, let's just finish with the stats for sanction checks 👍

PR #870, opened 2025-02-24
Use correct end date for test run depending on org set timezone.

apognu:
Creating a test run had two issues related to the end date selection:

* The end date would always be at the start of the selected day, where instinctively, you would think that selecting an end date of "Feb 27th" would still allow the test run to continue on the 27th.
* The frontend sends us a time relative to the browser's locale, which may or may not be the same as the organization's set timezone. The "end of day" calculation must therefore be performed in the org's timezone, but from the selected date (the user's intent).

This PR sets the end date and time for a test run as the end of the selected day in the organization's timezone.

### DIFF
--- a/api/handle_scenario_test_run.go
+++ b/api/handle_scenario_test_run.go
@@ -23,8 +23,15 @@ func handleCreateScenarioTestRun(uc usecases.Usecases) func(c *gin.Context) {
 			c.Status(http.StatusBadRequest)
 			return
 		}
+
+		orgUsecase := usecasesWithCreds(ctx, uc).NewOrganizationUseCase()
+		org, err := orgUsecase.GetOrganization(c.Request.Context(), organizationId)
+		if presentError(ctx, c, err) {
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


Pascal-Delange (approved):
👍 
(or could we have done the "enrich and adapt" logic directly in the usecase? it's not a big deal, but to avoid leaking this logic into the controller. It's really up to you though)

apognu:
> 👍 (or could we have done the "enrich and adapt" logic directly in the usecase? it's not a big deal, but to avoid leaking this logic into the controller. It's really up to you though)

This is a good question, though. This bit of logic is very specific to getting the input from a client which timezone we do not control, and that only gives the option for "accurate-to-the-day" timestamps, which is why I added it to the specific DTO that comes from the browser.

Adding it to the usecase, in my opinion, would couple the business logic to how the browser clients work, which would be a bad thing in my opinion (for example if we need, at some point, to allow creating test runs ending at specific times).

apognu:
Poke @Pascal-Delange for opinion.

Pascal-Delange:
@apognu thanks, makes sense !

PR #871, opened 2025-02-24
feat: make trigger condition optional

ChibiBlasphem:
This PR makes scenario trigger condition optional by:
- Setting the `trigger_condition_ast_expression` field to NULL when patching scenario iteration if the trigger is an undefined node
- Skipping trigger execution while evaluating scenario if the trigger is nil
- Getting rid of validation scenario error when trigger is nil


Pascal-Delange (approved):
👍

PR #873, opened 2025-02-25
Add git version on build in logs and OTEL metadata.

apognu:
This PR adds an endpoint to retrieve the running version of the API. It is a bit more involved than expected:

* Version defaults to `dev`
* If specified during build, it now uses `git describe --tags`, which will render:
  * `v0.36.1` if on a tag
  * `v0.36.1-2-g123456` if not (the latest tag, the number of commit since, and the sha of the commit), useful for staging
* It logs the version (still need to add logs to the other components)
* It sets the version in OTEL (@Pascal-Delange, do we want this?)
* It adds a `/version` endpoint for the frontend to use

@Pascal-Delange, right now the endpoint is unauthenticated, which could be considered a security issue (leaking the version of the software). But we want to display it on the login screen as well, which would require it being unauthenticated. Also, the version would be leaked by the frontend code itself anyway, so it would not be worse. Just pointing it out.

Also @Pascal-Delange, this touches the deploy script (through the `Dockerfile`, so I want to check I did not miss a usage of the initial `version` variable that I changed).


Pascal-Delange (approved):
👍  but the code to actually read the env var is missing, no ?
No worries about the unauthenticated endpoint, there is already another one (the liveness one)

apognu:
There is no env var, it uses a build option that sets a global variable during compilation (`-X main.apiVersion=`)

Pascal-Delange:
Oh, got it.
Didn't know you could set the value of a variable in a package like this from a build option 😅 
Then that's cool, but let's maybe add a small comment where the variable is defined to explain where its value comes from ? Otherwise it's a bit magical if you don't know this option.

PR #874, opened 2025-02-26
BuildKit: keep git directory to extract the version

apognu:
GitHub Actions uses BuildKit with the Git Context to build container images. By default, it does not keep the `.git` directory in the used working copy. This means that, by default, we cannot use `git` commands to extract the current version of the software.

This adds a BuildKit build argument to keep the directory around.


Pascal-Delange (approved):
👍

PR #879, opened 2025-03-03
WIP: Use embedded Postgres instead of Docker container.

apognu:
Right now, this uses my fork of embeddedpostgres with a patch allowing the use of UNIX sockets for the database, so not to have to worry about the availability of ports.

It simplifies setting up of the integration tests greatly. Also, running `m.Run` as `os.Exit` is no longer required, and lets us set up deferred shutdown properly.

This is a prorotype explortation, do not merge.


apognu:
Sent a PR to upstream project to have the UNIX socket patch merged in: https://github.com/fergusstrange/embedded-postgres/pull/152

apognu:
I'll close this for now.

PR #881, opened 2025-03-03
Fixed bug where all test runs would set as summarized.

apognu:


pascaldelange (approved):

Pascal-Delange (approved):

apognu:
This will require, after deployment, to reset the summary table (it that order):

```sql
update scenario_test_run set summarized = false;
truncate table scenario_test_run_summaries;
```

PR #882, opened 2025-03-04
Add async match enrichement 

apognu:
This also fixes an issue with overall enrichment where we replaced the payload completely (hence losing attributes specific to the search, e.g. `score`).

### DIFF
--- a/usecases/decision_usecase.go
+++ b/usecases/decision_usecase.go
@@ -123,6 +123,8 @@ type DecisionUsecase struct {
 	phantomUseCase            decision_phantom.PhantomDecisionUsecase
 	featureAccessReader       decisionUsecaseFeatureAccessReader
 	scenarioEvaluator         ScenarioEvaluator
+	openSanctionsRepository   repositories.OpenSanctionsRepository
+	taskQueueRepository       repositories.TaskQueueRepository
 }
 
 func (usecase *DecisionUsecase) GetDecision(ctx context.Context, decisionId string) (models.DecisionWithRuleExecutions, error) {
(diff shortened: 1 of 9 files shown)


usecases/decision_usecase.go
```diff
@@ -450,11 +452,16 @@ func (usecase *DecisionUsecase) CreateDecision(
...
-				decision.DecisionId, *decision.SanctionCheckExecution, true); err != nil {
+			sc, err := usecase.sanctionCheckRepository.InsertSanctionCheck(ctx, tx,
+				decision.DecisionId, *decision.SanctionCheckExecution, true)
+			if err != nil {
 				return models.DecisionWithRuleExecutions{},
 					errors.Wrap(err, "could not store sanction check execution")
 			}
+
+			if usecase.openSanctionsRepository.IsSelfHosted(ctx) {
+				_ = usecase.taskQueueRepository.EnqueueMatchEnrichmentTask(ctx, input.OrganizationId, sc.Id)
```
Pascal-Delange, line 463: let's still return the error, no ? Even if it's a secondary action

Pascal-Delange: (I can perhaps be converted to just logging it)

usecases/sanction_check_usecase.go
```diff
@@ -770,3 +789,23 @@ func (uc SanctionCheckUsecase) enforceCanReadOrUpdateSanctionCheckMatch(
 		match:    match,
 	}, nil
 }
+
+func mergePayloads(originalRaw, newRaw []byte) ([]byte, error) {
```
Pascal-Delange, line 793: good catch 👍

Pascal-Delange (approved):
👍

usecases/decision_usecase.go
apognu: Yes, I actually wanted to log and ignore the error, but forgot. I don't think the whole decision should fail if it fails to be enriched automatically, since it can always be enriched after the fact. I'll add the logging if you agree.

Pascal-Delange: yep, let's do that 👍  (you can use the `utils.LogAndReportSentryError(ctx, err)` function which, well, logs and reports to sentry)

PR #885, opened 2025-03-04
chore: try to optimize for covering index for PATCH/POST ingestion

Pascal-Delange:
Borrowing from the explanation comment in the PR:
```
// Try to only load the fields that are actually missing from the payloads (in the case of a partial update).
// The same method is used for the POST and PATCH endpoints, but if no fields are missing then there is a
// covering index on the object_id and updated_at columns that can be used.
// This makes the POST endpoint possibly faster to respond than the PATCH endpoint for ingestion (when there
// is data present and some payload have missing fields).
```

The PR that allowed partial payload ingestion (PATCH endpoints for ingestion) made changes that caused every ingestion call to always go read data from the base table (client data table), even though a covering index exists that is sufficient in many cases.

### DIFF
--- a/integration_test/scenario_flow_test.go
+++ b/integration_test/scenario_flow_test.go
@@ -369,7 +369,7 @@ func setupScenarioAndPublish(
 	// Wait for a potential 'index_creation' task to be created. If it is, we wait for both 'index_creation'
 	// and 'index_creation_status' to be completed before continuing.
 	if IfTaskIsCreated(ctx, riverClient, time.Second, "index_creation") {
-		WaitUntilAllTasksDone(t, ctx, riverClient, 5*time.Second, "index_creation", "index_creation_status")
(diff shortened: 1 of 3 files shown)


apognu (approved):

integration_test/scenario_flow_test.go
```diff
@@ -369,7 +369,7 @@ func setupScenarioAndPublish(
 	// Wait for a potential 'index_creation' task to be created. If it is, we wait for both 'index_creation'
 	// and 'index_creation_status' to be completed before continuing.
 	if IfTaskIsCreated(ctx, riverClient, time.Second, "index_creation") {
-		WaitUntilAllTasksDone(t, ctx, riverClient, 5*time.Second, "index_creation", "index_creation_status")
+		WaitUntilAllTasksDone(t, ctx, riverClient, 10*time.Second, "index_creation", "index_creation_status")
```
Pascal-Delange, line 372: @apognu  had to increase this, as I found that the integration test was failing randomly when running on my computed

PR #887, opened 2025-03-05
Implemented case snoozes.

apognu:
A case can now be snoozed, either until a specified date and time, or until a new decision is added to the case, whichever happens first.

A new filter can be provided when listing cases, `include_snoozed`, to keep displaying snoozed cases.

TODO:

 - [ ] Add case events when a case is snoozed?


usecases/case_usecase.go
```diff
@@ -362,6 +367,77 @@ func (usecase *CaseUseCase) UpdateCase(
...
+
+	return uc.transactionFactory.Transaction(ctx, func(tx repositories.Transaction) error {
+		if err := uc.repository.SnoozeCase(ctx, tx, req); err != nil {
+			return err
+		}
+
+		var previousSnooze *string
+
+		if c.IsSnoozed() {
+			previousSnooze = utils.Ptr(c.SnoozedUntil.Format(time.RFC3339))
```
apognu, line 388: @Pascal-Delange Quite unsure here, do we have another standard datetime format when we manually serialize it (without using a `timestmap`)? This seems aligned with what we return, but if it is not, I'll change it.

Pascal-Delange (approved):
The PR looks all good to me.
I think it's mergeable as it is, though we could also keep it open until someone (probably @carere ) works on the actual snooze action,  as it's unlikely to conflict with other changes.

usecases/case_usecase.go
```diff
@@ -435,6 +511,11 @@ func (usecase *CaseUseCase) AddDecisionsToCase(ctx context.Context, userId, case
 			return models.Case{}, err
 		}
 
+		err = usecase.repository.UnsnoozeCase(ctx, tx, caseId)
```
Pascal-Delange, line 514: Oh, one important thing.
There is another place (which represents most of the volume) where we add decisions to a case, it's in the UpdateDecisionsWithEvents method (which is called by the automatic decision to case workflow)

Pascal-Delange:
Same as the other PR, context on the next steps: https://linear.app/marble-eu/issue/MAR-812/revamp-case-statuses-to-better-match-client-workflows

Pascal-Delange (approved):

PR #891, opened 2025-03-06
Public API v1

apognu:
**Note:** there is quite a big change on how authorization is performed, both on the public API and the core API. I refactored the middleware to take a list of authorization methods allowed depending on the context (core API accepts either an API key or a bearer token, the public API only uses API keys).

This new public API is, for now, served as `/v1beta`.


specs/v1_public_api.yaml
```diff
@@ -0,0 +1,470 @@
...
+                  - type: object
+                    properties:
+                      data:
+                        $ref: '#/components/schemas/SanctionCheck'
+
+        400: { $ref: '#/components/responses/400' }
+        404: { $ref: '#/components/responses/404' }
+        422: { $ref: '#/components/responses/422' }
+
+  /decisions/{decisionId}/sanction-checks/search:
```
Pascal-Delange, line 90: A very minor question I have is, is this the right output format for our public API ?
Because here the output format is pretty strongly linked to the sanction check dto, while in general it would make sense to let customers just do a more "free form" search ?

repositories/sanction_check_whitelist_repository.go
```diff
@@ -0,0 +1,93 @@
...
+func (repo *MarbleDbRepository) SearchSanctionCheckMatchWhitelist(ctx context.Context,
+	exec Executor, orgId string, counterpartyId, entityId *string,
+) ([]models.SanctionCheckWhitelist, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+
+	filters := squirrel.Eq{"org_id": orgId}
+
+	if entityId != nil {
```
Pascal-Delange, line 59: I don't think we have an index on just entity_id (without counterparty_id), let's add one

Pascal-Delange (commented):
Looks good to me ! 
I just have this one open question where I'd like your opinion about the /search endpoint, which I think is a bit more awkward in the public API than in our internal API.

specs/v1_public_api.yaml
apognu: Will amend with the output just returning the list of raw payloads from OpenSanctions.

specs/v1_public_api.yaml
```diff
@@ -0,0 +1,470 @@
...
+                      - type: object
+                        properties:
+                          data:
+                            type: array
+                            items:
+                              $ref: '#/components/schemas/SanctionCheck'
+
+        404: { $ref: '#/components/responses/404' }
+
+  /decisions/{decisionId}/sanction-checks/refine:
```
Pascal-Delange, line 48: I just saw [this thread](https://checkmarble.slack.com/archives/C06P1FJV405/p1742305665652489) in our shared slack with Swan, which made me think: maybe for our public API, we don't want to be too strongly committed on the fact that there are exactly 0 or 1 sanction checks per decision, as they're making a good point of why they'd want to have several checks per decision.
On the GET endpoint above it's not a problem (it returns an array), but on the /refine endpoint here (and also for the search below, seeing as it uses the counterparty_id from the sanction check on the decision) I believe we should strive to to make it tightly coupled with this implementation choice.

apognu: Fair enough, so replacing `/decisions/{decisionId}` by `/sanction-checks/{sanctionCheckId]}` should cover this possible future need, right?

apognu: Up for review.

repositories/migrations/20250305110100_remove_whitelist_user_non_null.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+
+alter table sanction_check_whitelists
+    alter column whitelisted_by drop not null;
+
+-- +goose Down
+
+alter table sanction_check_whietlists
```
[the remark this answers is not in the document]
apognu: I know. 😢

specs/v1_public_api.yaml
```diff
@@ -263,14 +270,23 @@ paths:
...
+                      data:
+                        type: array
+                        items:
+                          $ref: '#/components/schemas/SanctionCheckWhitelistEntry'
 
         400: { $ref: '#/components/responses/400' }
 
 components:
   securitySchemes:
+    BearerTokenAuth:
```
Pascal-Delange, line 287: 🙌

Pascal-Delange (approved):
good to merge !

PR #895, opened 2025-03-06
fixes on watermark logic

Pascal-Delange:
The final release is going to be for tomorrow in the end.
There was a bit of a logic error related to the fact that the watermark is on the summaries, that no summaries are updated if no decisions exist in the window between [then, windowBound], where the code remains stuck (and a test run would remain unsummarized forever).
I fixed it by adding a "bump all watermarks after writing the summaries" step.
It was a bit of a pain to debug, because the issue would appear in a way that is very dependent on how the decisions on the scenarios are distributed in time.

### DIFF
--- a/usecases/scheduled_execution/test_run_summary_job.go
+++ b/usecases/scheduled_execution/test_run_summary_job.go
@@ -57,6 +57,9 @@ type RulesRepository interface {
 	SaveTestRunSummary(ctx context.Context, exec repositories.Executor,
 		testRunId string, stat models.RuleExecutionStat, newWatermark time.Time,
 	) error
+	BumpDecisionSummaryWatermark(ctx context.Context, exec repositories.Executor,
+		testRunId string, newWatermark time.Time,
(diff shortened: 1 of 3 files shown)


usecases/scheduled_execution/test_run_summary_job.go
```diff
@@ -16,7 +16,7 @@ const (
 	// The summary is not idempotent, so we cannot afford to have two processes running at the same time for the same organization.
 	// Be mindful to not set the timeout greater (or even close) to the interval, to prevent that.
 	TEST_RUN_SUMMARY_TIMEOUT         = 2 * time.Minute
-	TEST_RUN_SUMMARY_WORKER_INTERVAL = 5 * time.Minute
+	TEST_RUN_SUMMARY_WORKER_INTERVAL = 20 * time.Second
```
Pascal-Delange, line 19: to change before merging

apognu (approved):

PR #896, opened 2025-03-07
create at least one summary for test run stats even if no decisions a…

Pascal-Delange:
…re in the window, to avoid an infinite loop

I'm not super proud of this because of the coupling between the interface implementation and the place where it's used... But I don't want to invest much more time on this for now.


usecases/scheduled_execution/test_run_summary_job.go
```diff
@@ -16,7 +16,7 @@ const (
 	// The summary is not idempotent, so we cannot afford to have two processes running at the same time for the same organization.
 	// Be mindful to not set the timeout greater (or even close) to the interval, to prevent that.
 	TEST_RUN_SUMMARY_TIMEOUT         = 2 * time.Minute
-	TEST_RUN_SUMMARY_WORKER_INTERVAL = 5 * time.Minute
+	TEST_RUN_SUMMARY_WORKER_INTERVAL = 10 * time.Second
```
apognu, line 19: Penser à rollback.

apognu (approved):

PR #900, opened 2025-03-10
API endpoints for case assigment.

apognu:


api/routes.go
```diff
@@ -170,6 +170,8 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth Aut
 	router.POST("/cases/:case_id/decisions", tom, handlePostCaseDecisions(uc))
 	router.POST("/cases/:case_id/comments", tom, handlePostCaseComment(uc))
 	router.POST("/cases/:case_id/case_tags", tom, handlePostCaseTags(uc))
+	router.POST("/cases/:case_id/assignee", tom, handleAssignCase(uc))
```
apognu, line 173: Do we prefer the "assign" method to be part of the case `PATCH` endpoint?

api/handle_cases.go
```diff
@@ -259,6 +259,64 @@ func handlePostCaseTags(uc usecases.Usecases) func(c *gin.Context) {
...
+			return
+		}
+
+		req := models.CaseAssignementRequest{
+			UserId:     creds.ActorIdentity.UserId,
+			CaseId:     caseId,
+			AssigneeId: &payload.UserId,
+		}
+
+		if payload.UserId == "me" {
```
Pascal-Delange, line 281: to prepare for the day where, perhaps, we find time to replace all "strings" that in fact contain uuids by proper uuid types in the go code, i'd propose we avoid this and instead do the same if the id is empty (I mean, even like this it's easy enough to change later. But at least we won't need to change client code)

api/routes.go
Pascal-Delange: on one hand, having it separate makes sense to me, as it's an action that feels different from "update the body of the case"
on the other hand, I want to spend a little bit more time refining business rules we want to implement regarding case affectation (how it's related to doing other actions on it, etc).
In any case, we can merge this part for now.

Pascal-Delange (approved):
This PR looks all good to me.
I'd just like to have a quick chat with Arnaud about the product usage rules we want to put in place, based on actions taken on a case (for instance, does it auto-affect to you if you change status from pending - new status to be introduced - to investigating?)
But we can merge it like this in any case

Pascal-Delange:
For some context on what comes next (still refining): https://linear.app/marble-eu/issue/MAR-812/revamp-case-statuses-to-better-match-client-workflows

api/handle_cases.go
apognu: I agree that having proper UUIDs across would be a good idea indeed. I will change this.

PR #902, opened 2025-03-11
fix bug impacting index creation for test run

Pascal-Delange:
See linear for the details on the bug that triggered the ticket.
The fix was a bit trickier than I initially thought, because of how the ConcreteModel index was also used as a dto implicitly when serializing the job a rgs.
Also takes care of https://linear.app/marble-eu/issue/MAR-802/review-placement-of-index-name-logic-following-incident-on-index following the incident yesterday.

### DIFF
--- a/models/concrete_index.go
+++ b/models/concrete_index.go
@@ -1,11 +1,19 @@
 package models
 
 import (
+	"encoding/json"
+	"fmt"
 	"slices"
 	"strings"
 
 	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/google/uuid"
 	"github.com/hashicorp/go-set/v2"
+)
+
+const (
+	MAX_POSTGRES_INDEX_NAME_LENGTH      = 63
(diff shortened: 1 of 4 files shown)


apognu (approved):

models/concrete_index.go
```diff
@@ -17,9 +25,78 @@ type UnicityIndex struct {
 
 type ConcreteIndex struct {
 	TableName string
-	IndexName string
-	Indexed   []string
-	Included  []string
+	// name property is not thread safe. However, it is only expected to be accessed by one goroutine at a time.
```
apognu, line 28: I am trying to see the impact of this comment. Do you mean there could be a race between the goroutine reading it and the one writing it?

Pascal-Delange: Hmm no, I mean I propably over-commented here.
I just meant, weird things could happen if several goroutines started doing things on the struct concurrently, but then again that's always true, and in this case there is no reason (now, anyway) that the object be passed around goroutines after it's been initialized. I'll remove the comment

PR #903, opened 2025-03-12
add permission to write on list at builder role level (was publisher)

Pascal-Delange:
## Product changes
The builder role is able to add and remove items to a custom list. Previously, only publisher and admin could.

## Tech cleanup
I removed any permissions on some roles that seemed not useful to me, as far as I was able to verify.
This _should_ not break anything, but as always we'll need to test this properly in staging before we release. It's a good timing to do this anyway, because we have at least two other PRs on the frontend that will warrant in-depth QA before the next release (AST builder refactor and forms refactor).

- [x] tested locally for custom list entry edit

### dependencies
To merge together with https://github.com/checkmarble/marble-frontend/pull/737


models/role_permission.go
```diff
@@ -17,18 +17,15 @@ var VIEWER_PERMISSIONS = []Permission{
 }
 
 var (
-	BUILDER_PERMISSIONS   = append(VIEWER_PERMISSIONS, SCENARIO_CREATE)
-	PUBLISHER_PERMISSIONS = append(BUILDER_PERMISSIONS, SCENARIO_PUBLISH, CUSTOM_LISTS_CREATE)
+	BUILDER_PERMISSIONS   = append(VIEWER_PERMISSIONS, SCENARIO_CREATE, CUSTOM_LISTS_EDIT)
```
Pascal-Delange, line 20: To be clear, this line here was the only intended change in the linear ticket that brought us to this PR.

models/permission.go
```diff
@@ -84,7 +84,7 @@ func (r Permission) String() (string, error) {
 		"ANY_ORGANIZATION_ID_IN_CONTEXT",
 		"ANY_PARTNER_ID_IN_CONTEXT",
 		"CUSTOM_LISTS_READ",
-		"CUSTOM_LISTS_PUBLISH",
```
Pascal-Delange, line 87: for some reason "CUSTOM_LIST_CREATE" (the constant) was mapped to "CUSTOM_LISTS_PUBLISH" in the String() method, and this was used by the frontend.

PR #911, opened 2025-03-13
wip: try postgres 16 on marble

Pascal-Delange:
In my case at least, I wasn't able to just reuse the same volume.
It might just be that I had an error because I was using the pg_cron extension that is not present in the basic postgres16 image however, I needed to remove it anyway to run my test.
Procedure I followed: (Following https://thomasbandt.com/postgres-docker-major-version-upgrade)

```
docker exec -it  postgres pg_dumpall -U postgres > ./upgrade_backup.sql
./pg_extract.sh upgrade_backup.sql marble >> upgrade_backup_marble.sql
```
Where the script contains 
```bash
#!/bin/bash
[ $# -lt 2 ] && { echo "Usage: $0 <postgresql dump> <dbname>"; exit 1; }
sed  "/connect.*$2/,\$!d" $1 | sed "/PostgreSQL database dump complete/,\$d"
```

then
```
docker compose down
// -> modify postgres version in docker compose & change volume
docker compose up -d
cat ./upgrade_backup_marble.sql | docker exec -i postgres  psql -U postgres
```
Will not try to run with this locally for a few days.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,6 @@
 services:
   db:
     container_name: postgres
-    # image: postgres:15
-    # image: europe-west1-docker.pkg.dev/marble-infra/marble/postgresql-db:latest
     image: postgres:16
     shm_size: 1g
     restart: always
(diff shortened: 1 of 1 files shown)


Pascal-Delange:
Planning to merge this now. Continuation of migration process discussed in linear https://linear.app/marble-eu/issue/MAR-827/try-to-bump-to-postgres-16

apognu (approved):

PR #912, opened 2025-03-14
fix error: out of bounds panic when duplicates by object_id found in …

Pascal-Delange:
The implementation of the objection injection repository was not safe against duplicate entries by object id in the entry slice.
I did not previously catch it because there is an earlier check against this when sending objects by API, but it was not done by CSV ingestion.


.github/workflows/backend_test_workflow.yaml
```diff
@@ -24,4 +24,4 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        run: go test -v ./...
```
Pascal-Delange, line 27: Sneaked this into the PR.
Verbose tests are just impossible to read in the CI.

apognu (approved):

PR #914, opened 2025-03-14
E2E testing for public API

apognu:
This is a prototype for E2E testing on the public API. It deviates quite a lot from The Way™️ we do integration testing in the rest of the app, so to be considered.

It uses:

 * `testcontainers` for setting up the database. I still think we would be better off with embedded-postgres, but until my PR is merged, let's not. 
 * YAML fixtures to set up the database
 * Runs the API with a UNIX socket to not bother with port availability
 * ~[`hurl`](https://github.com/Orange-OpenSource/hurl) to write test specs~


apognu:
Actions failing on this is expected for now.

PR #916, opened 2025-03-17
Handle null index_version for OpenSanctions datasets.

apognu:
Before merging this, we need to understand tghe semantic behind `index_version`, should we say that we are up to date iof it is null or out of date?

### DIFF
--- a/repositories/httpmodels/http_opensanctions_dataset_freshness.go
+++ b/repositories/httpmodels/http_opensanctions_dataset_freshness.go
@@ -2,6 +2,7 @@ package httpmodels
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/checkmarble/marble-backend/models"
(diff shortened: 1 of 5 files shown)


repositories/httpmodels/http_opensanctions_dataset_freshness.go
```diff
@@ -57,16 +59,18 @@ func AdaptOpenSanctionsLocalDatasetFreshness(datasets HTTPOpenSanctionsLocalData
...
 	}
 
 	if version == nil {
-		return models.OpenSanctionsDatasetFreshness{}, errors.New(
-			"could not find upstream default dataset")
+		return models.OpenSanctionsDatasetFreshness{},
+			errors.New("could not find upstream default dataset")
 	}
 
-	lastUpdatedAt, err := time.ParseInLocation("20060102150405", (*version)[:len(*version)-4], time.UTC)
```
apognu, line 69: This logic is still unsound if we cannot rely on OpenSanctions version formats. They said the outliers in formatting were due to them not being updated in a long time, which _should_ not apply to the `default` dataset (which we are checking here), but it still raises the question of whether we should change the logic.

Pascal-Delange (approved):
let's already merge like this for the time being 👍

PR #918, opened 2025-03-18
chore(doc): Improve getting started section

siiick:
- Start the database container only in `docker compose` command.
- Use the default firebase cli install script since it's now fixed.
- Propose the usage of [mise](https://mise.jdx.dev/) to ease the env vars export and create the its config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,50 +9,67 @@ This repo is the Marble backend implementation:
 
 ### Setup your environment
 
-[Install Go](https://go.dev/doc/install) on your laptop (see the version in go.mod).
+> **Disclaimer**
+>
+> This repository’s README file is intended for internal use by our development team. The documentation provided here is specifically designed for setting up and running the project on macOS.
+>
+> While external contributions and interest are appreciated, please note that we do not officially support setups on other operating systems. If you encounter issues outside of the macOS environment, support may be limited.
+> 
+> For general documentation and user-facing guides, please refer to our main repository: [Marble Documentation](https://github.com/checkmarble/marble/blob/main/README.md).
 
-> NB: To handle different versions, you can look at [Managing Go installations](https://go.dev/doc/manage-install) or use a version manager tool like [asdf](https://github.com/kennyp/asdf-golang)
+[Install mise-en-place](https://mise.jdx.dev/getting-started.html) or alternatively [install Go](https://go.dev/doc/install) manually on your laptop (see the version in go.mod).
 
-Create you own `.env` file based on `.env.example`. You can fill it with your own values but it should work locally with the default values (except for third-party services functionalities).
+Create your own `.env` file based on `.env.example`. You can customize it with your own values, but it should work locally with the default values (except for third-party service functionalities).
 
-#### Setup the DB
+#### Setup the Database and Services
 
-Launch the postgres DB used by the backend:
+Launch the Postgres DB and other services used by the backend:
 
 ```sh
 docker compose up -d
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,31 +30,39 @@ docker compose up -d
 Install the [Firebase tools](https://firebase.google.com/docs/emulator-suite):
 
 ```sh
-# Install the Firebase CLI using npm, since the curl script is broken (on 06/09/2024)
-npm install -g firebase-tools
+# Install the Firebase CLI
+curl -sL https://firebase.tools | bash
```
apognu, line 34: I believe if default installations work, we should maybe use the OS packages managers to install Firebase instead of piping to she'll. We won't be covering all systems, but I think at least indicating the brew command for macOS and linking the the docs for other operating system would be better.

README.md
```diff
@@ -20,7 +20,7 @@ Create you own `.env` file based on `.env.example`. You can fill it with your ow
 Launch the postgres DB used by the backend:
 
 ```sh
-docker compose up -d
+docker compose up -d db
```
apognu, line 23: This would not launch the second container, Yente, which might be what we want for a speedy startup, but that will be missing a dependency for Sanctions Check.

Yente was added after the fact, and I think it was the introduction that was wrong.

README.md
```diff
@@ -30,31 +30,39 @@ docker compose up -d
...
-firebase --project GOOGLE_CLOUD_PROJECT emulators:start --import=./firebase-local-data --export-on-exit
+firebase --project [GOOGLE_CLOUD_PROJECT] emulators:start --import=./firebase-local-data --export-on-exit
 ```
 
 > NB: The `--import` flag is used to import the local data into the emulator. The `--export-on-exit` flag is used to export the data when the emulator is stopped so you don't lose your changes.
 
 ### Launch the project
 
-Export your `.env` file and run the root of the project:
+Export your `.env` file (e.g. using [direnv](https://direnv.net/) or [mise](https://mise.jdx.dev/)) and run the root of the project:
```
apognu, line 55: If we are suggesting mise, maybe add it to the top the file as well, when we instruct to install Go.

README.md
siiick: As discussed with @apognu and @Pascal-Delange, we can focus our main documentation on our customers ([Marble repository](https://github.com/checkmarble/marble)), while keeping the submodule documentation (backend and frontend) intended for the internal development team.

Let’s add a disclaimer to this README file to clarify that we officially support the developer experience on macOS only.

README.md
siiick: done

README.md
siiick: done

README.md
siiick: done

README.md
siiick: done

README.md
apognu: I am still of the opinion we should advertise `brew install firebase-cli` for this instead of piping to shell, which is considered a dubious way to install software at best or a security issue at worst.

siiick: I went with the automatic install script since the homebrew formula is not proposed in [the official doc](https://firebase.google.com/docs/cli#install-cli-mac-linux).
Also our old doc said "curl script".
What do you think if I add the homebrew formula as an alternative ?

apognu: I'd say keep the official way, and let's not muddy the waters by having too many options. 👍

Pascal-Delange (approved):

PR #919, opened 2025-03-19
Delete unused indices.

apognu:
When creating scenario iterations that perform aggregate queries, the prepare step creates database indices in the background. Those indices are never deleted if the live scenario iteration does not require them anymore, which contributes to database bloat and slower ingestion.

This PR adds a worker that compares all required indices to those existing in the database, and deletes them if they are deemed not necessary for operations.

Since index creation can be a time-consuming operation that can delay a scenario publication, we are somewhat lenient as to what we consider a required index. As it is, we expand the notion of "live iteration" to "live iteration and neighbors", which means the previous iteration and all next iterations.

For example, if a scenario has iterations 1-13, with iteration 10 being live, we will consider iterations 9-13 as live-adjacent and keep their indices.

We should only consider indices that are used for aggregations here, so we explicitly exclude other indices based on their names (which is why it is important, when adding a new type of index, to name it carefully). As of now, we have four types of indices on the client tables:

 * Primary keys (`*_pkey`)
 * Navigation indices (`nav_*`)
 * Unicity indices (`uniq_idx_*`)
 * Aggregation indices (`idx_`)

---

An edge case was found if a user re-prepares (after its induces were deleted) an old version of a scenario: the job would, if it runs after the preparation, immediately delete the newly-created index, and the old version could not be activated.

A tentative commit was added on top of this tree to bump an iteration `updated_at` when preparing it so it can be included in the "live-adjacent" iterations (if an iteration's `updated_at` is more recent than that of the live iteration, it means it could become live).

### DIFF
--- a/repositories/scenarios_read.go
+++ b/repositories/scenarios_read.go
@@ -82,11 +82,10 @@ func (repo *MarbleDbRepository) ListAllScenarios(ctx context.Context, exec Execu
 	)
 }
 
-// ListLiveIterationsAndNeighbors returns a list of scenario iterations, whatever the scenarios is,
-// that may considered live-adjacent.
-// It obviously returns the actual live iterations, but also one previous version and all next versions.
-// For example, if a scenario has a live iteration of 10, but has iterations from 1 to 13, this will returns
-// iterations 9 to 13.
+// ListLiveIterationsAndNeighbors returns a list of scenario iterations,
+// whatever the scenarios is, that may considered live-adjacent. It returns the
+// live iterations and all iterations that were live within a time period before
+// the current time.
 //
 // The final query looks like this (useful for debugging):
 /*
@@ -111,7 +110,7 @@ func (repo *MarbleDbRepository) ListAllScenarios(ctx context.Context, exec Execu
 	    where
 	      org_id = '<org_id>' and
 	      publication_action in ('publish', 'prepare') and
-	      created_at > now() - interval '1 hour'
+	      created_at > now() - interval '72 hour'
 	  )
 	select
 	  si.*
@@ -156,10 +155,10 @@ func (repo *MarbleDbRepository) ListLiveIterationsAndNeighbors(ctx context.Conte
 					Where(squirrel.And{
 						squirrel.Eq{
 							"org_id":             orgId,
-							"publication_action": []string{models.Prepare.String(), models.Publish.String()},
+							"publication_action": []string{models.Prepare.String(), models.Publish.String(), models.Unpublish.String()},
 						},
 						squirrel.Gt{
-							"created_at": time.Now().Add(-time.Hour),
+							"created_at": time.Now().Add(-3 * 24 * time.Hour),
 						},
 					})
 			})
--- a/usecases/scheduled_execution/index_rollover_job.go
+++ b/usecases/scheduled_execution/index_rollover_job.go
@@ -144,7 +144,7 @@ IndexDeletion:
 			}
 		}
 
-		logger.Debug(fmt.Sprintf("index %s.%s.%s is candidate for deletion", exec.DatabaseSchema().Schema, index.TableName, index.Name()),
+		logger.DebugContext(ctx, fmt.Sprintf("index %s.%s.%s is candidate for deletion", exec.DatabaseSchema().Schema, index.TableName, index.Name()),
 			"schema", exec.DatabaseSchema().Schema,
 			"table", index.TableName,
 			"index", index.Name(),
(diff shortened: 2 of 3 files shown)


usecases/scheduled_execution/index_rollover_job.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// TODO: also include required indices for future iterations, not yet live
+	requiredIndices, err := w.indexRepo.GetRequiredIndices(ctx, job.Args.OrgId)
+	if err != nil {
+		return err
+	}
+
+IndexLoop:
+	for _, index := range validIndices {
```
apognu, line 77: @Pascal-Delange Think about adding all whitelisted index prefixes from #934 to avoid deleting important indices.

repositories/scenarios_read.go
```diff
@@ -78,3 +81,102 @@ func (repo *MarbleDbRepository) ListAllScenarios(ctx context.Context, exec Execu
...
+					Where("si.org_id = ? and str.status = 'up'", orgId)
+			}).
+			With("neighbors", func(b squirrel.StatementBuilderType) squirrel.SelectBuilder {
+				return b.
+					Select("sp.scenario_iteration_id as id").
+					From(dbmodels.TABLE_SCENARIOS_PUBLICATIONS + " sp").
+					Where(squirrel.And{
+						squirrel.Eq{
+							"org_id":             orgId,
+							"publication_action": []string{models.Prepare.String(), models.Publish.String()},
```
Pascal-Delange, line 159: I think we want to also take into account recently deactivated iterations so we're still able to easily rollback in all cases for a couple of days - in other words also include the "unpublish" events, at which point we might as well just remove this filter altogether

usecases/scheduled_execution/index_rollover_job.go
```diff
@@ -0,0 +1,179 @@
...
+			continue
+		}
+
+		for _, optimalIndex := range optimalIndices {
+			if optimalIndex.Equal(index) {
+				continue IndexDeletion
+			}
+		}
+
+		logger.Debug(fmt.Sprintf("index %s.%s.%s is candidate for deletion", exec.DatabaseSchema().Schema, index.TableName, index.Name()),
```
Pascal-Delange, line 147: Let's `logger.XXXContext(` so we have traceIds

usecases/scheduled_execution/index_rollover_job.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	if len(errs) > 0 {
+		logger.Error("some indices failed to be deleted",
+			"indices", slices.Collect(maps.Keys(errs)),
+			"errors", pure_utils.Map(slices.Collect(maps.Values(errs)), func(err error) string {
+				return err.Error()
+			}))
+
+		return errors.New("could not delete all outdated indices")
```
Pascal-Delange, line 175: the only problem with this is that we lose sentry reporting of errors - maybe pick at least the first error out of the slice and log it with sentry using utils.LogAndReportSentryError ?

Pascal-Delange (commented):
I left a few last comments, looks about good to merge in dry run mode.
I just realized something that's missing (even in current code): we have a dead angle on any aggregates used in screening rule trigger conditions that we don't take into account anywhere. Possibly makes most sense to include this in a separate PR since it's more of a (possible, rare) bugfix than directly related to this feature, although it also indirectly concerns it (we could add the fetching of screening configs with the iterations in the big SQL query)

usecases/scheduled_execution/index_rollover_job.go
Pascal-Delange: Ok I think I got it.
The key in which to put the traces is provider specific, as such the place where it makes sense to handle it is not in LoggerFromContext (which is general purpose) but in the custom made GCP logger's `Handle(ctx context.Context, r slog.Record) error` function.

apognu: The handler makes sense to _marshal_ the traceId as `projects/x/traces/y`, but we could set the traceId on the logger upstream, and that attribute can be retrieved by the handler to be formatted according to the provider's idioms.

I mean, we do the same for `msg => message`, we retrieve the message set upstream and change its name, which, in my opinion, should be what we do, insert the traceId all the way up when the logger is created, and only format it in the handler.

Pascal-Delange: ok, got it.
Sure why not, it's probably a bit cleaner to do it this way around 👍

Pascal-Delange (approved):
LGTM!

PR #923, opened 2025-03-21
Free-form sanctioned entity search + rule snoozing

apognu:
This PR adds an endpoint to the public API to perform context-less, free-form searches for sanctioned entities. The first version does not allow for any configuration.

### DIFF
--- a/api/handle_snoozes.go
+++ b/api/handle_snoozes.go
@@ -67,7 +67,7 @@ func handleSnoozeDecision(uc usecases.Usecases) func(c *gin.Context) {
 			Duration:       input.Duration,
 			OrganizationId: organizationId,
 			RuleId:         input.RuleId,
-			UserId:         userId,
+			UserId:         &userId,
 		})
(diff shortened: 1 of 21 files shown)


Pascal-Delange (commented):
Reading this shorter PR made me realize something that's missing in several of the new public API endpoints (including this one): authorization per role/per organization on the actions that are taken.

apognu:
Aren't those going to be handled by the `enforce` functions in the usecases?

Pascal-Delange:
> Aren't those going to be handled by the `enforce` functions in the usecases?

Well yes they should, but that's the thing, there is for example no call to the "enforce" functions in CreateWhitelist, or in this added usecase method

apognu:
Good point. I thought `Execute` would donthose check, but it doesn't. I'll go through the endpoints and make sure there are permissions checks on the way.

Pascal-Delange (approved):

PR #924, opened 2025-03-21
Add rule snoozing through public API v1

apognu:


usecases/rule_snoozes.go
```diff
@@ -292,6 +294,8 @@ func (usecase RuleSnoozeUsecase) SnoozeDecision(
 		},
 	)
 
+	// TODO: do we really ignore errors here?
```
apognu, line 297: @Pascal-Delange I am under the impression that we ignore any error coming out of the transaction. Is this on purpose?

usecases/rule_snoozes.go
```diff
@@ -233,9 +229,9 @@ func (usecase RuleSnoozeUsecase) SnoozeDecision(
 		}
 
 		if len(snoozes) > 0 {
-			return models.SnoozesOfDecision{}, errors.Wrapf(
+			return models.SnoozesOfDecision{}, errors.WithDetail(
 				models.ConflictError,
-				"rule %s already has an active snooze %s", input.RuleId, input.DecisionId)
+				fmt.Sprintf("rule %s already has an active snooze %s", input.RuleId, input.DecisionId))
```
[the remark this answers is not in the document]
Pascal-Delange: ok you're out

usecases/rule_snoozes.go
```diff
@@ -169,16 +165,16 @@ func (usecase RuleSnoozeUsecase) SnoozeDecision(
...
-			"decision %s not found", input.DecisionId)
+			fmt.Sprintf("decision %s not found", input.DecisionId))
 	}
 	decision := decisions[0]
 
 	if decision.Case == nil {
-		return models.SnoozesOfDecision{}, errors.Wrapf(
-			models.BadParameterError,
-			"decision %s is not attached to a case and cannot be snoozed", input.DecisionId)
+		return models.SnoozesOfDecision{}, errors.WithDetail(
```
Pascal-Delange, line 175: This usecase method was used for the context of snoozing a decision rule by a human user (with inbox enforcement, audit trail etc).
Here the business logic for snoozing as an API key is a bit different, it's probably worth having a separate endpoint method.

apognu: Duplicated the methods to support needing a case from the frontend, and not from the API. Not very happy with the code duplication (or even with my attempt of extracting some of the logic to another function, I might roll that back).

Pascal-Delange:
@apognu have you pushed the latest changes to re-review ?

apognu:
Could not do it, will have to be tomorrow morning

PR #927, opened 2025-03-24
Fix memory leak in match enrichment

apognu:
Two issues surfaced in the match enrichment job:
* It would still run if the server was properly configured with self-hosted Yente, but not the worker. We want to short-circuit execution in the worker in those cases.
* If the enrichment process had a response from OS but a failing status code, the unclosed body would cause a memory leak.


Pascal-Delange (approved):
👍

PR #931, opened 2025-03-25
Implement object annotations.

apognu:
WIP PR to add ability to add annotations on ingested objects. The first version will support:

* Files
* Tags
* Comments


models/entity_annotation_payload.go
```diff
@@ -0,0 +1,31 @@
+package models
+
+type EntityAnnotationPayload interface {
+	entityAnnotationPayload()
+}
+
+type EntityAnnotationCommentPayload struct {
```
Pascal-Delange, line 7: Would be cool to have somewhere in the code the full expected format of what's in the annotation payloads - e.g. "tag" contains a tag id corresponding to table XXX, key/filename has the following format... etc.
Probably in this file is the most natural position. And link to it from the dto/dbmodels for discoverability ?

models/entity_annotation.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type EntityAnnotation struct {
+	Id             string
+	OrgId          string
+	ObjectType     string
+	ObjectId       string
+	AnnotationType EntityAnnotationType
+	Payload        json.RawMessage
+	AnnotatedBy    *UserId
```
Pascal-Delange, line 50: Another product requirement (was not explicited in the linear issue) would be to store the case id in the context of which an annotation was written - and correspondingly, to allow to list all annotations written in the context of a case.

Pascal-Delange (commented):
A last point, on the of the inline comments: it would be great to have a usecase/repository method that lists annotations of all types (without details, esp. without file urls) for a list of object_ids (corresponding to objects of the same type), ideally returned grouped by object_id

repositories/entity_annotation_repository.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	filters := squirrel.Eq{
+		"id":         req.AnnotationId,
+		"org_id":     req.OrgId,
+		"deleted_at": nil,
+	}
+
+	if req.AnnotationType != nil {
+		filters["annotation_type"] = req.AnnotationType.String()
```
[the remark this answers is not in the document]
apognu: None. The filter stayed in the copy and paste from the list method.

repositories/migrations/20250325084800_create_entity_annotations.sql
```diff
@@ -0,0 +1,31 @@
...
+    object_id text not null,
+    case_id uuid,
+    annotation_type text not null,
+    payload jsonb not null,
+    annotated_by uuid,
+    created_at timestamp with time zone not null default now(),
+    deleted_at timestamp with time zone default null,
+
+    primary key (id),
+    foreign key (org_id) references organizations (id),
```
Pascal-Delange, line 16: Last question I meant to ask you, what's your feeling on "on delete" conditions for foreign keys?
Most of the old ones have "on delete cascade/set null". Though I understand it can be a source of errors.

apognu: No, that's a good point, I tend to omit those completly. Overall, I am in favor, given that a sensible behavior is selected, and not cascade by default. It all depends on internal policy, do we want to support / help in the deletion of orgs, for example?

In this example, I agree that having `on delete set null` for `case_id` is a no brainer. But I have worked on projects where an `on delete cascade` of the magnitude of `orgs->world` was quite frowned upon.

Pascal-Delange: 👍 
Let's add the "set null" for the case, user id for starters.
We have a lot of "on delete org cascade", but I'll grant you that while it's convenient I'm not 100% at ease with keeping this at scale, as I would not want to deal with the fallout of an error. Something to think of for existing FKs.

Pascal-Delange: (PS: regarding users, we already have a soft delete on them so they really shouldn't be hard deleted - other then when we delete an org, as of today)

Pascal-Delange (approved):
🚀

PR #933, opened 2025-03-25
Remove duplicated operation ID and formatting.

apognu:
Small PR to fix errors on the public API v1 OpenAPI spec.

There is still a warning (body on `DELETE`), which has no semantics but is allowed with HTTP 1.1, so I unilaterally opt into ignoring it.


apognu:
Before merging this, I would like to try and integrate it (hidden) on Readme.

Pascal-Delange (approved):

PR #938, opened 2025-03-28
Add endpoint to get cases with same pivot value.

apognu:


repositories/case_repository.go
```diff
@@ -413,6 +413,25 @@ func (repo *MarbleDbRepository) GetCasesFileByCaseId(ctx context.Context, exec E
...
+
+	sql := NewQueryBuilder().
+		Select(columnsNames("c", dbmodels.SelectCaseColumn)...).
+		From(dbmodels.TABLE_DECISIONS + " d").
+		InnerJoin("cases c on c.id = d.case_id").
+		Where(squirrel.Eq{
+			"d.org_id":      orgId,
+			"d.pivot_value": pivotValue,
+		}).
+		OrderBy("c.created_at DESC").
```
apognu, line 429: This is going to change depending on which cases we should prioritize showing. We talked about pulling up cases that were confirmed for example. Let's wait until we have all those status to do that.

repositories/case_repository.go
```diff
@@ -413,6 +413,27 @@ func (repo *MarbleDbRepository) GetCasesFileByCaseId(ctx context.Context, exec E
...
+	}
+
+	sql := NewQueryBuilder().
+		Select(columnsNames("c", dbmodels.SelectCaseColumn)...).
+		Distinct().
+		From(dbmodels.TABLE_DECISIONS + " d").
+		InnerJoin("cases c on c.id = d.case_id").
+		Where(squirrel.Eq{
+			"d.org_id":      orgId,
+			"d.pivot_id":    pivotId,
```
Pascal-Delange, line 428: Filter by pivot_id is perhaps not useful.
In any case it's effectively not used anywhere currently, where we filter/group decisions by pivot_value. (nb: this could have some strange side effects for customers who use pivot values that are non-random, like incremented integers...).
Filtering by pivot_id, we lose the possibility to see any cases/decisions on objects of another type, that are still related to the same end user of the customer.

apognu: I probably do not understand something, but in my mind, that would be a justification to _actually_ filter by `pivot_id`. For example, let's say we have a `123` account and a `123` transaction. Retrieving cases that link to both when they are unrelated would not yield the correct result, would it?

That was my rationale behind adding the filter on `pivot_id`: getting cases that include decisions related to the same pivot, not "pivot with the same string in the value".

Let's discuss to make sure I understand this.

Pascal-Delange: This is technically correct, though the reality is a bit more nuanced, which largely goes down to how pivot values were first introduced (namely, for rule snoozing).

In the current implementation,
- every entity in the data model can have one (and only one) pivot defined, which can be a field of the entity itself or an object it links to (more precisely, a unique field of an object if links to, through a series of links)
- the pivot value defined as such has a pivot_id which just points to its configuration
- you cannot change a table's pivot value once it's been set (other by manually in the db)
- the way they are consumed today is:
   - for a given scenario, we maintain a set of snoozed rules, for decisions on the scenario that share the same pivot value (that actual computed, scalar, value, not the configuration) => it's "pivot_value" based but implicitly they are expected to share the same pivot_id, unless you modified it manually in db
   - we group decisions in the case manager by pivot_value, however in this case (as in the situation that started this conversation) we could in theory have the same scalar value, but corresponding to different fields/entities.

Also, implicitly we expect (nothing enforced to this effect yet) that the pivot configuration _SHOULD_ point to the same object/field pair, for every entity where it's configured. In other words, 
> let's say we have a 123 account and a 123 transaction

should not really happen if the customer is using the feature "the way we thought of it". That being said, the whole pivot value creation flow needs to be reviewed, if only because it's hard to discover and understand today.
(as a side note: I think [there was doc on this](https://docs.checkmarble.com/docs/adding-a-decision-to-an-existing-case), but the page seems to have disappeared 🤔)

The "technically correct" implementation - which we could try to move to but perhaps while keeping backwards compatibility for existing decisions - would be to store not (only?) the pivot_id but rather the entity/field pair that a given pivot_value corresponds to on the decision, and add that as a filter.

Pascal-Delange: For future record, as discussed in DM: "pivot_id" would group by "start table", we want to group by "destination field" - if we see the pivot value configuration as an arrow from an entity to its pivot value

Pascal-Delange (approved):
🚀

PR #939, opened 2025-03-28
Endpoint to get a case's pivot objects details

apognu:


Pascal-Delange (commented):
This PR looks like it starts to implement something in the same direction of what I wrote here https://github.com/checkmarble/marble-frontend/pull/760/files#diff-a8730fc82f48a0b94ce5f1c23c4b0d175b065f84a458089a07abea88e2c618f8R383 - I'd say you can pause it because I was planning to do this part

apognu:
So we can close this PR?

PR #940, opened 2025-03-28
feat(custom-list): Count the values in each list

siiick:
Add a values_count field to the custom list.

### DIFF
--- a/repositories/dbmodels/db_custom_list.go
+++ b/repositories/dbmodels/db_custom_list.go
@@ -15,7 +15,7 @@ type DBCustomListResult struct {
 	CreatedAt   time.Time  `db:"created_at"`
 	UpdatedAt   time.Time  `db:"updated_at"`
 	DeletedAt   *time.Time `db:"deleted_at"`
-	ValuesCount *int       `db:"values_count"`
+	NumberItems int        `db:"nb_items"`
 }
 
 const TABLE_CUSTOM_LIST = "custom_lists"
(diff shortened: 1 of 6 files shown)


Pascal-Delange (commented):
As it is, I think this all works 👍 
Regarding the counts, here is what I'd suggest ideally for this case:
- always return the count
- however, add a "limit N"  in a nested subquery to do the count, so that you'll never get more than N in the count
- consequently, don't return a raw count but instead a nested object with `{count: ..., maybe_more: true/false}`
I think this would be the best compromise for our frontend here.

If at any time we want to expose the list custom lists endpoint in the public API, I'd probably do it differently however (perhaps not returning the count by default, or even returning it only on a "get by id" call)

Btw, we've been trying to keep "design" discussions such as this in the github PR body & comments rather than on slack - to the extent that it's practical - especially for the sake of "our github is public so let's give a good image of our work on it" :-)

Pascal-Delange (approved):
👍 
You can try to have another go at it with squirrel, but honestly don't get stuck and lose time with it if it doesn't work out of the box, better to move on to the next thing that has impact

repositories/dbmodels/db_custom_list.go
```diff
@@ -15,20 +15,28 @@ type DBCustomListResult struct {
...
 		Name:           db.Name,
 		Description:    db.Description,
 		CreatedAt:      db.CreatedAt,
 		UpdatedAt:      db.UpdatedAt,
 		DeletedAt:      db.DeletedAt,
-	}, nil
+	}
+
+	customList.ValuesCount = &models.ValuesInfo{
+		Count:   db.NumberItems,
```
Pascal-Delange, line 37: ah sorry, I missed a point and seeing your frontend PR made me realize it.
The idea here was to always return "100" at most, so that we can display "> 100 values" in the app.
So the logic here should be to set `count: max(db.NumberItems, models. VALUES_COUNT_LIMIT)` (pseudo-code)

siiick: Done in https://github.com/checkmarble/marble-backend/pull/940/commits/341d1587837023ec05cf5be1b882bd61653aaf74

We actually want the **min** value between the count and the limit.

Pascal-Delange: right, my bad

PR #943, opened 2025-04-03
New statuses and outcomes for cases + SAR + boost

apognu:
This PR implements the status refactor for cases. Namely, we are changing the existing statuses, and adding a new set of outcomes for each case.

It also adds support for managing Suspicious Activity Reports in a case.

TODO:

 - [x] Add new statuses
 - [x] Add outcome
 - [x] Add migrations betweens statuses
 - [x] Implement business rules around statuses
 - [x] Create / update / delete suspicious activity reports
 - [x] Upload files to suspicious activity reports
 - [x] Add Waiting For Action (`boost`) on cases


repositories/migrations/20250403105900_cases_new_statuses.sql
```diff
@@ -0,0 +1,42 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- TODO: let's determine if we want explicit concurrency control here (by locking the table, for example)
```
apognu, line 4: This, obviously.

usecases/case_usecase.go
```diff
@@ -388,6 +399,8 @@ func (uc *CaseUseCase) Snooze(ctx context.Context, req models.CaseSnoozeRequest)
 		return err
 	}
 
+	// TODO: can we snooze/unsnooze a case if its status is pending or closed?
```
apognu, line 402: This applies potentially to other endpoints, namely:

* Do we want to change a case's status on some action (like assign, update, etc)?
* Do we want to prevent some actions depending on status (like "cannot reassign a closed case")?

apognu:
@Pascal-Delange Not finished but requesting a milestone review to steer back into the right direction if necessary.

Pascal-Delange (commented):
sorry it's a bit late - I read the parts on SAR, I'll go through the parts related to case status tomorrow morning.

repositories/migrations/20250407110500_add_case_waiting_for_action.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
```
Pascal-Delange, line 1: for once we're locking the tables anyway and we're not doing the index concurrently, so we can run it in a transaction (goose statementBegin) I think?

repositories/migrations/20250403105900_cases_new_statuses.sql
Pascal-Delange: also, before we run it for real, let's do a copy of the tables table and do a test run in prod to time it

repositories/migrations/20250407110500_add_case_waiting_for_action.sql
```diff
@@ -0,0 +1,25 @@
...
+alter table cases
+    add column boost text default null;
+
+create index idx_cases_add_to_case_workflow
+    on cases (org_id, inbox_id, id)
+    where (status in ('pending', 'investigating'));
+
+drop index cases_add_to_case_workflow_idx;
+
+create index idx_inbox_cases
```
Pascal-Delange, line 12: given the order we use in the query in the repo, we need an index on `cases(org_id, inbox_id, boost, created_at [, id])` no ?
(and we'll want to have another one on assignee_id, boost, created_at to filter for "my cases" I suppose, but not directly related to this PR).

Btw, why the mix on using waiting/boost for filtering/sorting ? My understanding was that "boost" just holds the reason for the "waiting" flag "just in case", but all business logic should work just based on the value of "waiting" ? (avoids handing "nulls last" etc in the query)

Pascal-Delange (commented):
👍 
A few questions remaining, mostly on indexes, ordering of cases related to the "waiting/boost" flag

repositories/migrations/20250407110500_add_case_waiting_for_action.sql
apognu: The `waiting` column shouldn't be there anymore, I changed it but forgot to remove it from the other migration. I didn't the the necessity to have two column descriving the same thing, one of them only being a boolean.

models/case.go
```diff
@@ -36,26 +39,55 @@ func (c Case) IsSnoozed() bool {
...
 
 type CaseStatus string
 
 const (
-	CaseOpen          CaseStatus = "open"
+	CasePending       CaseStatus = "pending"
 	CaseInvestigating CaseStatus = "investigating"
-	CaseDiscarded     CaseStatus = "discarded"
-	CaseResolved      CaseStatus = "resolved"
+	CaseClosed        CaseStatus = "closed"
```
Pascal-Delange, line 58: Capture d’écran 2025-04-08 à 23 21 05

usecases/case_usecase.go
Pascal-Delange: => maybe just remove the comment, to avoid future ambiguity

repositories/case_repository.go
```diff
@@ -434,5 +452,5 @@ func (repo *MarbleDbRepository) GetCasesWithPivotValue(ctx context.Context, exec
 }
 
 func orderConditionForCases(p models.PaginationAndSorting) string {
-	return fmt.Sprintf("c.%s %s, c.id %s", p.Sorting, p.Order, p.Order)
+	return fmt.Sprintf("c.boost nulls last, c.%s %s, c.id %s", p.Sorting, p.Order, p.Order)
```
Pascal-Delange, line 455: I thought you reworked on this, but not yet in fact ?

apognu: Not yet indeed, planned for tomorrow first thing.

apognu:
Getting ready to merge.

PR #948, opened 2025-04-08
chore: use file instead of inline env var for JWT signing key

siiick:
The example JWT signing key contains characters that break the inline environment variable parser.
To avoid parsing issues and improve safety, the key has been moved from the .env file to a separate file.


Pascal-Delange (approved):
👍

apognu (approved):

PR #949, opened 2025-04-08
feat: fuzzy match aggregate filter

siiick:
- **feat(AST-builder): add simple Fuzzy match filter**

### DIFF
--- a/usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
+++ b/usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
@@ -30,7 +30,8 @@ func (f FuzzyMatchOptionsEvaluator) Evaluate(ctx context.Context, arguments ast.
 		))
 	}
 
-	threshold, err := AdaptNamedArgument(arguments.NamedArgs, "threshold", promoteArgumentToFloat64)
+	threshold, err := AdaptNamedArgument(arguments.NamedArgs, "threshold",
+		promoteArgumentToFloat64) // Threshold is received as an integer (0–100) from the client and converted to a float (0.0–1.0) for internal processing.
 	if err != nil {
 		return nil, []error{err}
 	}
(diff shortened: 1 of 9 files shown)


usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/checkmarble/marble-backend/models/ast"
+	"github.com/cockroachdb/errors"
+)
+
+type FuzzyMatchOptionsEvaluator struct{}
+
+var allowedFuzzyMatchAlgorithms = []string{
+	"bag_of_words_similarity_db",
+	"direct_string_similarity_db",
+	"token_set_ratio",
```
Pascal-Delange, line 17: to be removed before merging, I just added this one to make it work with the frontend initially

usecases/indexes/aggregate_query.go
```diff
@@ -138,12 +138,15 @@ func aggregationNodeToQueryFamily(node ast.Node) (models.AggregateQueryFamily, e
 			if !family.EqConditions.Contains(fieldName) {
 				family.IneqConditions.Insert(fieldName)
 			}
+		case ast.FILTER_FUZZY_MATCH:
```
Pascal-Delange, line 141: oh, I didn't see this one when I first read the PR. An empty "case" in a switch statement in go is not implicitly grouped with the following one like in js (instead, it will just do nothing). The case should be added to the list in the one below.

Pascal-Delange (commented):
Two last requests:
- one comment inline, about switch case => see comment
- please add a unit test for the case of using fuzzy match filter - both algorithms. There already is a unit test file repositories/ingested_data_read_repository_test.go that you can extend, so it should be fairly simple to add it. This is because the query it generates is non-trivial, and the details of it are not tested at all when we dry run the scenarios.

usecases/indexes/aggregate_query.go
siiick: done thank you 👍

Pascal-Delange (approved):
yay, happy to see this merged ! 🎉

Pascal-Delange:
@siiick  I pushed some unit tests for the important new methods

siiick:
> @siiick I pushed some unit tests for the important new methods

Oh, thank you! I’ve been more focused on frontend these days.

PR #950, opened 2025-04-10
fix error where unnecessary indexes were being created

Pascal-Delange:
### Swan bug:
First commit
Index `(account_id DESC) INCLUDE (booked_at)` being created even though there is already an index `(account_id DESC, booked_at DESC, status DESC) INCLUDE (available_balance_after)`
=> Logic error in the "does an existing index cover our needs" logic

### Payrex bug:
Second commit
See issue https://github.com/checkmarble/marble/issues/91
Index parsing is broken for escaped field names. I thought we wouldn't have any, because we (currently) don't escape field names when we create the tables (we probably should, in which case this is a good preparation PR). However, we still have escaped field names if reserver sql keywords are used as index names.
E.g. you can create a table 
```sql
create table test_idx_table (
 id int primary key,
 time time
);

create index test_idx on test_idx_table(id, time)
```

without escaping the field name "time", but if we later read the index creation DDL using `pg_get_indexdef(pg_class_idx.oid) AS indexdef` it will have the escaped field name
```sql
CREATE INDEX test_idx ON marble.test_idx_table USING btree (id, "time")
```

### Josh's bug
Third commit
(see [slack discusion](https://marble-communitysiege.slack.com/archives/C06FRFB7V2M/p1744010732737959))
We previously modified our migrations so that they can be run with a postgres user other than the default superuser "postgres", but we missed one such migration. This fixes it. I also modified the postgres default user in the end to end test suite so we can detect this.

### DIFF
--- a/repositories/pg_indexes/pg_indexes.go
+++ b/repositories/pg_indexes/pg_indexes.go
@@ -8,7 +8,7 @@ import (
 	"github.com/checkmarble/marble-backend/pure_utils"
 )
 
-var columnNamesRegex = regexp.MustCompile(`(\([a-zA-Z0-9,_\ ]+\))`)
+var columnNamesRegex = regexp.MustCompile(`(\([\"a-zA-Z0-9,_\ ]+\))`)
 
 type PGIndex struct {
 	CreationInProgress bool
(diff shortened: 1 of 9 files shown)


repositories/pg_indexes/pg_indexes.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/checkmarble/marble-backend/pure_utils"
 )
 
-var columnNamesRegex = regexp.MustCompile(`(\([a-zA-Z0-9,_\ ]+\))`)
+var columnNamesRegex = regexp.MustCompile(`(\([\"a-zA-Z0-9,_\ ]+\))`)
```
apognu, line 11: Not specifically this line, but I would add the giant warning that if we ever change anything in those index, we should review the parsing logic. Especially, curious to know how it breaks if we start indexing expressions or use custom opclasses, which all require parentheses.

Not relevant now, but you know it will one day bite us.

apognu (changes requested):

repositories/pg_indexes/pg_indexes.go
Pascal-Delange: Yes, absolutely.
It would break.
Where would you ideally put the warning ?

apognu: Ideally next to the place where we build the command to create the indices. Arguably, this is what we'll change if we add fancy features to our index shapes.

Pascal-Delange: makes sense

apognu (approved):

PR #951, opened 2025-04-11
Add whitelist permissions to VIEWER role.

apognu:
The public API introduced a new permissions to manage sanction check whitelists, and this permission was only added to the `API_CLIENT` role, breaking whitelists from the UI.

This adds the permissions to the `VIEWER` role (the lowest that can update a case).


Pascal-Delange (commented):
Hmm, this is an interesting one.
Clearly this is the best solution we can do right now, but it raises some questions on how we should structure permission checks in the backend, for the future.
The _intention_ here was that the case usecase allows you to make anything that's logical within the context of the case, without relying on "global" permissions, something that is made more tricky when we inject another usecase that itself controls permissions to the case usecase (and it's only going to get more complex from there).

Seeing as the new public API can only be called with an API key not with an "end user" jwt (right?), this is ok as it actually does not allow a viewer to arbitrarily go and whitelist sanction checks - but can you please double check this for my peace of mind ?

apognu:
The public API prevents access with a user token, we get a 401 directly, so we should be good on that side.

This is something that I wanted to discuss this morning. Any "viewer", with this, can update cases and whitelist matches from the case UI. Is that what is expected? The name of the role told me "no", but could not find a more appropriate role (a viewer **can** update a case).

Pascal-Delange (approved):
👍

PR #954, opened 2025-04-11
Pascal/mar 896 order of db field options shown in the rule builder is

Pascal-Delange:
There was a logic error in the recursion of the data model exploration to return all possible DB field accesses.
Specifically, the second test case (added in `usecases/ast_expression_usecase_test.go`) was failing randomly, because it was only exploring links once by pair of "link name, target table", instead of once by pair of "link name, start table".

### DIFF
--- a/usecases/ast_expression_usecase_test.go
+++ b/usecases/ast_expression_usecase_test.go
@@ -46,7 +46,6 @@ func TestAstExpressionUsecase_getLinkedDatabaseIdentifiers(t *testing.T) {
 		},
 	}
 
-	u := AstExpressionUsecase{}
-	_, err := u.getLinkedDatabaseIdentifiers(scenario, model)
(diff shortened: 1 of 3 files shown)


usecases/ast_expression_usecase_test.go
```diff
@@ -46,7 +82,109 @@ func TestAstExpressionUsecase_getLinkedDatabaseIdentifiers(t *testing.T) {
...
+	}
+
+	identifiers, err := getLinkedDatabaseIdentifiers(scenario, model)
 	assert.NoError(t, err)
+
+	expectedStr := []string{
+		"transactions-[account]-id",
+		"transactions-[account]-company_id",
+		"transactions-[account company]-id",
+		"transactions-[company]-id",
```
Pascal-Delange, line 184: Without the changes to the getLinkedDatabaseIdentifiers in this PR, typically only one of those options would be returned, changing randomly on every run.

PR #955, opened 2025-04-11
fixes: make fuzzy match on aggregate work

Pascal-Delange:
Continued from @siiick 's PR. Making some changes to make it work, and to decide on the final design.

### Backend
(continuation of this PR, or the one it merges into)
- adapt list of algorithm names that are acceptable (won't be the same as for main ast fuzzy match)
- adapt proper score thresold (between 0 and 1 or 0 and 100?)
- finish writing proper logic for the final SQL query
### Frontend
- check that the examples of strings that we give, are still relevant for our final scoring logic
- adapt list of algorithm names that are acceptable (won't be the same as for main ast fuzzy match)
- adapt translations, view only component, and other visual stuff

### DIFF
--- a/usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
+++ b/usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
@@ -30,7 +30,8 @@ func (f FuzzyMatchOptionsEvaluator) Evaluate(ctx context.Context, arguments ast.
 		))
 	}
 
-	threshold, err := AdaptNamedArgument(arguments.NamedArgs, "threshold", promoteArgumentToFloat64)
+	threshold, err := AdaptNamedArgument(arguments.NamedArgs, "threshold",
+		promoteArgumentToFloat64) // Threshold is received as an integer (0–100) from the client and converted to a float (0.0–1.0) for internal processing.
 	if err != nil {
 		return nil, []error{err}
 	}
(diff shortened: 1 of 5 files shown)


usecases/ast_eval/evaluate/evaluate_fuzzy_match_options.go
```diff
@@ -49,7 +49,7 @@ func (f FuzzyMatchOptionsEvaluator) Evaluate(ctx context.Context, arguments ast.
 
 	return ast.FuzzyMatchOptions{
 		Algorithm: algorithm,
-		Threshold: threshold,
+		Threshold: threshold / 100,
```
Pascal-Delange, line 52: ok, so somewhere in the code we have to switch from manipulating numbers between 0 and 100 to numbers between 0 and 1 (at the latest at the repository level).
Unless we directly send numbers between 0 and 1 from the frontend, but then it's inconsistent with the other fuzzy matching modale which isn't great.
Doing it in the repository is probably risky because super prone to errors, if we make any changes on this. But doing it here feels a bit surprising as well.
My suggestion:
- leave a very visible comment next to the field in `ast.FuzzyMatchOptions` definition, that it expects a float between 0 and 1
- leave an inline comment next to where we parse the value (so next to `threshold, err := AdaptNamedArgument(arguments.NamedArgs, "threshold"...`) explaining why we receive a value between 0 and 100
- and perhaps optionally, as a failsafe keep the value as it is if a value between 0 and 1 is received in the first place (not sure about this one, let me know)

siiick: Thanks! I added the suggested comments here: https://github.com/checkmarble/marble-backend/pull/955/commits/00742314c0fc7412e1df4ba4072a8782a2c732d5.

Regarding the optional failsafe: since we’re not changing the contract with the frontend, I don’t see a scenario where we would receive a decimal value for the threshold. Let me know if I missed a possible case!

apognu (changes requested):

siiick (approved):

PR #957, opened 2025-04-14
Add endpoint to list paginated case decisions

apognu:


repositories/decisions_repository.go
```diff
@@ -253,6 +257,49 @@ func (repo *MarbleDbRepository) DecisionsByCaseId(
...
+		Limit(uint64(req.Limit) + 1).
+		OrderBy("created_at DESC, id DESC")
+
+	if req.CursorId != "" {
+		cursorDecision, err := repo.DecisionsById(ctx, exec, []string{req.CursorId})
+		if err != nil {
+			return nil, false, err
+		}
+		if len(cursorDecision) == 0 {
+			return nil, false, fmt.Errorf("could not find decision for cursor")
```
Pascal-Delange, line 281: wrap with models.NotFoundError ?

api/handle_cases.go
```diff
@@ -228,6 +228,40 @@ func handleUnsnoozeCase(uc usecases.Usecases) func(c *gin.Context) {
...
+			return
+		}
+
+		c.JSON(http.StatusOK, dto.CaseDecisionListDto{
+			Decisions: pure_utils.Map(decisions, func(d models.DecisionWithRuleExecutions) dto.DecisionWithRules {
+				return dto.NewDecisionWithRuleDto(d, nil, false)
+			}),
+			Pagination: dto.CaseDecisionListPaginationDto{
+				HasMore:      hasMore,
+				NextCursorId: decisions[len(decisions)-1].DecisionId,
```
Pascal-Delange, line 259: Beware, this could panic (if no decisions in the case). Unlikely to happen much in practice right now, but let's still avoid it.

api/handle_cases.go
```diff
@@ -228,6 +228,40 @@ func handleUnsnoozeCase(uc usecases.Usecases) func(c *gin.Context) {
 	}
 }
 
+func handleListCaseDecisions(uc usecases.Usecases) func(c *gin.Context) {
```
Pascal-Delange, line 231: Since we're reusing the same decision structs as dto as on other decision endpoints, but we're not actually joining cases to return data on cases, I think it'd be good to leave a comment (above the api handler probably is a good place) as a disclaimer

Pascal-Delange: Nb: also in the openapi in the frontend repo, as it is a DecisionDto contains a "case" which has several non-nullable values (as per the openapi file, so also in ts typings)

models/case.go
```diff
@@ -130,3 +130,12 @@ type CaseAssignementRequest struct {
 	CaseId     string
 	AssigneeId *UserId
 }
+
+const CaseDecisionsPerPage = 50
```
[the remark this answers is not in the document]
Pascal-Delange: 50 is a fine default I think, but given how many lines of code it is I'd still say we add the possibility to customize it ? (This way if we want to change it later, it's just a frontend code change)

api/handle_cases.go
apognu: It really couldn't, repository returns a NotFound error if no decisions were found, so we wouldn't even get here. But I get the point, let's not risk it, I'll change it.

In any case, maybe returning an error if there are no decisions is a bit harsh. I'll also change that.

api/handle_cases.go
apognu: I'm not sure I follow, the OpenAPI spec for this DTO defines those required fields:

```yaml
- id
- created_at
- trigger_object
- trigger_object_type
- outcome
- pivot_values
- scenario
- score
```

Which are all returned by this endpoint. I agree we do not have the data on the case (which I think is expected since the case is the input), so I don't see what I write here.

Can you clarify, if there is something I missed?

Pascal-Delange: Hmm yes my bad, I missed the fact that DecisionDto.case is not required.
nevermind then.

repositories/decisions_repository.go
```diff
@@ -253,6 +257,48 @@ func (repo *MarbleDbRepository) DecisionsByCaseId(
...
+		query = query.Where(squirrel.Expr("(created_at, id) < (?, ?)",
+			cursorDecision[0].CreatedAt, req.CursorId))
+	}
+
+	decisionsChan, errChan := repo.channelOfDecisions(ctx, exec, query)
+
+	decisions := ChanToSlice(decisionsChan)
+	err := <-errChan
+
+	if len(decisions) == 0 {
```
Pascal-Delange, line 293: we should return error if `err!=nil` before this early return

Pascal-Delange (commented):
last very minor point, otherwise good to go !

Pascal-Delange (approved):

PR #958, opened 2025-04-14
Get ID of the next pending or unassigned case.

apognu:
Next case is determined finding the next older case in the same inbox that is either unassigned or pending.


repositories/case_repository.go
```diff
@@ -471,3 +471,43 @@ func (repo *MarbleDbRepository) GetCasesWithPivotValue(ctx context.Context, exec
...
+	query := NewQueryBuilder().
+		Select("id").
+		From(dbmodels.TABLE_CASES).
+		Where(squirrel.And{
+			squirrel.Eq{
+				"org_id":      c.OrganizationId,
+				"inbox_id":    c.InboxId,
+				"assigned_to": nil,
+			},
+			squirrel.Expr("(created_at, id) < (?, ?)", c.CreatedAt, c.Id),
```
Pascal-Delange, line 489: I'm unsure about this logic, as it would exclude new cases that have been created (if any, and if there are no older ones left). Maybe just order by `boost is null, created_at desc, id desc limit 1` should be enough ?

apognu: That is a good question, exactly what do we want to do here?

* The **next** available case
* The **first** other available case
* A mix of the two where we would send the next if any, or the first if any, excluding the current one

Simply ordering by the key you gave would effectivlly render older cases invisible as long as new cases are created faster than we can process the current one.

If I were to define what I would expect, as a user, when cliking on "next", it's that it would literally give me the available case created before my current one, and if there aren't any, maybe get the first available overall. But you tell me.

Pascal-Delange: I'd say, go to the highest priority case in the inbox, meaning the oldest unassigned case, considering boosted ones first (maybe a maximalist position would be to order by boosting date DESC among those that are boosted, but can do without it). In any case I don't think the intent here is that it depends on "which case I'm currently on" - in my understanding it's more a "get me to my next case as an analyst" (user point of view), much more than "get me to the next case in current pagination order" (technical point of view)

apognu: I believe the current code fulfills what you wanted. I will merge and we can come back and amend the order after the fact.

PR #961, opened 2025-04-16
Differentiate between different types of tags.

apognu:
This PR adds a `target` column on the `tags` table indicating what resource the tag can be used with. Valid values are `case` and `object`.

It also modified all endpoints to take a `target` parameter.

All existing tags are marked as targeting cases.

### DIFF
--- a/repositories/migrations/20250416141100_create_tag_types.sql
+++ b/repositories/migrations/20250416141100_create_tag_types.sql
@@ -1,0 +1,10 @@
+-- +goose Up
+
+alter table tags
+    add column target text not null default 'case',
+    add constraint tag_kind_check check (target in ('case', 'object'));
+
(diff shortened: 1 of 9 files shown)


repositories/migrations/20250416141100_create_tag_types.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+
+alter table tags
+    add column target text not null default 'case',
+    add constraint tag_kind_check check (target in ('case', 'object'));
+
+update tags
+set target = 'case';
```
Pascal-Delange, line 8: not necessary because handled by the default value, no ? (not a big deal of course)

Pascal-Delange (approved):
👍

PR #962, opened 2025-04-16
Implemented case escalation.

apognu:
This PR allows the configuration of an escalation inbox on an inbox. That inbox will be the target of the escalation event that will move a case there, unassign and unsnooze it, and boosts it so it is marked as waiting for event.

It also implements the endpoints used to configure the target as well as the one to trigger the event.

It adds a case event related to escalation.

### DIFF
--- a/usecases/case_usecase.go
+++ b/usecases/case_usecase.go
@@ -61,6 +61,8 @@ type CaseUseCaseRepository interface {
 	BoostCase(ctx context.Context, exec repositories.Executor, id string, reason models.BoostReason) error
 	UnboostCase(ctx context.Context, exec repositories.Executor, id string) error
 
+	EscalateCase(ctx context.Context, exec repositories.Executor, id, inboxId string) error
+
 	GetCasesWithPivotValue(ctx context.Context, exec repositories.Executor,
 		orgId, pivotId, pivotValue string) ([]models.Case, error)
 }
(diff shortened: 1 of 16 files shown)


usecases/case_usecase.go
```diff
@@ -1400,3 +1401,64 @@ func (usecase *CaseUseCase) ReadCasePivotObjects(ctx context.Context, caseId str
...
+		return errors.Wrap(err, "could not read source inbox")
+	}
+	if sourceInbox.EscalationInboxId == nil {
+		return errors.Wrap(models.UnprocessableEntityError,
+			"the source inbox does not have escalation configured")
+	}
+
+	targetInbox, err := uc.inboxReader.GetInboxById(ctx, *sourceInbox.EscalationInboxId)
+	if err != nil {
+		return errors.Wrap(err, "could not read target inbox")
```
Pascal-Delange, line 1434: Here we'll need to be more specific on error handling, or maybe more easily to just read the inbox status through the inbox repository directly: inboxReader verifies permissions on the inbox, but it is not necessary to have access to the escalation inbox, to send a case to it (the default assumption would even be the opposite, though it will vary from client to client)

usecases/inboxes_usecase.go
```diff
@@ -76,7 +77,7 @@ func (usecase *InboxUsecase) CreateInbox(ctx context.Context, input models.Creat
 	return inbox, nil
 }
 
-func (usecase *InboxUsecase) UpdateInbox(ctx context.Context, inboxId, name string) (models.Inbox, error) {
+func (usecase *InboxUsecase) UpdateInbox(ctx context.Context, inboxId, name string, escalationInboxId *string) (models.Inbox, error) {
```
Pascal-Delange, line 80: So just to have it written black on white at least once in the PR:
- only org admin can set the escalation inbox target for a given inbox
- anyone can see the escalation inbox id, for an inbox they can see
- _if_ we want to display the name of the escalation inbox in the frontend (say, if we want to display "escalate case to inbox XXX..."), then we're missing a small thing because a non-org admin user won't be able to see the name of that inbox => to discuss with @carere @Arnaudschw. Nb: this is not blocking as per the designs on figma today, just wanted to point it out as long as it's fresh.

Pascal-Delange (commented):
Thanks, one small thing to resolve related to permissions

usecases/inboxes_usecase.go
Pascal-Delange: seems good to me 👍

Pascal-Delange (approved):

PR #966, opened 2025-04-17
chore: improve migrations by not making them create a postgres user

Pascal-Delange:
As discussed in slack: the user is not strictly required, in particular it's dropped since Nov 2024.
Worst case side effect of this migration would if if someone currently on a version of marble from between March 2024 and Nov 2024 upgrades to the most recent version, they would remain with an unused user "analytics" (without permissions on anything) => acceptable.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,6 @@
       "go.toolsEnvVars": {
             "GOFUMPT_SPLIT_LONG_LINES": "on"
       },
-      "go.lintTool": "golangci-lint",
       "go.lintOnSave": "package",
       "search.exclude": {
             "go.sum": true,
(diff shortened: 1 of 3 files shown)


apognu (approved):

.vscode/settings.json
```diff
@@ -15,14 +14,12 @@
...
             ".git/logs": true,
       },
-      "editor.tabSize": 4,
       "files.associations": {
             // This is in order to deactivate github copilot on the .env file. As a result syntax coloring will be disabled because
             // VSCode will treat it as plain text, you can choose your language mode manually on the file
             ".env": "env"
       },
       "github.copilot.enable": {
-            "*": true,
```
apognu, line 25: 🎉

PR #968, opened 2025-04-18
return ingested object tags along with pivot objects in a case

Pascal-Delange:
### Content

- [x] return annotations on reading pivot objects in case
- [x] return annotations on reading client objects as table

### Related
See openapi spec change in https://github.com/checkmarble/marble-frontend/pull/799

 ### Comments
I went ahead and implemented grouped by tag type in the backend, because I figured _someone_ would have to do it systematically when displaying them, so we might as well do it in the backend where it minimizes data manipulation. But you let me know if this is overengineered @ChibiBlasphem .

Nb: the `annotations` nested object is not required, and will be only present if any annotations are present.

New example of output payload:
```json
...
{
            "pivot_object_id": "account_id",
            "pivot_value": "account_id",
            "pivot_id": "038f72a2-2c98-47e6-ba85-28547075aafc",
            "pivot_type": "object",
            "pivot_object_name": "accounts",
            "pivot_field_name": "object_id",
            "is_ingested": true,
            "pivot_object_data": {
                "metadata": {
                    "valid_from": "2025-04-09T04:25:21.952291+08:00",
                    "object_type": "accounts"
                },
                "data": {
                    "account_id": null,
                    "company_id": "company_id",
                    "company_number": null,
                    "object_id": "account_id",
                    "updated_at": "2006-01-02T23:04:05.999999+08:00"
                },
                "related_objects": [],
                "annotations": {
                    "comments": [],
                    "tags": [
                        {
                            "id": "d878666a-53a7-489a-b38b-5b2f64e89a35",
                            "type": "tag",
                            "payload": {
                                "tag": "291ec3e1-66a0-47a4-98c8-f989873a586a"
                            },
                            "annotated_by": "0cd34ca6-de6e-440d-9460-8f2ae2824e37",
                            "created_at": "2025-04-18T14:39:43.903321+08:00"
                        }
                    ],
                    "files": []
                }
            },
            "number_of_decisions": 24
        },
... 
```

### DIFF
--- a/dto/client_data_object.go
+++ b/dto/client_data_object.go
@@ -86,17 +86,17 @@ type ClientObjectMetadata struct {
 }
 
 type ClientDataListResponse struct {
-	Data       []models.ClientObjectDetail `json:"data"`
-	Pagination ClientDataListPagination    `json:"pagination"`
+	Data       []ClientObjectDetail     `json:"data"`
+	Pagination ClientDataListPagination `json:"pagination"`
 }
 
 func (c ClientDataListResponse) MarshalJSON() ([]byte, error) {
 	if c.Data == nil {
-		c.Data = make([]models.ClientObjectDetail, 0)
+		c.Data = make([]ClientObjectDetail, 0)
 	}
 	return json.Marshal(struct {
-		Data       []models.ClientObjectDetail `json:"data"`
-		Pagination ClientDataListPagination    `json:"pagination"`
+		Data       []ClientObjectDetail     `json:"data"`
+		Pagination ClientDataListPagination `json:"pagination"`
 	}{
 		Data:       c.Data,
 		Pagination: c.Pagination,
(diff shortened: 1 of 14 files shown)


models/entity_annotation_payload.go
```diff
@@ -61,7 +61,7 @@ type EntityAnnotationFilePayloadFile struct {
 func (EntityAnnotationFilePayload) entityAnnotationPayload() {}
 
 type EntityAnnotationTagPayload struct {
-	Tag string `json:"tag" binding:"required"`
+	TagId string `json:"tag_id" binding:"required"`
```
Pascal-Delange, line 64: This one is a question of taste honestly - I don't mind rolling it back, just let me know

dto/client_data_object.go
```diff
@@ -2,22 +2,101 @@ package dto
...
+	}{
+		Metadata:       c.Metadata,
+		Data:           c.Data,
+		RelatedObjects: c.RelatedObjects,
+		Annotations:    c.Annotations,
+	})
+}
+
+type RelatedObject struct {
+	LinkName string             `json:"link_name"`
```
apognu, line 71: @ChibiBlasphem Regarding what we discussed yesterdfay, is this where you wanted to also have the target table name, and not only the link name?

Pascal-Delange: Hey,
That should already be present in the ClientObjectDetail.Metadata object

Pascal-Delange: It's not supposed to be undefined - if it is, it means I made an error. 
P.S. Ok I see that I didn't mark it as required in the openapi spec - let me double check the backend implem, in any case metadata should be present

ChibiBlasphem (approved):
LGTM

dto/client_data_object.go
Pascal-Delange: @ChibiBlasphem ok, here's how I'll modify the API: metadata will be required, but only the object type will be required on it (valid_from will be optional, only returned if the object actually was ingested).

Pascal-Delange: Which is already almost what is actually implemented in the backend, except it will return a "zero" value for valid_from instead of null

Pascal-Delange:
Last thing to do, check that the frontend (openapi) PR has exactly the correct schema (I'm mostly thinking of required/not required character of keys)

PR #969, opened 2025-04-22
Add data model table options.

apognu:


apognu:
@carere I don't know what you did, but the diff now contains changes from another commit from another branch...

carere:
@apognu  Just updated the branch by rebasing (all by just clicking on the GUI button "update branch" from Github), maybe the history was already broken ?

dto/data_model_dto.go
```diff
@@ -160,3 +160,17 @@ type CreateNavigationOptionInput struct {
 	FilterFieldId   string `json:"filter_field_id"`
 	OrderingFieldId string `json:"ordering_field_id"`
 }
+
+type UpdateDataModelOptionsInput struct {
+	DisplayedFields *[]string `json:"displayed_fields"`
```
Pascal-Delange, line 165: I always find pointers to slices a bit confusing, does it really serve a purpose here to use a *[]string instead of a []string ? IIUC the only difference is it will unmarshal to a nil pointer to a slice if the propery is not passed, instead of a nil slice

usecases/data_model_usecase.go
```diff
@@ -708,3 +708,70 @@ func (usecase *DataModelUseCase) CreateNavigationOption(ctx context.Context, inp
...
+		return models.DataModelOptions{}, err
+	}
+	if tableMeta.OrganizationID != orgId {
+		return models.DataModelOptions{}, errors.Wrap(models.NotFoundError, "table not found")
+	}
+
+	if req.DisplayedFields != nil && len(*req.DisplayedFields) > 0 {
+		for _, fieldId := range *req.DisplayedFields {
+			fieldMeta, err := uc.dataModelRepository.GetDataModelField(ctx, exec, fieldId)
+			if err != nil {
```
Pascal-Delange, line 760: I would say this is not the best way to do it, because it hides any error from the repository under a "not found error". How about just reading the full data model (using the usecase "getDataModel", or the same method on the repository) and checking presence of the fields in the returned table ?

Pascal-Delange (approved):
thanks,
I left a couple of small comments, but I'm approving already seeing as I won't be able to do so in the afternoon - feel free to make any last changes and merge 👍

Pascal-Delange (approved):

PR #971, opened 2025-04-24
fix: fuzzy match filter

siiick:


apognu:
See my [comments here](https://linear.app/marble-eu/issue/MAR-812/revamp-case-statuses-to-better-match-client-workflows#comment-aa438466) for the few issues we have.

apognu (approved):

PR #977, opened 2025-04-29
Change Yente healthcheck to /readyz

apognu:
An undocumented "feature" of Cloud Run is that they reserve the `/healthz` path for internal purpose, meaning that the backend service will never get requests destined to that endpoint.

As we use that endpoint on Yente to check if it is up, this failed constantly with a 404 error.

This PR moves our healthcheck to `/readyz`, which is not reserved and serves - more or less - the same purpose.


ChibiBlasphem (approved):
LGTM

siiick (approved):

PR #982, opened 2025-05-01
Pascal/mar 937 improve sentry error reporting backend

Pascal-Delange:
# Objective
- reduce spam on Sentry by grouping DB connection related issues
- make issues more easily readable
- add more context on issues (concerned user)

See [Sentry doc on fingerprinting](https://docs.sentry.io/concepts/data-management/event-grouping/fingerprint-rules/#combining-matchers)

## Content
- Rewrite events before sending to Sentry to have better issue titles in Sentry
- Add information on user/API key as user attribute or tags for display on Sentry (and for more precise "archive until it happens to another XX users" feature)

### Left for future work:
- add useful breadcrumbs to issues
- do something similar on the frontend project
    - use similary Sentry config on the frontend project to automatically ignore errors that are caused by a 500 on the backend ? => Ok I added this `error.value:"*HTTP Client Error with status code: 500*" -> api-client-error`

### Before:
Capture d’écran 2025-04-30 à 21 40 50
--- 
### After (see how the title is now "cannot scan ...")
Capture d’écran 2025-04-30 à 21 40 41

## Limitations
Because we use `UnwrapAll` to get the first error, in the case of errors generated by third party libraries, the errors we'll see on Sentry will only be as good as what they generate.

## Related change (non-code)
In Sentry, I set this fingerprinting behavior for errors on the backend project:
```
# group errors from creating a DB connection. Group them by the DB that is concerned.
error.type:"\*pgconn.ConnectError" -> db_connection_error, {{ error.type }}

# for local testing: group db connection errors, grouping them by URL/PORT
error.value:"*connect: connection refused*" -> db_connection_error, {{ error.type }}

# for cloudsql db: group db connection errors, effectively grouping them by unix socket (hence by db)
error.value:"*/cloudsql*" error.value:"*failed to receive message: unexpected EOF" -> db_connection_error, {{ error.type }}
```

If it does what I think it does (tested locally to some extent), it should group all errors related to DB connections, grouping them further by the DB that is concerned (so that client DBs will be grouped separately from the Marble DB and from each other)

### DIFF
--- a/infra/sentry.go
+++ b/infra/sentry.go
@@ -1,6 +1,9 @@
 package infra
 
-import "github.com/getsentry/sentry-go"
+import (
+	"github.com/cockroachdb/errors"
+	"github.com/getsentry/sentry-go"
+)
 
 func SetupSentry(dsn, env, apiVersion string) {
 	if err := sentry.Init(sentry.ClientOptions{
@@ -34,6 +37,10 @@ func SetupSentry(dsn, env, apiVersion string) {
 		BeforeSend: func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
 			if event.Request != nil {
 				event.Request.Headers["X-Api-Key"] = "[redacted]"
+			}
+			if hint != nil && event != nil && len(event.Exception) > 0 {
+				originalErr := errors.UnwrapAll(hint.OriginalException)
+				event.Exception[len(event.Exception)-1].Type = originalErr.Error()
 			}
 			return event
 		},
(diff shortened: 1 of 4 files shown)


infra/sentry.go
```diff
@@ -35,6 +38,10 @@ func SetupSentry(dsn, env, apiVersion string) {
 			if event.Request != nil {
 				event.Request.Headers["X-Api-Key"] = "[redacted]"
 			}
+			if hint != nil && event != nil && len(event.Exception) > 0 {
+				originalErr := errors.UnwrapAll(hint.OriginalException)
+				event.Exception[len(event.Exception)-1].Type = originalErr.Error()
```
Pascal-Delange, line 43: So, this is a bit hacky, since it effectively modifies the stack of errors received by Sentry. But It's the best I was able to find in order to modify the displayed title of the issue on Sentry (and what was suggested on different forums).
By default, Sentry takes the most recent error in the chain, which typically is not the most relevant one to use as a title.

siiick: Indeed, it is well aligned with the documentation 👍

apognu (approved):

PR #993, opened 2025-05-12
Decision filtering by trigger object_id.

apognu:


Pascal-Delange (commented):
Implem-wise, this most definitely does it (let's just hope that postgres isn't creative when selecting an index for this query...).

Let's just add it to the openapi spec and push it to readme please.
I suppose we still want to add this one on the pre-v1 public API, as we're not done migrating... or should we use the occasion to push people who want it to the v1 ?

apognu:
To be honest, if this internal endpoint is shared a public API one, I would vote to reimplement it in the V1 public API to encourage customer to use it.

Even though we are going to maintain the legacy public API, being tied to maintaining an endpoint that is used both internally and publically is dangerous.

So if you agree, I can reimplement this with the new standards on the public API tomorrow.

Pascal-Delange:
> So if you agree, I can reimplement this with the new standards on the public API tomorrow.

I agree and I had the same reasons in mind, go for it :)

PR #995, opened 2025-05-12
Various case manager improvements

apognu:
* Change default case ordering as specified by MAR-959.
* Centralize case action side effects (unboosting and self-assigning).

### DIFF
--- a/repositories/migrations/20250512144400_change_case_index.sql
+++ b/repositories/migrations/20250512144400_change_case_index.sql
@@ -1,0 +1,23 @@
+-- +goose Up
+-- +goose StatementBegin
+
+alter index idx_inbox_cases rename to idx_inbox_cases_2;
+
+create index idx_inbox_cases
+    on cases (org_id, inbox_id, (boost is null), (assigned_to is not null), created_at desc, id desc);
+
+drop index idx_inbox_cases_2;
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+alter index idx_inbox_cases rename to idx_inbox_cases_2;
+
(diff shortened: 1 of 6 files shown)


repositories/migrations/20250512144400_change_case_index.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+
+alter index idx_inbox_cases rename to idx_inbox_cases_2;
+
+create index idx_inbox_cases
+    on cases (org_id, inbox_id, (boost is null), (assigned_to is not null), status, created_at desc, id desc);
```
Pascal-Delange, line 6: The order on status does not seem to be the same as the one you use in the condition above.
Btw, considering the amount of business logic that is now getting into this, I think we're that 👌 close to maybe wanting to make a generated column out of it ,

repositories/migrations/20250512144400_change_case_index.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+
+alter index idx_inbox_cases rename to idx_inbox_cases_2;
+
+create index idx_inbox_cases
```
Pascal-Delange, line 5: concurrently 🙏

Pascal-Delange: well, or keep it blocking (it should be a _fairly_ fast one) but then wrap it in a goose statement so it's idempotent in case of errors - I'm a bit uneasy with the rename not being idempotent (we had errors before with some customers who had two migrations involving indexes running in //, so that one of them errored out)

Pascal-Delange (approved):
👍  thanks!

PR #997, opened 2025-05-13
feat(case-manager): make GetNextCase relative to the provided case

siiick:


repositories/case_repository.go
```diff
@@ -504,6 +504,8 @@ func (repo *MarbleDbRepository) GetNextCase(ctx context.Context, exec Executor,
 				"assigned_to": nil,
 			},
 			squirrel.NotEq{"status": "closed"},
+			squirrel.NotEq{"id": c.Id},
+			squirrel.GtOrEq{"created_at": c.CreatedAt},
```
apognu, line 508: As outlined verbally, this would fail if we are already on the most recent unassigned case, we basically reversed the initial issue.

Pascal-Delange (commented):
We rediscussed this this morning (?) with Arnaud and Antoine (sorry, was orally in the office)
The short version is (and I should write it up):
- the order/filters for this should be: boosted before non boosted, ignore assigned closed and snoozed, otherwise oldest first
- if the current case is not in this category (unassigned open and not snoozed), navigate to the highest (in order) such case
- if the case is already in this category, navigate to the next one in order (if there is none, move back to inbox homepage ?)

where "in order" above means (boosted, created_at ASC) as mentioned. It would probably be most easily done with two sql queries and some code logic to glue them, than just one query on the full order.

siiick:
I hardly see the need to build two distinct queries.
In both cases, we’re looking for a case that is:
 - unassigned
 - open
 - not snoozed
 - in the same org/inbox as the current one

Additionally, we always want to prioritize cases that are:
 - boosted
 - oldest first

If I understand your comment correctly, we can keep the same base query (after my changes: [48413031](https://github.com/checkmarble/marble-backend/pull/997/commits/484130310029b4d691f03761fbf1d04b57a34062) and [ff08c982](https://github.com/checkmarble/marble-backend/pull/997/commits/ff08c98235b71414d8720df7c1829d6bdc612f54)).

If the current case is unassigned, open, and not snoozed, we can just filter for cases created after it, and keep the same ordering logic.

Does that sound good @Pascal-Delange ?

repositories/case_repository.go
```diff
@@ -505,11 +505,24 @@ func (repo *MarbleDbRepository) GetNextCase(ctx context.Context, exec Executor,
...
+			},
 			squirrel.NotEq{"status": "closed"},
 		}).
-		OrderBy("boost is null", "created_at", "id").
+		OrderBy("boost is not null", "created_at", "id").
 		Limit(1)
 
+	if c.AssignedTo == nil && c.Status != "closed" && (c.SnoozedUntil == nil || c.SnoozedUntil.Before(time.Now())) {
+		query = query.Where(
+			squirrel.And{
```
Pascal-Delange, line 519: As discussed orally earlier: this "pagination" like condition does not exactly match the order that we defined above - also, the index that we created is on "boost is null" (which is similar enough to "boost is not null" but the DB cannot translate) => would be better to replace by "boost is null desc" perhaps. In any case, let me just a moment to review this

Pascal-Delange:
@siiick @apognu I'm going to close this one for now. There _might_ still be something wrong with what we implemented, but it's defocus and we have not had enough feedback on the use of this to spend energy on it right now IMO.

PR #999, opened 2025-05-14
Improved logging for decisions.

apognu:
- Decision storage in now measured and logged
 - Each decision, on top of logging the overall evaluation duration now also logs:
   - Each macro-step in the decision process (trigger expression, rules execution, sanctions check, name recognition)
   - Each rule individual duration
 - Add a `LOG_LEVEL` knob (set to `info` or `debug`) to control how verbose JSON and GCP adapters are

### DIFF
--- a/usecases/decision_usecase.go
+++ b/usecases/decision_usecase.go
@@ -527,12 +527,15 @@ func (usecase *DecisionUsecase) CreateDecision(
 		scenarioExecution.ExecutionMetrics.Steps[evaluate_scenario.LogStorageDurationKey] = storageDuration.Milliseconds()
 
 		utils.LoggerFromContext(ctx).InfoContext(ctx,
-			fmt.Sprintf("created decision %s in %dms", decision.DecisionId,
-				decisionDuration.Milliseconds()), "org_id", scenario.OrganizationId, "decision_id",
-			decision.DecisionId, "scenario_id", scenario.Id, "score", scenarioExecution.Score,
-			"outcome", scenarioExecution.Outcome, "duration", decisionDuration.Milliseconds(),
-			"rules", scenarioExecution.ExecutionMetrics.Rules, "steps",
-			scenarioExecution.ExecutionMetrics.Steps)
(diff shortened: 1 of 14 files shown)


apognu:
Maybe this is the time to decode if some of our current `INFO` entries could be set as `DEBUG` instead.

usecases/decision_usecase.go
```diff
@@ -517,6 +520,22 @@ func (usecase *DecisionUsecase) CreateDecision(
...
+	if scenarioExecution.ExecutionMetrics != nil {
+		storageDuration := time.Since(storageStart)
+		decisionDuration := time.Since(decisionStart)
+
+		scenarioExecution.ExecutionMetrics.Steps[evaluate_scenario.LogStorageDurationKey] = storageDuration.Milliseconds()
+
+		utils.LoggerFromContext(ctx).InfoContext(ctx,
+			fmt.Sprintf("created decision %s in %dms", decision.DecisionId,
+				decisionDuration.Milliseconds()), "org_id", scenario.OrganizationId, "decision_id",
+			decision.DecisionId, "scenario_id", scenario.Id, "score", scenarioExecution.Score,
```
Pascal-Delange, line 532: purely for my personal OCD, I'd love one line per key/value :D

Pascal-Delange (commented):
nice 👍 
I do think it would be great to tune down the highest volume logs to debug, and this PR would be a great place to do so - mostly I'm thinking of the following ones:
- all those logs about rule execution and scenario execution
- ingestion logs about "XX objects ingested"
- "send webhook" logs
- "do test run decision" logs

But really the easiest way to find them would be to look at the logs explorer and filter all stdout logs on the backend.

usecases/decision_usecase.go
apognu: Oh thank God, me too. I didn't want to upset the auto-formatter, but perfect.

apognu:
Yes. I also started looking at production logs to see what we could cut on. Glad we agree, I'll append a commit transitioning some of those infos into debugs. 💰

Pascal-Delange (approved):

PR #1000, opened 2025-05-14
fix error on returned database accessors

Pascal-Delange:
## Content

Two bugs became visible with Swan's data model:
- some database accessors that should be available in the builder, would be returned only randomly (e.g. sometimes you'd get `transactions.accounts.account_holder.aml_score`, sometimes you'd get `transactions.card.accounts.account_holder.aml_score` instead, etc)
- sometimes, the backend would return database accessors that are plainly not valid - like `transactions.card.accounts.account_holder.project.aml_score`. This was due to some pointer confusion.

The diff is more test cases than actual code change.
As you may guess by reading it, it took more time to find the issue than to fix it (perfect storm of recursion and slice pointer spaghetti....).

## Kinda breaking change
Previously, the database accessors that were returned allowed loops (accessors going through the same table several times), as long as any given link was walked only once (to avoid infinite loops).
With this change, loops are no longer possible (any table can be present only once in any given access path).
I think this is ok - I don't know of any clients paid or not that we know of that designed a loopy data model, and the use case is a bit far fetched. _If_ anyone really complains, I'm willing to reopen this, but generally speaking I think it's better we don't push in this direction anyway.

### DIFF
--- a/usecases/ast_expression_usecase.go
+++ b/usecases/ast_expression_usecase.go
@@ -76,7 +76,10 @@ func getLinkedDatabaseIdentifiers(scenario models.Scenario, dataModel models.Dat
 				continue
 			}
 			visited = append(visited, relation)
-			pathForLink := append(path, linkName)
+
+			// deepcopy so that different identifiers don't collide
+			pathForLink := append(make([]string, 0, len(path)+1), path...)
+			pathForLink = append(pathForLink, linkName)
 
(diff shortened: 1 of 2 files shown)


apognu (approved):

apognu:
Btw, I'm somehow disappointed your hoarded PR #1000 for such a boring issue, but I'm willing to look past that. ;)

PR #1001, opened 2025-05-15
Created a /config endpoint.

apognu:
This endpoint will rule them all, it merges:

 - /version
 - /is-sso-available
 - /signup-status

And adds more data for the backend to be able to send the frontend its configurationon the fly.

---

The goal, ultimately, is to offload the frontend from configuration, as much as possible for directives not strictly related to the frontend (or secret ones). We will take advantage that there is constant communication between the two to provide essential parameters through the new `/config` endpoint.

There are several advantages:

 - It becomes impossible to have configuration drift between the backend and frontend
 - Configuration is more centralized and partly less duplicated
 - Relaunching the backend is usually sufficient to perform configuration changes

This PR also tries to determine a few sane defaults for some of those pesky directives that trip up a lot of our users, initially.

 - `FIREBASE_PROJECT_ID` defaults to `GOOGLE_CLOUD_PROJECT`.
 - `FIREBASE_AUTH_DOMAIN` now defaults to the emulator host when using the emulator, or to `${PROJECT_ID}.firebaseapp.com` otherwise, which is most cases renders the directives moot.

There might others we want to to introduce.

---

This is a breaking change!

Some of the configuration directives should now be added to the backend environment. The backend can start and function without them, but will transmit empty values to the frontend. Therefore, we must, on the frontend, introduce a transition period where the local environment is still observed if the backend does not return values.

Note that even with this, it still might be a breaking change in this case:

 - The frontend had a correct environment variable
 - The backend had the same, unused, but invalid directive that the user added when trying to set up Marble

In this case, the frontend would prioritize the setting returned by the backend and use the wrong value.

Should we reverse the priority? Use local environment if present, and fall back to the `/config`-provided one? If we do that, we need to request the users to change their config with a clear deprecation deadline, otherwise we will be stuck there.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -113,6 +113,10 @@ LOGGING_FORMAT=text
 
 # Uncomment this line if you are using the Firebase emulator for testing.
 # FIREBASE_AUTH_EMULATOR_HOST="localhost:9099"
+# Set your Firebase settings as described in your app
+FIREBASE_PROJECT_ID=
+FIREBASE_API_KEY=
+FIREBASE_AUTH_DOMAIN=
 
 # If running the worker process in an environment requiring all processes to respond on an HTTP port, uncomment this setting.
 # For example, Cloud Run requires all services to process liveness checks to keep running.
--- a/.env.example
+++ b/.env.example
@@ -41,8 +41,8 @@ CREATE_ORG_ADMIN_EMAIL=jbe@zorg.com
 # Configure the ID of your Firebase project for authentication and the path to the service account's JSON private key file.
 #  - The 'Project ID' can be found in the 'General' section of your Firebase project's settings page.
 #  - The private key must be generated in the 'Service accounts' tab by clicking 'Generate new private key'
-# If using the Firebase emulator, this should match the emulator's project ID.
-GOOGLE_CLOUD_PROJECT="firebase-project-id"
+# If using the Firebase emulator, this should match the emulator's project ID. In a real deployment, it should be auto-detected from your credentials.
+# GOOGLE_CLOUD_PROJECT=test-project
 GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/private/key.json
 
 # Configure the document blob storage backend.
@@ -113,10 +113,13 @@ LOGGING_FORMAT=text
 
 # Uncomment this line if you are using the Firebase emulator for testing.
 # FIREBASE_AUTH_EMULATOR_HOST="localhost:9099"
+
 # Set your Firebase settings as described in your app
-FIREBASE_PROJECT_ID=
-FIREBASE_API_KEY=
-FIREBASE_AUTH_DOMAIN=
+FIREBASE_API_KEY=dummy
+# For SSO, if your auth domain is different from <project_id>.firebaseapp.com, define it here.
+# FIREBASE_AUTH_DOMAIN=
+# If your Firebase project differs from the global Google Cloud project you use, you can set it here.
+# FIREBASE_PROJECT_ID=test-project
 
 # If running the worker process in an environment requiring all processes to respond on an HTTP port, uncomment this setting.
 # For example, Cloud Run requires all services to process liveness checks to keep running.
--- a/api/configuration.go
+++ b/api/configuration.go
@@ -1,6 +1,10 @@
 package api
 
-import "time"
+import (
+	"time"
+
+	"github.com/checkmarble/marble-backend/infra"
+)
 
 type Configuration struct {
 	Env                 string
@@ -15,4 +19,14 @@ type Configuration struct {
 	BatchTimeout        time.Duration
 	DecisionTimeout     time.Duration
 	DefaultTimeout      time.Duration
+
+	FirebaseConfig FirebaseConfig
+	MetabaseConfig infra.MetabaseConfiguration
+}
+
+type FirebaseConfig struct {
+	EmulatorUrl string
+	ProjectId   string
+	ApiKey      string
+	AuthDomain  string
 }
--- a/api/configuration.go
+++ b/api/configuration.go
@@ -25,8 +25,12 @@ type Configuration struct {
 }
 
 type FirebaseConfig struct {
-	EmulatorUrl string
-	ProjectId   string
-	ApiKey      string
-	AuthDomain  string
+	EmulatorHost string
+	ProjectId    string
+	ApiKey       string
+	AuthDomain   string
+}
+
+func (cfg FirebaseConfig) IsEmulator() bool {
+	return cfg.EmulatorHost != ""
 }
--- a/api/configuration.go
+++ b/api/configuration.go
@@ -10,6 +10,7 @@ type Configuration struct {
 	Env                 string
 	AppName             string
 	Port                string
+	MarbleApiUrl        string
 	MarbleAppUrl        string
 	MarbleBackofficeUrl string
 	RequestLoggingLevel string
--- a/api/routes.go
+++ b/api/routes.go
@@ -39,6 +39,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 	r.GET("/version", tom, handleVersion(uc))
 	r.POST("/token", tom, tokenHandler.GenerateToken)
 	r.GET("/validate-license/*license_key", tom, handleValidateLicense(uc))
+	r.GET("/config", tom, handleGetConfig(uc, conf))
 	r.GET("/is-sso-available", tom, handleIsSSOEnabled(uc))
 	r.GET("/signup-status", tom, handleSignupStatus(uc))
 
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"net/http"
 	"os/signal"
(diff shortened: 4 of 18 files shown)


cmd/server.go
```diff
@@ -36,6 +36,21 @@ func RunServer(config CompiledConfig) error {
...
+			JwtSigningKey:       []byte(utils.GetEnv("METABASE_JWT_SIGNING_KEY", "")),
+			TokenLifetimeMinute: utils.GetEnv("METABASE_TOKEN_LIFETIME_MINUTE", 10),
+			Resources: map[models.EmbeddingType]int{
+				models.GlobalDashboard: utils.GetEnv("METABASE_GLOBAL_DASHBOARD_ID", 0),
+			},
+		},
+		FirebaseConfig: api.FirebaseConfig{
+			EmulatorUrl: utils.GetEnv("FIREBASE_AUTH_EMULATOR_HOST", ""),
+			ApiKey:      utils.GetEnv("FIREBASE_API_KEY", ""),
+			AuthDomain:  utils.GetEnv("FIREBASE_AUTH_DOMAIN", ""),
```
apognu, line 51: I think we could safely defaults that to `<project>.firebaseapp.com` if no emulator, or the emulator domain otherwise. That could also remove a setting for the usual case.

api/routes.go
```diff
@@ -38,6 +38,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 	r.GET("/version", tom, handleVersion(uc))
 	r.POST("/token", tom, tokenHandler.GenerateToken)
 	r.GET("/validate-license/*license_key", tom, handleValidateLicense(uc))
+	r.GET("/config", tom, handleGetConfig(uc, conf))
 	r.GET("/is-sso-available", tom, handleIsSSOEnabled(uc))
 	r.GET("/signup-status", tom, handleSignupStatus(uc))
```
apognu, line 43: Ultimately, the new endpoint would deprecate `/version`, `/is-sso-available` and `/signup-status`.

Pascal-Delange:
```
Should we reverse the priority? Use local environment if present, and fall back to the /config-provided one? If we do that, we need to request the users to change their config with a clear deprecation deadline, otherwise we will be stuck there.
```
I think we should, because we want to do this anyway. Warn customers that we make the config deprecated and that it will removed 6 months from now, add a warning log every time the frontend is launched with them, and a last big release note warning in the release where we actually remove them.

api/configuration.go
```diff
@@ -15,4 +19,14 @@ type Configuration struct {
 	BatchTimeout        time.Duration
 	DecisionTimeout     time.Duration
 	DefaultTimeout      time.Duration
+
+	FirebaseConfig FirebaseConfig
+	MetabaseConfig infra.MetabaseConfiguration
+}
+
+type FirebaseConfig struct {
+	EmulatorUrl string
```
Pascal-Delange, line 28: small naming point, the relevant env var is called `FIREBASE_AUTH_EMULATOR_HOST`, and it indeed expects a hostname (typically localhost:9099) not a full url

cmd/server.go
Pascal-Delange: in fact, do you see any case currently where it could be different from that ?

cmd/server.go
```diff
@@ -119,6 +127,25 @@ func RunServer(config CompiledConfig) error {
 	marbleJwtSigningKey := infra.ReadParseOrGenerateSigningKey(ctx, serverConfig.jwtSigningKey, serverConfig.jwtSigningKeyFile)
 	license := infra.VerifyLicense(licenseConfig)
 
+	if apiConfig.FirebaseConfig.ProjectId == "" {
```
Pascal-Delange, line 130: In fact, I think that currently the app would not work if the frontend received another project id (as firebase project id) than the backend is using when creating the firebase client, which is created from default application credentials (see infra/firebase_client.go).

In fact, the GOOGLE_CLOUD_PROJECT in the backend is used only to start the tracing client, I don't think the firebase client uses it.

So probably, a reasonable thing to do would be to just get the project id - from a combination of default application credentials and the file pointed by GOOGLE_APPLICATION_CREDENTIALS ? - and send that.

api/configuration.go
Pascal-Delange: (translation to a url is already done in the frontend)

ChibiBlasphem (commented):
Is it wanted that value that are not defined in the config return a empty string ? I have for example "auth.firebase.api_key" or "urls.marble" which are empty string, which makes the frontend validation harder.

cmd/server.go
apognu: If that's the case, even better, we lose an extra configuration directive.

apognu:
> Is it wanted that value that are not defined in the config return a empty string ? I have for example "auth.firebase.api_key" or "urls.marble" which are empty string, which makes the frontend validation harder.

@ChibiBlasphem You did not say what you would be better for "validation", so it is hard to answer.

apognu:
@Pascal-Delange The commit I just pushed is still a work in progress, but I extracted what I did for my keyless signing branch to determine the project dynamically. I'll tune it out tomorrow.

ChibiBlasphem:
> > Is it wanted that value that are not defined in the config return a empty string ? I have for example "auth.firebase.api_key" or "urls.marble" which are empty string, which makes the frontend validation harder.
> 
> @ChibiBlasphem You did not say what you would be better for "validation", so it is hard to answer.

@apognu Yes, I should have been explicit on that, my bad.

.env.example
```diff
@@ -42,7 +42,6 @@ CREATE_ORG_ADMIN_EMAIL=jbe@zorg.com
 #  - The 'Project ID' can be found in the 'General' section of your Firebase project's settings page.
 #  - The private key must be generated in the 'Service accounts' tab by clicking 'Generate new private key'
 # If using the Firebase emulator, this should match the emulator's project ID.
-GOOGLE_CLOUD_PROJECT="firebase-project-id"
```
apognu, line 45: I removed both `GOOGLE_CLOUD_PROJECT` and `FIREBASE_PROJECT_ID` since they should be autodetected most of the time. Should we still add them, commented out, so people can override them in the few cases where it cannot be autodetected?

As a reminder, those cases are:

 - If the ADC is of a user account (we currently require a private key file for blob signing, so it should not happen).
 - If the user uses several Google Cloud projects (separate for Firebase, production and the service account).

Pascal-Delange: Yes let's definitely keep then in here (even if commented) - it's the only logical place we have to document what those do and what is expected.
And I still think GOOGLE_CLOUD_PROJECT will be needed if you run with the emulator ?

cmd/server.go
```diff
@@ -114,11 +130,31 @@ func RunServer(config CompiledConfig) error {
...
+
+		apiConfig.FirebaseConfig.ProjectId = gcpConfig.ProjectId
+	}
+	if apiConfig.FirebaseConfig.ApiKey == "" {
+		logger.Warn("no FIREBASE_API_KEY specified, this will be an error in the future")
+	}
+	if apiConfig.FirebaseConfig.AuthDomain == "" {
+		if apiConfig.FirebaseConfig.EmulatorHost != "" {
+			apiConfig.FirebaseConfig.AuthDomain = apiConfig.FirebaseConfig.EmulatorHost
+			logger.Warn("no FIREBASE_AUTH_DOMAIN specified, defaulting to FIREBASE_AUTH_EMULATOR_HOST",
```
Pascal-Delange, line 149: actually, I don't think the frontend needs the AuthDomain (or indeed any other param other than FIREBASE_AUTH_EMULATOR_HOST) if the emulator is used. So we could just ignore it if emulator is used

cmd/server.go
```diff
@@ -114,11 +130,31 @@ func RunServer(config CompiledConfig) error {
...
+				"auth_domain", apiConfig.FirebaseConfig.EmulatorHost)
+		} else {
+			apiConfig.FirebaseConfig.AuthDomain = fmt.Sprintf("%s.firebaseapp.com", apiConfig.FirebaseConfig.ProjectId)
+			logger.Warn(fmt.Sprintf("no FIREBASE_AUTH_DOMAIN specified, defaulting to %s", apiConfig.FirebaseConfig.AuthDomain))
+		}
+		logger.Warn("no FIREBASE_API_KEY specified, this will be an error in the future")
```
Pascal-Delange, line 155: and this one should go in the "else" block above where emulator is not used

Pascal-Delange: actually - since it's the same for the log at line 144 - it might just be best to first have a big block for the "use emulator" case (where all we need to set additionally is the project id), and then a block for the "normal firebase" case ?

cmd/server.go
apognu: The auth domain is required for SSO, it is the "authorize URL" (in OIDC terms) that the frontend redirects to for authentication.

cmd/server.go
apognu: The second warning was indeed a bad copy and paste. I reorganized to separate emulator/no-emualator. And forced the auth domain when using the emulator.

cmd/server.go
apognu: Plus, we need to domain to whitelist it in the Content Security Policy (otherwise the frame will not be able to communicate).

cmd/server.go
apognu: Better?

Pascal-Delange (approved):
🙏  great ! 
I'd just like to be sure that we test this thoroughly before it goes to prod - let's perhaps merge it after the next release, hopefully early next week ?

PR #1010, opened 2025-05-18
[WIP] improve indexes to query decisions

Pascal-Delange:


apognu:
> also, vacuum full only rewrites the table in the same "shape", it does not move things to toast table if the setting has changed :'(

As a note, `vacuum full` would also _actually_ drop the dropped columns, instead of simply marking them as invisible, which would help with row packing and therefore loads. We still cannot do it on prod, though. :)

Pascal-Delange:
Finished in this PR:
- simplification of the queries for decisions, cases without returned rank
- BIG cleanup of duplicated, confusing, code in the decisions repository (esp. the infamous "sql to channel to slice of models" stuff
- re-normalization of decisions.scenario version, name & desc (columns to be removed in a later PR)
- tweaking of some VACUUM parameters for large tables
- dropping of two FKs and the indexes needed for their cascade on decisions

Planned in the scope of the task but will be done in a future PR:
- documenting indexes on decisions and why we need them
- actually act on my draft plan of removing some indexes / creating new ones

repositories/migrations/20250905163100_various_schema_improvements.sql
```diff
@@ -0,0 +1,120 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE decisions
```
apognu, line 3: This is mostly a style / ease-to-read note, but I would merge settings that applies on the same table into a single statement. Knowing what applies to which table is a bit hard to figure out when they are all separate and spread out.

```
ALTER TABLE webhook_events
SET (
  autovacuum_analyze_scale_factor = 0.01,
  autovacuum_vacuum_insert_scale_factor = 0.02
);
```

The same applies to the `drop`s.

PR #1012, opened 2025-05-20
feat(cases): rework case ordering

siiick:
- Remove ordering based on assignment status
- Boosted cases are now prioritized depending on the pagination order

### DIFF
--- a/repositories/case_repository.go
+++ b/repositories/case_repository.go
@@ -477,7 +477,7 @@ func (repo *MarbleDbRepository) GetCasesWithPivotValue(ctx context.Context, exec
 }
 
 func orderConditionForCases(p models.PaginationAndSorting) string {
-	return fmt.Sprintf("c.boost is null, c.assigned_to is not null, c.%s %s, c.id %s", p.Sorting, p.Order, p.Order)
+	return fmt.Sprintf("c.boost is null %s, c.%s %s, c.id %s", p.Order, p.Sorting, p.Order, p.Order)
 }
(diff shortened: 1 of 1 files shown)


Pascal-Delange (approved):

repositories/case_repository.go
```diff
@@ -477,7 +477,7 @@ func (repo *MarbleDbRepository) GetCasesWithPivotValue(ctx context.Context, exec
 }
 
 func orderConditionForCases(p models.PaginationAndSorting) string {
-	return fmt.Sprintf("c.boost is null, c.assigned_to is not null, c.%s %s, c.id %s", p.Sorting, p.Order, p.Order)
+	return fmt.Sprintf("c.boost is null %s, c.%s %s, c.id %s", p.Order, p.Sorting, p.Order, p.Order)
```
apognu, line 480: Nothing to do with your PR specifically, but @Pascal-Delange , at some point, we may consider using indexed placeholders for some of those (`%[1]s`, ...) instead of repeating the same param over and over again.

This one is reasonable, but I remember seeing one with the same parameter more than ten times.

apognu (approved):

repositories/case_repository.go
Pascal-Delange: I know exactly which one you're talking about - the fuzzy match aggregate filter.
TBH, I just never realized this form of syntax existed, and I always missed it from python.
Let's go for it.

PR #1014, opened 2025-05-21
Add all datasets, tagged, to the list of datasets

apognu:
~This adds a new section when configuring screening for a scenario including all lists pertaining to politically exposed people.~

This PR:

 - Removes the restriction on _which_ datasets we consider for showing in the app (before, we only displayed `sanctions` datasets
 - We add a new field to the dataset spec we return, `tag`, which represents what kind of category the dataset is. The list of possible tags for now is:
   - `adverse-media`
   - `third-parties`
   - `peps`
   - `sanctions`

To note, quite a few dataset are not classified into one of those superset datasets, and therefore will not have a tag.

The OpenSanctions / Yente endpoint we use also changes to target `/match/default`, instead of `/match/sanctions`.

Finally, to prepare for !1023 (that will separate label name recognition config and mangle them to be a `StringConcat`, the DTO will now wrap a plain payload into a `StringConcat`.

### DIFF
--- a/repositories/httpmodels/http_opensanctions_dataset.go
+++ b/repositories/httpmodels/http_opensanctions_dataset.go
@@ -1,6 +1,7 @@
 package httpmodels
 
 import (
+	"fmt"
 	"maps"
 	"slices"
 	"strings"
@@ -24,6 +25,7 @@ var (
 		"South America":  "sa",
 		"United Nations": "un",
 		"Other":          "other",
+		"Internal":       "internal",
 	}
 )
 
@@ -42,7 +44,7 @@ type HTTPOpenSanctionCatalogDataset struct {
 func AdaptOpenSanctionCatalog(datasets []HTTPOpenSanctionCatalogDataset) models.OpenSanctionsCatalog {
 	sections := make(map[string]*models.OpenSanctionsCatalogSection, len(OPEN_SANCTIONS_CONTINENT_CODES))
 	datasetMap := make(map[string]*HTTPOpenSanctionCatalogDataset, len(datasets))
-	loadedDatasets := make(map[string]set.Set[string])
+	loadedDatasets := make(map[string]*set.Set[string])
 
 	slices.SortFunc(datasets, func(lhs, rhs HTTPOpenSanctionCatalogDataset) int {
 		return strings.Compare(lhs.Title, rhs.Title)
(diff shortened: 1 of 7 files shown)


repositories/httpmodels/http_opensanctions_dataset.go
```diff
@@ -99,6 +110,12 @@ func AdaptOpenSanctionCatalog(datasets []HTTPOpenSanctionCatalogDataset) models.
 		return *section
 	}
 	sortF := func(lhs, rhs models.OpenSanctionsCatalogSection) int {
+		if lhs.Name == OPEN_SANCTIONS_PEPS_DATASET {
```
apognu, line 113: Order TBD (we might even show it completely separately, which makes this moot).

apognu:
Need to go over it again to be aligned with the new design.

repositories/httpmodels/http_opensanctions_dataset.go
apognu: Not relevant anymore.

repositories/httpmodels/http_opensanctions_dataset.go
```diff
@@ -113,6 +130,25 @@ func AdaptOpenSanctionCatalog(datasets []HTTPOpenSanctionCatalogDataset) models.
 	}
 }
 
+func findPepsDatasets(sections map[string]*models.OpenSanctionsCatalogSection, datasets []HTTPOpenSanctionCatalogDataset, peps []string) {
+	for _, dataset := range datasets {
+		if slices.Contains(peps, dataset.Name) {
+			if _, ok := sections[OPEN_SANCTIONS_PEPS_DATASET]; !ok {
+				sections[OPEN_SANCTIONS_PEPS_DATASET] = &models.OpenSanctionsCatalogSection{
+					Name:     OPEN_SANCTIONS_PEPS_DATASET,
+					Title:    "Politically Exposed People",
```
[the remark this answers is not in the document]
apognu: Not relevant anymore.

apognu:
We will probably not merge it now. It would work (it is backward compatible), but there would **a lot** of datasets listed without tags, so maybe not what we want. But it's reviewable.

Pascal-Delange (commented):
I remember that you set up a custom list in a local instance, could you please double check what it shows up like - and if we can configure it make it come up as "internal" ?
(Otherwise good to merge!)

PR #1020, opened 2025-05-23
Reorder and add more case action side effects.

apognu:
This touches on the side effects performed on a case when an action is taken, namely:

 - The case going to "Investingating" if it is pending is now a side effect.
 - Trigger side effect on rule snooze and suspicious activity report actions.
 - Fix side effects on case snoozing and unsnoozing (since one of the side effect is to unboost, we lost the snooze effect).
 - Generalize side effects when assigning a case.

Also, remove the 409 if trying to unsnooze a case that is not snoozed.

### DIFF
--- a/repositories/case_repository.go
+++ b/repositories/case_repository.go
@@ -203,6 +203,7 @@ func (repo *MarbleDbRepository) UnsnoozeCase(ctx context.Context, exec Executor,
 	sql := NewQueryBuilder().
 		Update(dbmodels.TABLE_CASES).
 		Set("snoozed_until", nil).
+		Set("boost", nil).
 		Where(squirrel.Eq{"id": caseId})
 
 	return ExecBuilder(ctx, exec, sql)
--- a/usecases/case_usecase.go
+++ b/usecases/case_usecase.go
@@ -465,11 +465,11 @@ func (uc *CaseUseCase) Snooze(ctx context.Context, req models.CaseSnoozeRequest)
 			previousSnooze = utils.Ptr(c.SnoozedUntil.Format(time.RFC3339))
 		}
 
-		if err := uc.repository.BoostCase(ctx, tx, req.CaseId, models.BoostUnsnoozed); err != nil {
+		// Case side effects should be called before snoozing, since it removes the boost.
+		if err := uc.PerformCaseActionSideEffects(ctx, tx, c); err != nil {
 			return err
 		}
-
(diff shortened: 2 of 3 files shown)


repositories/case_repository.go
```diff
@@ -203,6 +203,7 @@ func (repo *MarbleDbRepository) UnsnoozeCase(ctx context.Context, exec Executor,
 	sql := NewQueryBuilder().
 		Update(dbmodels.TABLE_CASES).
 		Set("snoozed_until", nil).
+		Set("boost", models.BoostUnsnoozed).
```
apognu, line 206: Blanket boosting of the case when manually unsnoozing, not sure if this is what we want.

usecases/case_usecase.go
```diff
@@ -465,11 +465,10 @@ func (uc *CaseUseCase) Snooze(ctx context.Context, req models.CaseSnoozeRequest)
 			previousSnooze = utils.Ptr(c.SnoozedUntil.Format(time.RFC3339))
 		}
 
-		if err := uc.repository.BoostCase(ctx, tx, req.CaseId, models.BoostUnsnoozed); err != nil {
```
apognu, line 468: This reorders the side effects to happen before the boost, since one of the side effects is unboosting. A bit brittle, should I pass an option saying if unboost should happen?

repositories/case_repository.go
apognu: > we'd want to boost (preemtively) on snooze and unboost on unsnooze ?

The first part is already what we do, but yeah, my question was mostly about the second case. If we consider that a manual unsnoozing is not natural unsnoozing, it would make sense to change this and remove the boost in that case.

My thought, initially, was that a case that was snoozed suddenly is "urgent" enough to warrant its immediate unsnoozing, so I would boost it.

Pascal-Delange: Hmm yea honestly I don't know. On one hand that makes sense. On the other hand you're already on the case when you do this.
So no strong opinion.

apognu: I can remove it for now.

My thought was in the case someone else unsnoozes one of your cases, so it would pop up in your list. Might be a very niche use case or just me being paranoid. :D

Pascal-Delange: yep let's do that.
otherwise PR is good to go

Pascal-Delange (approved):

PR #1021, opened 2025-05-26
Replace migration moving pg_trgm to check if we are extension owner.

apognu:
Moving the extension to be housed in a specific schema requires that the current user be the owner of the extension. This causes issue on platform where the managed database precreates the extension with their administrative roles.

This checks whether the current user is the owner, and only moves the extension if so.


Pascal-Delange (approved):
👍

PR #1025, opened 2025-06-02
Case events for annotations

apognu:
This PR creates a link between an entity annotation and the case it was added from (if applicable):

 * Any annotation created will also create a case event of type `entity_annotated`.
 * A file annotation will attach the file to the case.

It also fixes an issue with the client object viewer where the models would not be serialized.

### DIFF
--- a/api/handle_client_data.go
+++ b/api/handle_client_data.go
@@ -34,7 +34,12 @@ func handleGetIngestedObject(uc usecases.Usecases) func(c *gin.Context) {
 			return
 		}
 
-		c.JSON(http.StatusOK, objects[0])
+		output, err := dto.AdaptClientObjectDetailDto(objects[0])
+		if presentError(ctx, c, err) {
+			return
+		}
+
+		c.JSON(http.StatusOK, output)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


api/handle_client_data.go
```diff
@@ -34,7 +34,12 @@ func handleGetIngestedObject(uc usecases.Usecases) func(c *gin.Context) {
 			return
 		}
 
-		c.JSON(http.StatusOK, objects[0])
+		output, err := dto.AdaptClientObjectDetailDto(objects[0])
```
Pascal-Delange, line 37: thanks - I missed this one when I split the clientObjectDetail model into separate models and dto

usecases/entity_annotations_usecase.go
```diff
@@ -61,6 +69,19 @@ func (uc EntityAnnotationUsecase) List(ctx context.Context, req models.EntityAnn
...
+		OrgId:          orgId,
+		AnnotationId:   id,
+		IncludeDeleted: true,
+	})
+
+	if err != nil {
+		return models.EntityAnnotation{}, err
+	}
+
+	return annotations[0], nil
```
Pascal-Delange, line 83: we need to handle the empty case here, since the repository uses the "list of models" helper - and return the NotFoundErr here (this could panic)

Pascal-Delange (approved):

PR #1028, opened 2025-06-04
WIP use uuid for inbox ids

Pascal-Delange:
PR generated with [Jules](https://jules.google/) 👀  - well, the first 25% anyway

--- 

## Context 

So, I experimented with Jules (Google's beta autonomous coding agent). I thought I'd try if I can use it efficiently to make it perform a relatively straightforward (no ambiguous business logic knowledge or interpretation required) task that is annoying to do manually.

Concretely, I tried the migration of string=>uuid for all our models that contain "inbox ids" (the DB already contains uuids AKA arrays of 16 bytes - it's essentially a matter of replacing "string" by "uuid" in a lot of places.

My prompt was something along the lines of 
```
This backend service uses "string" instead of "uuid" types for the "inbox" related models everywhere, replace it so we get better type safety.
```

FYI the way it works is, I need to connect it to our repository, which it clones in a VM, runs a first analysis of the code to propose intermediate resolution steps, you validate them, then it runs them and returns the result to you (display in real time or pushes to a branch).

It will run tests, try to compile etc as best as it can to validate is job.

## Debrief
It didn't go quite as well as I supposed it would.
Mainly two things I wasn't super happy about:
- for some reason, it seems to have stopped at about half the scope of the changes to do. Not sure why. (that means that there remained a layer of parsing uuid/casting to string at the interface between both "parts" of the codebase)
- it left ugly comments that made no sense whatsoever outside the context of the PR, in a lot of places
- it displayed sort of bad taste in how it adapted my unit tests (but that's ok - they were probably not the best unit tests to begin with and I can live with that

One more thing that was a bit annoying (at least I was a bit disappointed it didn't understand it better) was that it failed to run the tests with `go test ./...` because of our integration test (permission denied on the VM to pull the docker image of postgres) - the error makes sense, that it tried 15 times despite a relatively clear error message less so.

There was also one slightly less trivial part on correctly parsing between null and absent values for a uuid input, but it "smartly" avoided that part of the migration altogether (see pure_utils/nullable.go).

In the end I continued the work where the bot had started, and it still took me non-negligible time to finish it cleanly. I might have been more efficient by just running the same thing again with a better prompt, who knows it's part of experimenting.

## Next steps
It's still a good first attempt of starting to migrate from strings everywhere to uuids everywhere. I'm still cautiously optimistic that we can make this work at a larger scale for this kind of "dumb" migrations, given more cautious prompting perhaps.

Do not hesitate to try it out if you see some reasonable subjects to test it on.

--- 

## Issues

Some problems with binding uuid type to URI, query params https://github.com/gin-gonic/gin/issues/3060. Still WIP.

### DIFF
--- a/.github/workflows/backend_deploy_workflow.yaml
+++ b/.github/workflows/backend_deploy_workflow.yaml
@@ -61,7 +61,7 @@ jobs:
             echo "✅ Image with commit hash ${{ github.sha }} already exists, will reuse it"
             echo "should_build=false" >> "$GITHUB_OUTPUT"
           else
-            echo "❌ No image found for commit hash ${{ github.sha }}, build a new one"
+            echo "No image found for commit hash ${{ github.sha }}, build a new one"
             echo "should_build=true" >> "$GITHUB_OUTPUT"
           fi
 
@@ -86,8 +86,6 @@ jobs:
 
       - name: Set up Cloud SDK
         uses: "google-github-actions/setup-gcloud@v2"
-        with:
-          install_components: beta
 
       - name: Deploy migration job
         run: |


pure_utils/nullables.go
```diff
@@ -0,0 +1,53 @@
...
+type NullUUID struct {
+	UUID  uuid.UUID
+	Valid bool // Valid is true if UUID is not NULL
+	Set   bool // Set is true if the value was present in JSON, even if it was null
+}
+
+func (u *NullUUID) UnmarshalJSON(data []byte) error {
+	u.Set = true // Set to true if the value was present in JSON
+
+	if string(data) == "null" {
```
apognu, line 19: I don't think that we need the performance gain here (especially considering the loss in legibility, but all JSON value types can be determined by their first byte, so it would be `null` if the first byte is `n`.

Not that we need to change it, but a bit of trivia.

repositories/inbox_users_repository.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/repositories/dbmodels"
+	"github.com/google/uuid" // Ensure this import is present
```
apognu, line 8: What about that comment?

usecases/case_usecase.go
```diff
@@ -114,16 +113,12 @@ func (usecase *CaseUseCase) ListCases(
 	ctx context.Context,
 	organizationId string,
 	pagination models.PaginationAndSorting,
-	filters dto.CaseFilters,
+	filters models.CaseFilters,
```
apognu, line 116: Is the change of package wanted?

usecases/case_usecase.go
```diff
@@ -114,16 +113,12 @@ func (usecase *CaseUseCase) ListCases(
...
 ) (models.CaseListPage, error) {
 	if !filters.StartDate.IsZero() && !filters.EndDate.IsZero() &&
 		filters.StartDate.After(filters.EndDate) {
 		return models.CaseListPage{}, fmt.Errorf("start date must be before end date: %w", models.BadParameterError)
 	}
-	statuses, err := models.ValidateCaseStatuses(filters.Statuses)
```
apognu, line 123: Same remark with that. It that related to the UUID change at all? Or is a change of convenience or a side effect?

apognu:
For argument's sake, and after trying this morning, we do have a way to automatically deserialize into a type that wrap a `uuid.UUID` through gin's `BindUnmarshaler`. Consider the following handler:

```go
package api

import (
	"net/http"

	"github.com/gin-gonic/gin"
	"github.com/google/uuid"
)

type playgroundInput struct {
	FirstId  UriUuid `uri:"firstId"`
	SecondId UriUuid `uri:"secondId"`
}

type UriUuid struct {
	_uuid uuid.UUID
}

func (u UriUuid) Uuid() uuid.UUID {
	return u._uuid
}

func (u *UriUuid) UnmarshalParam(s string) error {
	parsed, err := uuid.Parse(s)
	if err != nil {
		return err
	}

	*u = UriUuid{
		_uuid: parsed,
	}

	return nil
}

func handlePlayground() gin.HandlerFunc {
	return func(c *gin.Context) {
		ctx := c.Request.Context()

		var params playgroundInput

		if err := c.ShouldBindUri(&params); err != nil {
			presentError(ctx, c, err)
			return
		}

		c.JSON(http.StatusOK, gin.H{
			"firstId":  params.FirstId.Uuid(),
			"secondId": params.SecondId.Uuid(),
		})
	}
}
```

This adds one level of indirection (because we need to extract the inner UUID from our wrapper struct), but arguably we only need to do it once, since we will usually pass the UUID itself to the usecase.

apognu (approved):

apognu (changes requested):
Sorry, I misclicked the approval. I still think we should iterate over this

pure_utils/nullables.go
Pascal-Delange: It's funny that you'd mention it, because I stumbled upon the same piece of trivia while looking up how best to do this.
I think I like the clarity of writing it out in full, but good to know !

usecases/case_usecase.go
Pascal-Delange: Yep this one is, because it used to be the usecase received a instance of "dto" filters, did some validation, and translated it to a "models" filter.
In this PR, I moved the validation part of the "dto" filters to the API handler level - validating statuses (that answers your other comment just below), and string to uuid for inbox_ids (I really tried to get something to work for "slice of UUID query params with binding", but didn't manage to)

api/handle_inboxes.go
```diff
@@ -152,17 +165,23 @@ func handlePatchInbox(uc usecases.Usecases) func(c *gin.Context) {
 			return
 		}
 
+		inboxId, err := uuid.Parse(getInboxInput.InboxId)
```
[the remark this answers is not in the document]
Pascal-Delange: done in a new commit

dto/case_dto.go
```diff
@@ -102,6 +104,19 @@ type CreateCaseCommentBody struct {
 	Comment string `json:"comment" binding:"required"`
 }
 
+// type UnmarshallingUuid struct {
```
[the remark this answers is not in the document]
Pascal-Delange: as discussed - we kept it, I moved it to models though

Pascal-Delange:
@apognu  I iterated on this according to what we discussed, should be good now

PR #1029, opened 2025-06-04
basic setup for per-user AI assist config

Pascal-Delange:
I accidentally ended up doing some refactoring in the same PR.
Mostly I merged methods in the "OrganizationRepository" "UsersRepository" concrete classes into the MarbleDbRepository.

### DIFF
--- a/api/handle_organization.go
+++ b/api/handle_organization.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
@@ -9,6 +10,7 @@ import (
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/pure_utils"
 	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/checkmarble/marble-backend/utils"
 )
 
 func handleGetOrganizations(uc usecases.Usecases) func(c *gin.Context) {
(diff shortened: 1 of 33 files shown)


api/handle_organization.go
```diff
@@ -110,10 +112,24 @@ func handleDeleteOrganization(uc usecases.Usecases) func(c *gin.Context) {
...
 		ctx := c.Request.Context()
-		organizationID := c.Param("organization_id")
+
+		creds, found := utils.CredentialsFromCtx(ctx)
+		if !found {
+			presentError(ctx, c, fmt.Errorf("no credentials in context"))
+			return
+		}
+
+		organizationIdFromPath := c.Param("organization_id")
```
Pascal-Delange, line 122: permissions where incorrectly configured, one could pass any org id in the path and get this org's feature access.

models/organization_feature_access.go
```diff
@@ -117,5 +123,9 @@ func (f DbStoredOrganizationFeatureAccess) MergeWithLicenseEntitlement(
 		o.Sanctions = MissingConfiguration
 	}
 
+	if user != nil && *&user.AiAssistEnabled {
```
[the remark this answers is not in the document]
Pascal-Delange: Oups that's a typo, thanks

usecases/feature_access/feature_access_reader.go
```diff
@@ -53,17 +54,27 @@ func NewFeatureAccessReader(
 func (f FeatureAccessReader) GetOrganizationFeatureAccess(
 	ctx context.Context,
 	organizationId string,
+	userId *models.UserId,
```
apognu, line 57: I'm okay with the PR as it is, but I think ultimately we should generalize what I started with the security enforcers being able to return the current org and user instead of passing them everywhere.

EDIT: this would not be applicable for methods that need to be called from the backoffice and therefore would need different org/used than those of the credentials.

apognu (approved):

usecases/feature_access/feature_access_reader.go
Pascal-Delange: I see - in this precise case, what would you suggest ? Here I need it more for business logic than for permission enforcement. 
Do you have in mind to pass the methods to retrieve orgId/userId to the "contextful" usecase ?

apognu: Initially, that was intent, yes. We have usecases that are created in the context of a credentials (org/user), I feel it would be ideal that we retrieve those from it, since they **MUST** be there for the usecase to be conceptually valid in the first place.

This is a larger refactoring, so I would not vote that we add it in this PR, though.

PR #1037, opened 2025-06-09
Consider all entity referents when checking against whitelists.

apognu:
We used to only check for whitelisted match against the main entity ID at the time of the screening. This completely shadowed that an entity can have its ID changed over its lifetime.

This PR now considers all previous "referents" IDs when checking whether the match was whitelisted or not.


usecases/sanction_check_usecase.go
```diff
@@ -448,6 +452,10 @@ func (uc SanctionCheckUsecase) FilterOutWhitelistedMatches(ctx context.Context,
 	sanctionCheck.Matches = matchesAfterWhitelisting
 	sanctionCheck.Count = len(sanctionCheck.Matches)
 
+	if sanctionCheck.Count == 0 {
```
apognu, line 455: I don't know if this was on purpose or an omission, but it seemed that if all matches were whitelisted, the status of the screening kept its iniital value.

It seems appropriate to change it to `no_hit` if we go down to zero match.

Pascal-Delange (approved):

PR #1038, opened 2025-06-09
PR suggestions/comments

Pascal-Delange:
I did a full (well, as best as I could) read of the code related to sanction checks in VScode, to focus on parts that were difficult to access in the github diff.
I thought the easiest way to point out my comments, and especially to write my comments in context, at this point was to open a PR against yours - feel free to reuse the branch or to just continue you work on your branch and close this one.

### DIFF
--- a/usecases/case_usecase.go
+++ b/usecases/case_usecase.go
@@ -1199,7 +1199,9 @@ func (usecase *CaseUseCase) CreateCaseFiles(ctx context.Context, input models.Cr
 	return usecase.getCaseWithDetails(ctx, exec, input.CaseId)
 }
 
-func (uc *CaseUseCase) AttachAnnotation(ctx context.Context, tx repositories.Transaction, annotationId string, annotationReq models.CreateEntityAnnotationRequest) error {
+func (uc *CaseUseCase) AttachAnnotation(ctx context.Context, tx repositories.Transaction,
+	annotationId string, annotationReq models.CreateEntityAnnotationRequest,
+) error {
 	return uc.repository.CreateCaseEvent(ctx, tx, models.CreateCaseEventAttributes{
 		CaseId:         *annotationReq.CaseId,
 		UserId:         (*string)(annotationReq.AnnotatedBy),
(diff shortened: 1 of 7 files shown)


usecases/case_usecase.go
```diff
@@ -1401,6 +1403,7 @@ func (usecase *CaseUseCase) ReviewCaseDecisions(
 	}
 
 	if input.ReviewStatus == models.ReviewStatusApprove {
+		// GetActiveSanctionCheckForDecision expets a sanction check id last, not a decision id. (also, would be great to add named params to the interface definition of it above)
```
apognu, line 1406: Actually, now that I read this again, this shouldn't use this method, should it? We need to iterate over **all** screenings for this decision and bail if one of them is still pending.

usecases/evaluate_scenario/preprocessing.go
```diff
@@ -22,7 +22,10 @@ type ScreeningPreprocessor func(
 	scc models.SanctionCheckConfig,
 ) ([]models.OpenSanctionsCheckQuery, error)
 
-func SkipIfUnder(ctx context.Context, e ScenarioEvaluator, queries []models.OpenSanctionsCheckQuery, iteration models.ScenarioIteration, scc models.SanctionCheckConfig) ([]models.OpenSanctionsCheckQuery, error) {
+func SkipIfUnder(ctx context.Context, e ScenarioEvaluator, queries []models.OpenSanctionsCheckQuery,
+	iteration models.ScenarioIteration, scc models.SanctionCheckConfig,
+) ([]models.OpenSanctionsCheckQuery, error) {
+	// nitpicking, but it's not "really" an edge case
```
apognu, line 28: Not sure I understand the comment. This is to skip when in the "disabled" state (which is `0`, since it doesn't really make sense to skip things with 0 length).

usecases/scenarios/scenario_validation.go
```diff
@@ -207,6 +210,9 @@ func (self *ValidateScenarioIterationImpl) Validate(ctx context.Context,
 
 			result.SanctionCheck[idx].Query = queryValidation
```
apognu, line 211: There is also a letover above of only checking for the `name` field, which will need to be generic now that we accept other fields.

usecases/case_usecase.go
Pascal-Delange: Hmm yes, that is certainly a more correct way to put it

Pascal-Delange:
cool, yes no problem the branch wasn't necessarily supposed to be kept long term

PR #1041, opened 2025-06-11
return information on "can be annotaged" status of a client object

Pascal-Delange:
Incremental changes to the API consumed by the case manager, to facilitate some data display by the frontend following QA of object annotations.


ChibiBlasphem (approved):
LGTM

apognu:
Looks good, but asked something I am wondering about.

PR #1043, opened 2025-06-12
Pascal/simplifiy ci

Pascal-Delange:
- Commit 1: remove deprecated cloud run jobs from deployment step (then, will also remove them from terraform)
- Commit 2: try an improvement where we tag images with the commit hash, and reuse existing images for prod deployments if available


Pascal-Delange:
@apognu  agreed, I pushed a new commit, let me know if it looks better for you.
New flow:
- look up image by sha, build it if not found
- always tag the image with its version tag (latest or git tag depending on staging or prod)

PR #1044, opened 2025-06-12
rework dtos and add related cases data in export

Pascal-Delange:
- improves zip export by writing all to a writer without intermediate bytes
- rework dtos for agent data export
- add exported data on related cases, ingested client data as csv

### DIFF
--- a/usecases/ai_agent_usecase.go
+++ b/usecases/ai_agent_usecase.go
@@ -6,9 +6,9 @@ import (
 	"encoding/json"
 	"io"
 
-	"github.com/checkmarble/marble-backend/dto"
 	"github.com/checkmarble/marble-backend/dto/agent_dto"
 	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
 	"github.com/checkmarble/marble-backend/repositories"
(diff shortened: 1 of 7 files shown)


usecases/ai_agent_usecase.go
```diff
@@ -207,6 +213,32 @@ func (uc AiAgentUsecase) GetCaseDataZip(ctx context.Context, caseId string) (io.
...
+			return
+		}
+
+		for pivotObjectStr, data := range relatedDataPerClient {
+			pivotObjectFolder := fmt.Sprintf("related_data/%s/", pivotObjectStr)
+			_, err := zipw.Create(pivotObjectFolder)
+			if err != nil {
+				pw.CloseWithError(errors.Wrapf(err, "could not create %s folder in zip", pivotObjectStr))
+				return
+			}
```
apognu, line 229: If I remember correctly, you don't need to explicitly create the directories in which you put files when creating a ZIP. Just create files with a path prefix and they will be placed in the appropriate folders lazily.

apognu (changes requested):

usecases/ai_agent_usecase.go
Pascal-Delange: that's good to know, thanks !

PR #1047, opened 2025-06-13
Case action side effect would set status to investigating on close.

apognu:
One of the side effect of most case actions was to set the case to investigating if it was pending. This was incompatible with one action: closing a case while it was still pending.


Pascal-Delange (commented):
It pains me a bit to have this large reaching change in the function signature if it's going to be used everywhere, just for the edge case of ignoreStatus.
Would it make sense for you to keep the old "perform side effect" method, and just use a slightly different on in the "UpdateCase" method ? 
(Less DRY, but simpler signatures and less weird edge conditions)

apognu:
@Pascal-Delange Something along those lines?

PR #1060, opened 2025-06-20
Case review AI agent pipeline

Pascal-Delange:
# High level summary
### Input data
The agent takes some input data, namely the following (all in json format unless specified otherwise):
- data model of the org
- case base metadata
- case events
- Marble-input description of the organization's activity
- definition of all the rules that are used in decisions present in the case
- decision result with rule execution details for every decision in the case
- for every "pivot object" present in the case (should be just one in most cases):
   - previous cases for the entity (without rule exec details)
   - data from the tables where we can navigate to from the pivot object entity, as in the case manager, formatted in csv format:
      - all the related rows/columns for entities that have fewer than 100 rows
      - 500 most recent entries for the 15 most relevant/useful columns (as determined by a prompt given data samples) for entities that yield more than 100 rows

### Pipeline
The data is pipelined through a relatively simple DAG of prompts run on GCP's Vertex AI. The output of a prompt is used as input for the next prompt in the DAG (or as output).
The last-but-one prompt generates the actual case review, while the last one is a sanity check against hallucinations and other thing we want to protect against.

Capture d’écran 2025-07-01 à 09 35 40

### Google search input
Vertex AI models handle google search to get contextual data autonomously (without needed to do it "in code"), which I activated for the "rules definition review" and the "case review prompt".

## Output
The endpoint returns:
```json
{
    "ok": true,
    "output": "the case review goes here",
    "sanity_check": "comment from the sanity check goes here, if ok=false",
    "thought": "model thought output goes here, if there is any (depends on the model that is used in particular)"
}
```

### Permissions
Uses the same permissions (user-based) as the previously created endpoint that returns a .zip summary of a case

### Billing attribution
All LLM calls can be assigned to the context org with a label that is automatically set

### Prompts
In the future, for self-hosted user we will probably want to include the prompts used by the agent in the code. I'm thinking of compiling them into the binary, without necessarily having them hard coded in the backend repo. Probably still version controlled in a private git repo very soon. For the first iterations, just experiment with a (maybe version controlled) GCS bucket. To be discussed.

In any case, the set of (pretty basic prompts) used in this technical demonstrator are found [here](https://www.notion.so/checkmarble/Prompts-21fd1b7bc7c480518007f31e609d35f8) in notion.

### Libraries
This POC entirely uses Google's [genai](https://pkg.go.dev/google.golang.org/genai@v1.12.0) SDK for interacting with models. It can be used in "service account" mode or in "use my API key" mode, though this PR only does the former.
I found it fairly easy to use and work with and it's rather well documented.
As far as I was able to find, there is no "universal adapter" SDK that works. Go's adapted [langchain](https://pkg.go.dev/github.com/tmc/langchaingo) library is still quite a bit below the python version's expressivity.

### Other LLM-calling details
There are two places where I force an output schema for the response. It's fairly easy to setup. If we want to do "tool calling" in models, it's only a slightly more complex syntax and it should be easy to add when we want to.

## Dependencies
- a folder `prompts/case_review` that contains prompt templates in markdown. Must be present at runtime. The prompt **_should_** use the keys that are passed to the template in the go code. The prompt _**must not**_ use any template keys that are not passed by the go code.
   - optionally, a set of files `prompts/org_desc/{orgId}.md` that contain any raw text information on the organization's activity
- a file `prompts/ai_agent_models.json` that configures which model is used for which prompt (and the default fallback)
- all the files above are read at request runtime. The idea is to load them as a volume backed by GCS (to be created with terraform in a follow-up PR). We will want to put a cache on it probably, later. 

Also,
- `VERTEX_AI_GCP_REGION` env var (default to `global`) that selects which region models are run in. Beware that model availability is region-dependent, e.g. the fast and cheap preview `gemini-2.5-flash-lite-preview-06-17` is only available in `global`. 
- permission for the service account executing the backend to use vertex AI (need to set it up on terraform, I've been testing with my end user credentials so far)

# Main side changes
- introduce a new package agent_dto (nested in dto): we need some more focused DTOs to share data with the agent, that focus on text fields and less on internal ids etc.
    - some of the dto adapters also require a callback to get data on related objects, this made it easier to collect all the relevant data to make a proper dto that doesn't just consist of a bunch of ids
- introduce an optional (variadic) param to the method that lists client objects to select fields that we want
- add option to read data histograms together with the data model (rather than just in the new endpoint that returns statistics on fields)

# Next up
We'll see how we refactor this in the future to make the configuration easier - if and when we need to.
I'm afraid I'll try to squash before mergin so I can more easily cherry-pick for a prod hotfix, but after the review.

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -226,6 +226,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 	router.POST("/cases/:case_id/escalate", tom, handleEscalateCase(uc))
 
 	router.GET("/cases/:case_id/data_for_investigation", timeoutMiddleware(conf.BatchTimeout), handleGetCaseDataForCopilot(uc))
+	router.POST("cases/:case_id/review", timeoutMiddleware(conf.BatchTimeout), handleCreateCaseReview(uc))
 
 	router.GET("/inboxes/:inbox_id", tom, handleGetInboxById(uc))
 	router.GET("/inboxes/:inbox_id/metadata", tom, handleGetInboxMetadataById(uc))
--- a/pure_utils/maps.go
+++ b/pure_utils/maps.go
@@ -1,0 +1,11 @@
+package pure_utils
+
+func Keys[K comparable, V any](m map[K]V) []K {
+	keys := make([]K, len(m))
+	i := 0
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	return keys
+}
--- a/usecases/ai_agent_usecase.go
+++ b/usecases/ai_agent_usecase.go
@@ -116,7 +116,7 @@ func (uc *AiAgentUsecase) GetClient(ctx context.Context) (*genai.Client, error) 
 }
 
 func (uc *AiAgentUsecase) GetCaseDataZip(ctx context.Context, caseId string) (io.Reader, error) {
-	caseDtos, relatedDataPerClient, err := uc.getCaseData(ctx, caseId)
+	caseDtos, relatedDataPerClient, err := uc.getCaseDataWithPermissions(ctx, caseId)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get case data")
 	}
@@ -215,16 +215,16 @@ func (uc *AiAgentUsecase) generateContent(
 	data map[string]any,
 	generateContentConfig *genai.GenerateContentConfig,
 	previousContents ...*genai.Content,
-) (output string, previous []*genai.Content, err error) {
+) (out GenerateContentResult, err error) {
 	prompt, err := readPrompt(promptPath)
 	if err != nil {
-		return "", nil, err
+		return out, err
 	}
 
 	// Load model configuration on each call
 	modelConfig, err := models.LoadAiAgentModelConfig("prompts/ai_agent_models.json")
 	if err != nil {
-		return "", nil, errors.Wrap(err, "could not load AI agent model configuration")
+		return out, errors.Wrap(err, "could not load AI agent model configuration")
 	}
 
 	// Get the appropriate model for this prompt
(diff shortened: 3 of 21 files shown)


api/routes.go
```diff
@@ -49,7 +49,8 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 		}
 
 		pubapi.InitPublicApi()
-		pubapiv1.Routes(cfg, r.Group("/v1beta"), auth.AuthedBy(utils.PublicApiKey, utils.BearerToken), uc)
+		pubapiv1.Routes(cfg, r.Group("/v1beta"),
```
apognu, line 52: Can we find what exactly does those? Not that they're all bad, but some of those reformating are quite unreadable (not this one, but it happens).

Pascal-Delange: That would be `GOFUMPT_SPLIT_LONG_LINES` in .vscode/settings.json.
I'm open to remove it, it's just that on the other hand I also find it inconvenient when lines become really too long (I understand 80 chars is outdated, but also think > 120 or 130 is too much).

usecases/ai_agent_usecase.go
```diff
@@ -81,36 +89,313 @@ func NewAiAgentUsecase(
...
+	for k, v := range data {
+		b, err := json.Marshal(v)
+		if err != nil {
+			return "", errors.Wrapf(err, "could not marshal %s", k)
+		}
+		marshalledMap[k] = string(b)
+	}
+
+	t := template.Must(template.New(promptPath).Parse(prompt))
+	buf := bytes.Buffer{}
```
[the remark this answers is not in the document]
Pascal-Delange: Ah yes, I didn't notice the Reset() method.
However I also just refactored the part using the buffer into its own function, and passing it a shared buffer to avoid reallocations is probably too early an optimization (in any case, there are still quite a few other things to refactor better in this PR)

usecases/ai_agent_usecase.go
```diff
@@ -200,22 +203,25 @@ func readPrompt(path string) (string, error) {
 	return string(promptBytes), nil
 }
 
+// Mutates the input (variadic)previousContents by appending the new input content and any genereated output, and returns them for convenience.
```
apognu, line 206: Mmm, just in case, a function cannot change the length or capacity of a passed-in slice. The calling function will not have its array pointer updated and will not see new items (it would work, though, with your all-time favorite pointer to slice, which cannot be done with variadic parameters).

Returning the new slice it the Right Thing to do, but the comment is a bit misleading (but it's not convenience, it's the only way to affect to calls site.

Pascal-Delange: Oh thanks, that's a good point. Anyway, I rather wanted to point out that the function _may, in some cases_ (if capacity is sufficient) change values in the slice, and that you should consume the output slice rather than reuse the input slice in any way after calling the function.

pure_utils/maps.go
```diff
@@ -0,0 +1,11 @@
+package pure_utils
+
+func Keys[K comparable, V any](m map[K]V) []K {
```
apognu, line 3: FYI: `slices.Collect(maps.Keys(m))`

apognu (approved):

PR #1061, opened 2025-06-21
Generalize workflow system.

apognu:
This PR improves the decision workflow feature to make it more expandable and useful. It allows the user to create an ordered series of `rules` on a scenario, of which the first one which `conditions` all evaluate to true will execute all `actions`.

This first version does not implement any further actions than those that were already there, so we have `CREATE_CASE` and `ADD_TO_CASE_IF_POSSIBLE` and three conditions, `always`, `never` and `outcome_in`.

So far, it only implement actual executions, not the management API or any kind of validation for the conditions and actions.

---

Outstanding TODOs:

 * [x] Backfill stable IDs for all iteration rules
 * [x] Do proper error handling and logging for workflow rules


apognu:
We need the migration to convert the old column-based workflows into the new system. Luckily, it should be quite straightforward side there weren't lots of options.

usecases/workflows_usecase.go
```diff
@@ -0,0 +1,324 @@
...
+	InsertWorkflowAction(ctx context.Context, exec repositories.Executor, action models.WorkflowAction) (models.WorkflowAction, error)
+	UpdateWorkflowAction(ctx context.Context, exec repositories.Executor, action models.WorkflowAction) (models.WorkflowAction, error)
+	DeleteWorkflowAction(ctx context.Context, exec repositories.Executor, ruleId, actionId string) error
+	ReorderWorkflowRules(ctx context.Context, exec repositories.Executor, scenarioId string, ids []uuid.UUID) error
+}
+
+func (uc *WorkflowUsecase) ListWorkflowsForScenario(ctx context.Context, scenarioId string) ([]models.Workflow, error) {
+	exec := uc.executorFactory.NewExecutor()
+
+	if err := uc.enforceSecurity.ListScenarios(uc.enforceSecurity.OrgId()); err != nil {
```
Pascal-Delange, line 47: this should rather get the scenario by its ID then check its org id - as it is, I think the permission check doesn't really restrict anything other than "the user can read scenarios" which is all roles

usecases/workflows_usecase.go
```diff
@@ -0,0 +1,324 @@
...
+	scenario, err := uc.scenarioRepository.GetScenarioById(ctx, exec, rule.ScenarioId)
+	if err != nil {
+		return models.WorkflowRule{}, err
+	}
+
+	if err := uc.enforceSecurity.PublishScenario(scenario); err != nil {
+		return models.WorkflowRule{}, err
+	}
+
+	return uc.repository.InsertWorkflowRule(ctx, exec, rule)
```
Pascal-Delange, line 71: isn't it necessary to call the "validate" function here as well ? Or are they always created empty ?

usecases/workflows_usecase.go
```diff
@@ -0,0 +1,324 @@
...
+	scenario, err := uc.scenarioRepository.GetScenarioById(ctx, exec, rule.ScenarioId)
+	if err != nil {
+		return models.WorkflowAction{}, err
+	}
+
+	if err := uc.enforceSecurity.PublishScenario(scenario); err != nil {
+		return models.WorkflowAction{}, err
+	}
+
+	return uc.repository.InsertWorkflowAction(ctx, exec, action)
```
Pascal-Delange, line 210: same as we have a condition validator, we probably need an action validator too right ?

repositories/workflows_repositories.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/checkmarble/marble-backend/repositories/dbmodels"
+	"github.com/google/uuid"
+)
+
+func (repo *MarbleDbRepository) ListWorkflowsForScenario(ctx context.Context, exec Executor, scenarioId string) ([]models.Workflow, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+
+	sql := NewQueryBuilder().
```
Pascal-Delange, line 17: I think the query as it is, is a bit incorrect (or more precisely, it's probably technically working in its current implementation but will start generating garbage if we every allow >1 action per workflow): doing the two left joins before grouping will generate more rows than it should if we every have >1 condition AND action for a given rule.
(I've been bitten by a similar thing on scenario rules in the past)

I suggest we either:
- do two subqueries with separate left joins then merge them
- OR we handle the join more "in code"
- OR we keep it as it is but we clearly document in the code that there should not be >1 action or the repo will break, and perhaps add some validation for this

Pascal-Delange (commented):
first review done !

usecases/workflows_usecase.go
apognu: Yeah, `WorkflowRule`s do not contain conditions and actions, always empty, though we will need to create an action validator.

usecases/workflows_usecase.go
apognu: Ah, well, see previous comment.

repositories/workflows_repositories.go
apognu: I think you're right for the effect (it would producer duplicate rows), but I believe using `array_agg(distinct row(c.*))` would fix it.

Pascal-Delange: I think that's right - and good enough at least as long as we expect limited number of rows of both per rule

PR #1065, opened 2025-06-25
Check yente service in liveness and enhance liveness response with detailes status

ghost:
The liveness endpoint only tests if the database is reachable. This PR adds a test for the yente service if the openSanction is configured.

We take the opportunity to enhance the endpoint response by adding details.

If one check failed, we return an 500 error as before to maintain compatibility. 
Suggestion: With the detailed response, we could send a 200 because the server is alive but in degraded mode.

The timeout is not catched by `PresentError` function

# UPDATE 1
Don't touch liveness endpoint. Keep the behavior. 
Instead, move the new behavior in a dedicated endpoint `health`


dto/liveness_dto.go
```diff
@@ -0,0 +1,35 @@
...
+type LivenessItemStatusResponse struct {
+	Name   string `json:"name"`
+	IsLive bool   `json:"is_live"`
+	Error  string `json:"error"`
+}
+
+func AdaptLivenessItemStatus(status models.LivenessItemStatus) LivenessItemStatusResponse {
+	var errorMessage string
+	if status.Error != nil {
+		errorMessage = status.Error.Error()
```
Pascal-Delange, line 21: I would rather avoid leaking raw error stacktraces/strings to the API response body - either we have a clear error string to which we want to map an error, or we use something akin to "withDetail" as used in the public API response format, or more likely (in this case) skip this altogether

api/handle_liveness_probe.go
```diff
@@ -11,13 +12,14 @@ func handleLivenessProbe(uc usecases.Usecases) func(c *gin.Context) {
 	return func(c *gin.Context) {
 		ctx := c.Request.Context()
 		usecase := uc.NewLivenessUsecase()
-		err := usecase.Liveness(ctx)
-		if presentError(ctx, c, err) {
```
Pascal-Delange, line 15: what was the problem with presentError ?
I'm rather in favor of using it instead of manual c.JSON, because it does a couple of other things (error logging, sentry reporting).

Pascal-Delange (commented):
On this
> Suggestion: With the detailed response, we could send a 200 because the server is alive but in degraded mode.

I think I'm against it, based on experience the way people consume liveness checks never goes (99.99% of cases) deeper than "did I get a response and is its status < 300"
This could be useful for a "detailed service availability" kind of endpoint, but then I wouldn't call it liveness.

api/handle_liveness_probe.go
Pascal-Delange: Ok I see,
But cf the global comment I left, I don't think it's the best tradeoff to lose logging & sentry reporting for a liveness check endpoint where I don't believe people will typically consume the response body detail.
@apognu  what is your take on this ?

leo-marble: Or maybe it doesn't matter and I can do something like 

```go
		for _, s := range status.Statuses {
			if presentError(ctx, c, s.Error) {
				return
			}
		}

		c.JSON(http.StatusOK, dto.AdaptLivenessStatus(status))
```

Pascal-Delange: you can and it will catch at least the first error (if any present) which is certainly good enough

apognu: I am still unsure that we want to be considering OpenSanctions being down as a break-the-world issue for the backend API. Having some reporting is one thing, but do we really want to prevent **any** operation on Marble if a non-essential dependency is down?

As it is, it would be impossible to review cases, for example, if there is a problem with Yente, not to mention that schedulers would restart the backend, and fail to bring it back up.

---

The way I see it, if a required dependency is missing (database), return a 503, there is nothing we can do anyway. On the other hand, if something like OpenSanctions is down, return it in the JSON, log it, publish it to Sentry, what have you, but it feels weird to bring down the entire platform.

Pascal-Delange: and with this, we have a relatively high confidence that we'll get the right error codes, at least as best as possible (e.g. 408 on context timeout)

Pascal-Delange: > I am still unsure that we want to be considering OpenSanctions being down as a break-the-world issue for the backend API. Having some reporting is one thing, but do we really want to prevent **any** operation on Marble if a non-essential dependency is down?
> 
> As it is, it would be impossible to review cases, for example, if there is a problem with Yente, not to mention that schedulers would restart the backend, and fail to bring it back up.
> 
> The way I see it, if a required dependency is missing (database), return a 503, there is nothing we can do anyway. On the other hand, if something like OpenSanctions is down, return it in the JSON, log it, publish it to Sentry, what have you, but it feels weird to bring down the entire platform.

Ok, I think that's correct, but also highlights that there are different usages of the "liveness" endpoint which are perhaps conflicting.
There is at least the "Cloud Run internal usage health check" usage (for which your comment is 100% spot on) and then there is external monitoring of the infrastructure (for our status page or for customers who want to monitor us) for which I think any business critical dependency should be taken into account to evaluate uptime.
Is the conclusion that we need two distinct endpoints ?

apognu: I believe so. The licenses endpoint should communicate through statuses and the other one should just provide a listing of dependencies (always 200).

usecases/liveness.go
```diff
@@ -11,11 +12,56 @@ type livenessRepository interface {
...
+		statuses = append(statuses, models.LivenessItemStatus{
+			Name:   models.DatabaseLivenessItemName,
+			IsLive: true,
+			Error:  nil,
+		})
+	}
+
+	// Check Open Sanctions health
+	if u.hasOpensanctionsSetup {
+		if ok, err := u.openSanctionsRepository.IsConfigured(ctx); err != nil || !ok {
```
apognu, line 49: I think we can collapse those tests to be more readable. With `IsLive: err == nil` and `Error: err`.

leo-marble: Better, thanks 👍

api/handle_liveness_probe.go
leo-marble: Yes 😃 
I pushed the changes, restored the liveness handler and I created a new one called `Health`. It is the same as what I did before but in a new endpoint. 
The details are in a list, we can add or remove dependencies as we want. 

We can decide if a bool is enough for the status value or we need a string like "healthy", "degraded", "not ready" etc ... 

We lose the error information, should I store the err of each dependencies and log/sentry them?

models/health.go
```diff
@@ -0,0 +1,26 @@
+package models
+
+type HealthItemName string
+
+const (
+	DatabaseHealthItemName      HealthItemName = "Database"
```
apognu, line 6: A bit pedanctic, but could we have slugs instead of full strings with uppercase letters and spaces?

If you look into `models/errors.go` (line 82), you already have some semblance of that with the `Requirement` system (and `open_sanctions` is already a constant there.

apognu (changes requested):

Pascal-Delange (commented):
got a remaining question on where we should place this - otherwise looks good to me

apognu (approved):

PR #1067, opened 2025-06-25
fix(annotation): stop checking if the target object exists

ghost:
Currently, the system does not allow the creation of an annotation on an item that has not been ingested. This restriction prevents users from creating annotations on an "object_id" that does not exist in the database, which can be a limitation in certain use cases.

To address this issue, I have removed the check for the existence of the object in the database. This change enables users to create annotations on "object_id" even if it is not currently ingested.

I also changed the function which returns `case` data by forcing annotation fetch before checking if the pivot object exists in client ingested data. Without that, we didn't return annotations if the object didn't exist.

### DIFF
--- a/usecases/entity_annotations_usecase.go
+++ b/usecases/entity_annotations_usecase.go
@@ -65,7 +65,7 @@ type EntityAnnotationUsecase struct {
 }
 
 func (uc EntityAnnotationUsecase) List(ctx context.Context, req models.EntityAnnotationRequest) ([]models.EntityAnnotation, error) {
-	if err := uc.checkObject(ctx, req.OrgId, req.ObjectType, req.ObjectId); err != nil {
+	if err := uc.checkObject(ctx, req.OrgId, req.ObjectType); err != nil {
 		return nil, errors.Wrap(models.NotFoundError, err.Error())
 	}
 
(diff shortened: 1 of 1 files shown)


usecases/ingested_data_reader_usecase.go
leo-marble, line 1: I moved the annotation block upper on the function because this step doesn't depend on the IngestedData in the client DB. We can have annotations related to a data which is not ingested yet by the client.

usecases/entity_annotations_usecase.go
```diff
@@ -89,10 +92,8 @@ func (uc EntityAnnotationUsecase) Get(ctx context.Context, orgId, id string) (mo
 func (uc EntityAnnotationUsecase) ListForObjects(ctx context.Context,
 	req models.EntityAnnotationRequestForObjects,
 ) (map[string][]models.EntityAnnotation, error) {
-	for _, objectId := range req.ObjectIds {
-		if err := uc.checkObject(ctx, req.OrgId, req.ObjectType, objectId); err != nil {
-			return nil, errors.Wrap(models.NotFoundError, err.Error())
-		}
+	if err := uc.checkObject(ctx, req.OrgId, req.ObjectType); err != nil {
+		return nil, errors.Wrap(models.NotFoundError, err.Error())
```
leo-marble, line 96: @apognu I didn't notice any impact by changing the logic here.

leo-marble:
bugbot run

usecases/ingested_data_reader_usecase.go
```diff
@@ -236,6 +234,21 @@ func (usecase IngestedDataReaderUsecase) enrichPivotObjectWithData(
 		return pivotObject, nil
 	}
 
+	// Annotations don't depend on the existence of the pivot object in IngestedData, so we can get them first
```
Pascal-Delange, line 237: 👍  ok makes sense

Pascal-Delange (approved):
Ok looks good ! 
Did you run it once end to end with the frontend?

apognu (approved):

PR #1068, opened 2025-06-27
Add initial queries even if they all have been removed by preprocessing.

apognu:
We used to bail early if preprocessing removed all queries, not setting the initial query.


Pascal-Delange (approved):
thanks ! cc @carere  it's available

PR #1071, opened 2025-07-01
feat(metrics): add metrics collection mechanism

leo-marble:
This PR introduces the concept of a metric collector, a new use case that runs periodically to launch all collectors. It introduces the notion of a Collector which has a list of collectors (both global and organization-specific), a collector is a module which implement how to compute/extract a metric.

A versioning system is implemented within the Collector to determine the list of collectors we are likely to receive and to know how to parse the results accordingly.

Now that the mechanism is implemented, the next steps will be to define the list of metrics and implement the collectors in usecases/metrics_collection, following the example of the stub.

This PR also introduces the concept of periodicity to determine if the metric is monthly (calendar-based), daily, or instantaneous. If there is an overlap between two periods, the collector will handle the separation and emit two items. A utility is provided to manage these date notions.

There are still some TODOs remaining, and there are PRs branching from this one to implement these TODOs. Some configuration still needs to be set up.


pure_utils/time_ranges.go
```diff
@@ -0,0 +1,67 @@
...
+			To:   periodEnd,
+		})
+
+		current = periodEnd
+	}
+
+	return ranges, nil
+}
+
+func getNextPeriodBoundary(current time.Time, frequency Frequency) (time.Time, error) {
```
apognu, line 54: This may have two logic issues as far as date handling is concerned:

First, I think we should normalize to UTC before doing any kind of calculation, otherwise, those `AddDate()` will do different things depending on the timezone, and therefore might provide invalid results.

Also, I don't think we can use something as simple as `AddDate()` for this algorithm, where, for example, `time.Date(2025, 1, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 1, 0)` gives March 3rd.

You took leap years old into account in your tests, but this is a more general fact that not all months are the same length, so the big would technically trigger every last day of 31-day months.

usecases/metrics_collection/collector_test.go
```diff
@@ -0,0 +1,391 @@
...
+	// Assert
+	require.NoError(t, err)
+	assert.Equal(t, "test-v1", result.Version)
+	assert.NotEmpty(t, result.CollectionID)
+
+	// Should have 3 metrics total (1 global + 2 org metrics)
+	assert.Len(t, result.Metrics, 3)
+
+	// Verify metrics content
+	assert.True(t, slices.ContainsFunc(result.Metrics, func(m models.MetricData) bool {
```
apognu, line 123: There are `assert.Contains()` and `assert.Containsf()` methods that will give better error messages when failing.

pure_utils/time_ranges.go
leo-marble: Thank you. Initially, I did it without the AddDate() function and changed the month and day directly when creating the new date because the `time.Date()` manages the cases. However, I found it more explicit and clear to use AddDate() without having to think about the edge case you spotted. I have fixed the issue and added more tests.

Regarding UTC, I think it is better to keep the timezone as it is. The function is a utility, and it should be flexible. I prefer to let the caller be responsible for the timezone they want to manipulate.

For metrics, we have to use UTC date, especially if the usage is for billing and they want a uniform and agnostic timezone.

usecases/metrics_collection/collector_test.go
leo-marble: It is better, thanks :-)

leo-marble:
cf: https://github.com/checkmarble/marble-backend/pull/1074 -> Introduce the watermark to bound the time range

apognu:
Notes from our discussion :

 * Maybe, for collection reports **with** a license key, check that the license key is valid.
 * Outside the scope of this PR, but think about adding some safeguards, since those are unauthenticated requests (rate limiting on the Load Balancer?).

dto/metrics_dto.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/google/uuid"
+)
+
+type MetricCollectionFrequencyDto string
+
+const (
```
apognu, line 13: To be discused, but I think we are not going to make any difference between metric frequency, they are all going to be treated the same, so those might be moot.

pure_utils/bigquery.go
```diff
@@ -0,0 +1,23 @@
+package pure_utils
+
+import "cloud.google.com/go/bigquery"
+
+func NullStringFromPtr(ptr *string) bigquery.NullString {
```
apognu, line 5: Could we specify a name that signifies this is specifically for BigQuery? Like `BigQueryNullStringFrom`?

repositories/decisions_repository.go
```diff
@@ -1041,3 +1041,51 @@ func (repo *MarbleDbRepository) RemoveDecisionRulePayload(ctx context.Context, t
...
+
+	rows, err := exec.Query(ctx, sql, args...)
+	if err != nil {
+		return map[string]int{}, err
+	}
+	defer rows.Close()
+
+	counts := make(map[string]int)
+	for rows.Next() {
+		var orgId string
```
apognu, line 1072: This can be extracted outside of the loop, to be reused for all rows.

repositories/decisions_repository.go
```diff
@@ -1041,3 +1041,51 @@ func (repo *MarbleDbRepository) RemoveDecisionRulePayload(ctx context.Context, t
...
+		if err := rows.Scan(&orgId, &count); err != nil {
+			return map[string]int{}, err
+		}
+		counts[orgId] = count
+	}
+
+	// If the orgIds are not in the counts, the count will be 0
+	countsMap := make(map[string]int, len(orgIds))
+	for _, orgId := range orgIds {
+		if count, exists := counts[orgId]; exists {
```
apognu, line 1083: This can be collapsed (and `countsMap` can be removed) if you only test if an org ID is not present, then you set to 0.

cmd/server.go
```diff
@@ -193,6 +197,13 @@ func RunServer(config CompiledConfig) error {
 		return err
 	}
 
+	bigQueryClient, err := infra.NewBigQueryClient(ctx, bigQueryConfig)
```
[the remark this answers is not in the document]
apognu: If we commit to always use a dataset in the same project as the service account, then yes, I agree.

repositories/decisions_repository.go
Pascal-Delange: unnecessary optimization no ?
The query should yield tens, not thousands of rows

models/metrics_collection.go
```diff
@@ -0,0 +1,109 @@
...
+
+type MetricData struct {
+	Name           string
+	Numeric        *float64
+	Text           *string
+	Timestamp      time.Time
+	OrganizationID *string    // Only for org-specific metrics
+	From           *time.Time // Optional start time for time range metrics
+	To             *time.Time // Optional end time for time range metrics
+	Frequency      MetricCollectionFrequency
```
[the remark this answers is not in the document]
apognu: Do we actually care about the recorded time? Especially considering river could trigger the job late, a network issue could make the collection catch up way later, or we're going to have backfill which will all have around the same (very delayed) collection time?

The `From` and `To` renfer to watermarks, which map 1:1 to the time period we collect the data for. I agree those may be a bit skewed, but we cannot really consider events are related to the time Marble will receive them.

pure_utils/bigquery.go
apognu: They are, for BQ's `OPTIONAL` columns.

pure_utils/time_ranges.go
```diff
@@ -0,0 +1,78 @@
...
+	return ranges, nil
+}
+
+// getNextPeriodBoundary returns the start of the next calendar period based on the given frequency.
+// For daily frequency, it returns the next day at 00:00:00.
+// For monthly frequency, it returns the first day of the next month at 00:00:00.
+func getNextPeriodBoundary(current time.Time, frequency Frequency) (time.Time, error) {
+	switch frequency {
+	case FrequencyDaily:
+		return time.Date(current.Year(), current.Month(), current.Day()+1, 0, 0, 0, 0, current.Location()), nil
```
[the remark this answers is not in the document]
apognu: Go actually encode `max(unit)+1` for each time unit to `min(unit+1)`. So day 7 is Sunday, month 12 is January, Jan 32nd is Feb 1st. As far as I know, it **is** the correct way to do this, since it is adding "units" that does the wrong thing.

We discussed this before, see [this comment](https://github.com/checkmarble/marble-backend/pull/1071/files#r2181740200).

Pascal-Delange:
Side question btw, you did the dataset and table creation on BQ manually ?

apognu:
> Side question btw, you did the dataset and table creation on BQ manually ?

For now, yes, manually in the sandbox account. Once we end up on a finalized schema, I will write the Terraform for it.

api/handle_metrics_ingestion.go
```diff
@@ -0,0 +1,40 @@
...
+		err := usecase.IngestMetrics(c.Request.Context(), metricsCollection)
+		if presentError(c.Request.Context(), c, err) {
+			logger.WarnContext(c.Request.Context(), "Failed to ingest metrics", "error", err)
+			return
+		}
+
+		// Success response
+		c.JSON(http.StatusOK, gin.H{
+			"status":        "success",
+			"message":       "Metrics collection processed successfully",
```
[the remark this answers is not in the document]
leo-marble: It was for my test. I removed the body because it doesn't give any important info.

pure_utils/time_ranges.go
```diff
@@ -0,0 +1,78 @@
...
+	// If the from time is equal to the to time, return a single period
+	if from.Equal(to) {
+		return []TimeRange{{From: from, To: to}}, nil
+	}
+
+	// Split into individual periods aligned to calendar boundaries
+	var ranges []TimeRange
+	current := from
+
+	// Protect against infinite loops, 1000 iterations represents 1000 days, 1000 months (83 years), etc.
```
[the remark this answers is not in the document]
leo-marble: You mean, compute the number of potential time range based on a computation `to - from` before doing the loop ?

Something like 

```go
CONST MAX_RANGE = 10000

if monthy && (to - from).month > MAX_RANGE {
  return error
}
if daily && (to - from).day > MAX_RANGE {
  return error
}
```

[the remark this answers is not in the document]
Pascal-Delange: I'm not sure of the syntax, but yea essentially this

repositories/decisions_repository.go
leo-marble: I'm not sure if I understand what you said. 
I used a utility function to get the result and parsed it into a struct. 

I also removed the need for an intermediary map variable.

usecases/metrics_ingestion.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		if !errors.Is(err, models.NotFoundError) {
+			utils.LogAndReportSentryError(ctx, err)
+		}
+
+		logger.WarnContext(ctx, "Invalid license", "license", licenseKey)
+		return models.License{}, errors.Wrap(models.UnAuthorizedError, "invalid license")
+	}
+	return license, nil
+}
```
leo-marble, line 72: @apognu @Pascal-Delange Should we raise an error and return a 401 error ? 

If not, what is the right behavior? Ignore and send the bad license key in BQ ?

Pascal-Delange: I think I had a chat about this but it must have been with Antoine ?
Anyway, I think we maybe log an info log, but mostly just ignore the invalid license key and treat it as a call without license key.

models/health.go
```diff
@@ -5,6 +5,7 @@ type HealthItemName string
 const (
 	DatabaseHealthItemName      HealthItemName = "database"
 	OpenSanctionsHealthItemName HealthItemName = "open_sanctions"
+	BigQueryHealthItemName      HealthItemName = "bigquery"
```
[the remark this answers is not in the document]
leo-marble: Deleted

usecases/metrics_collection/case_collector.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/checkmarble/marble-backend/repositories"
+	"github.com/checkmarble/marble-backend/usecases/executor_factory"
+	"github.com/checkmarble/marble-backend/utils"
+)
+
+const (
```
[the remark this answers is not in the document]
leo-marble: You are right!

infra/bigquery_client.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+
+	"cloud.google.com/go/bigquery"
+)
+
+type BigQueryConfig struct {
+	ProjectID string
+}
+
+type BigQueryClient struct {
```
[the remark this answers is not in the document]
leo-marble: I changed the struct to store the Table object from BigQuery. In our code, we don't need to manage the client. 

I also took the opportunity to create an init function, if we are not running on a GCP prod/staging machine, we don't init BQ client.

cmd/server.go
```diff
@@ -193,6 +199,13 @@ func RunServer(config CompiledConfig) error {
...
```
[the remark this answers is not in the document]
leo-marble: cf: https://github.com/checkmarble/marble-backend/pull/1071#discussion_r2207725854

docker-compose.yml
```diff
@@ -52,6 +52,7 @@ services:
 
 volumes:
   postgres-db:
+    name: marble-backend-postgres-db
```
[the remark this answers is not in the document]
leo-marble: It is not backward compatible, if we `docker compose up`, a new volume will be created.

usecases/scheduled_execution/metrics_collection_job.go
```diff
@@ -0,0 +1,161 @@
...
+	return river.NewPeriodicJob(
+		river.PeriodicInterval(config.JobInterval),
+		func() (river.JobArgs, *river.InsertOpts) {
+			return models.MetricsCollectionArgs{}, &river.InsertOpts{
+				Queue: "metrics",
+				UniqueOpts: river.UniqueOpts{
+					ByQueue:  true,
+					ByPeriod: config.JobInterval,
+				},
+				MaxAttempts: 1, // No retries
```
leo-marble, line 31: Don't know if we allow retries or not. 

To send data to server, there is a retry mechanism

usecases/scheduled_execution/metrics_collection_job.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type WatermarkRepository interface {
+	GetWatermark(ctx context.Context, exec repositories.Executor, orgId *string,
+		watermarkType models.WatermarkType) (*models.Watermark, error)
+	SaveWatermark(ctx context.Context, tx repositories.Transaction,
+		orgId *string, watermarkType models.WatermarkType, watermarkId *string, watermarkTime time.Time, params json.RawMessage) error
+}
+
+type MetricCollectionWorker struct {
```
Pascal-Delange, line 49: It is best practice to also explicitly define the worker's timeout by implementing Timeout() (see the other worker examples).
It will use the default river timeout if undefined, but it's just nice to have it explicit.

repositories/watermark_repository.go
```diff
@@ -0,0 +1,61 @@
...
+
+	sql := NewQueryBuilder().
+		Select(dbmodels.SelectWatermarkColumn...).
+		From(dbmodels.TABLE_WATERMARKS).
+		Where(whereClause)
+
+	return SqlToOptionalModel(ctx, exec, sql, dbmodels.AdaptWatermark)
+}
+
+func (repo *MarbleDbRepository) SaveWatermark(ctx context.Context, tx Transaction,
```
Pascal-Delange, line 35: because this only does one piece of work, it can expect an exec Executor like the other repos

usecases/scheduled_execution/metrics_collection_job.go
```diff
@@ -0,0 +1,161 @@
...
+	exec := w.executorFactory.NewExecutor()
+	watermark, err := w.repository.GetWatermark(ctx, exec, nil, models.WatermarkTypeMetrics)
+	if err != nil || watermark == nil || watermark.WatermarkTime.IsZero() {
+		return time.Now().UTC().Add(-fallbackDuration)
+	}
+	return watermark.WatermarkTime
+}
+
+func (w MetricCollectionWorker) saveWatermark(ctx context.Context, newWatermarkTime time.Time) error {
+	return w.transactionFactory.Transaction(ctx, func(tx repositories.Transaction) error {
```
Pascal-Delange, line 121: No need to wrap this in a transaction (see the other comment)

usecases/scheduled_execution/metrics_collection_job.go
Pascal-Delange: hmm no, probably not strictly necessary ?

usecases/metrics_ingestion.go
```diff
@@ -0,0 +1,67 @@
...
+	license, err := u.licenseRepository.GetLicenseByKey(ctx,
+		u.executorFactory.NewExecutor(), licenseKey)
+	if err != nil {
+		if !errors.Is(err, models.NotFoundError) {
+			utils.LogAndReportSentryError(ctx, err)
+		}
+
+		logger.InfoContext(ctx, "Invalid license fetched in metrics collection", "license", licenseKey)
+		return models.License{}, errors.New("invalid license")
+	}
```
Pascal-Delange, line 65: a last bit of pedantic comment - not because I think it is critical here, but because it's a pattern I'm globally trying to push:
We try to handle errors only once - you should see this applied pretty consistently in the codebase. In this context, logging the error is handling it. In practice, this usually means:
- if we catch a generic error, attach any useful additional context to it by using errors.Wrap, then propagate it higher up the stack where some boundary is responsible for catching it (typically, the worker middleware or the API handler)
- if we catch a very specific error for which we want to apply specific logic, apply that logic right where we caught it  and then don't propagate an error.

Concretely, what I'd do here is:
- if we get a repository error other than "not found", just annotate it and return it (the handler will log it)
- if we get a "not found error", maybe info log it and don't return an error, just ignore the license key
- if we found it, add the name to the collection

Following this patter, in the current situation, also points out that cutting the logic into nested function calls with validateLicense being called from IngestMetrics is probably detrimental to the software architecture, and that they would live better together in one unit of code.

Pascal-Delange (commented):
Last round of comments, I promise 🙏

repositories/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/google/uuid"
+)
+
+func selectMetadata() squirrel.SelectBuilder {
+	return NewQueryBuilder().
+		Select(dbmodels.MetadataFields...).
+		From(dbmodels.TABLE_METADATA)
+}
+
+func (repo *MarbleDbRepository) GetMetadata(ctx context.Context, exec Executor, orgID *uuid.UUID, key models.MetadataKey) (models.Metadata, error) {
```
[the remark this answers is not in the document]
leo-marble: This method uses the `SqlToRow` function, which returns an error if the model is not found. My initial intention was to have the caller handle the error, rather than checking both the error and the pointer value within this function.

repositories/watermark_repository.go
Pascal-Delange: No that's not a problem, because repositories.Executor and repositories.Transaction are interfaces, and any Transaction also implements Executor

infra/bigquery.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func InitializeBigQueryInfra(ctx context.Context, config BigQueryConfig) (*BigQueryInfra, error) {
+	projectId, err := GetProjectId()
+	if err != nil {
+		return nil, err
+	}
+
+	// Init BigQuery client only for marble saas projects
+	if !slices.Contains(MarbleSaasProjectIds, projectId) {
```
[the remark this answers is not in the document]
leo-marble: Since the usecase is only called by a route which is under SaasOnly flag, the error will not be printed anymore in case of self-hosted

Pascal-Delange (approved):
👍

PR #1074, opened 2025-07-03
Refactor offloading watermark and add watermark in metrics collection job

ghost:
This PR modifies the `offloading_watermark` table to be renamed to `watermark`. This change allows us to move away from the specific offloading use case and makes the table more generic.
Additionally, this PR introduces the concept of watermarking in the metric collection system to provide a reference for the lower bound on dates. The param attribute allows adding extra parameters based on the type of watermark if the existing fields are insufficient. I introduce the notion of watermark `type`

I have ensured that the same fields as in the offloading table are retained, although some field names have been changed for clarity and consistency.

The migration process is done in two steps:

    Create the new watermark table.
    Migrate the data from the offloading_watermark table to the new watermark table.
    Delete the old offloading_watermark table.

During the migration, the table is locked to prevent any access and ensure data integrity.

Additionally, this PR introduces the concept of watermarking in the metric collection system to provide a reference for the lower bound on dates.

### DIFF
--- a/models/offloading_watermark.go
+++ b/models/offloading_watermark.go
@@ -2,9 +2,10 @@ package models
 
 import (
 	"encoding/json"
+	"errors"
 	"time"
 
-	"github.com/cockroachdb/errors"
+	"github.com/google/uuid"
 )
 
 type WatermarkType string
@@ -18,18 +19,19 @@ func (t WatermarkType) String() string {
 	return string(t)
 }
 
-func WaterMarkTypeFromString(s string) (WatermarkType, error) {
+func WatermarkTypeFromString(s string) (WatermarkType, error) {
 	switch s {
 	case "decision_rules":
 		return WatermarkTypeDecisionRules, nil
 	case "metrics":
 		return WatermarkTypeMetrics, nil
 	default:
-		return "", errors.Newf("invalid watermark type: %s", s)
+		return "", errors.New("invalid watermark type")
 	}
 }
 
 type Watermark struct {
+	Id            uuid.UUID
 	OrgId         *string
 	Type          WatermarkType
 	WatermarkTime time.Time
--- a/models/metrics_collection.go
+++ b/models/metrics_collection.go
@@ -15,8 +15,8 @@ const (
 )
 
 type MetricData struct {
-	Name           string // Can be int, string, float, etc.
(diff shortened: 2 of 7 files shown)


repositories/decisions_repository.go
```diff
@@ -980,12 +980,17 @@ func (repo *MarbleDbRepository) GetOffloadableDecisionRules(ctx context.Context,
...
 	if req.Watermark == nil {
-		req.Watermark = &models.OffloadingWatermark{
+		req.Watermark = &models.Watermark{
 			WatermarkTime: time.Time{},
-			WatermarkId:   uuid.UUID{}.String(),
+			WatermarkId:   utils.Ptr(uuid.UUID{}.String()),
 		}
 	}
 
+	// HELP: Not sure about raising the error here
```
leo-marble, line 989: Don't know how to deal with the error

apognu: I think returning an error is correct. it would happen if the function was passed a global watermark instead of a decision rules one. Maybe we can clarify the error to reflect that.

apognu (approved):
Just left a response to one of your question. Let's do that and merge into the other branch, so we can review it fully functional!

models/offloading_watermark.go
```diff
@@ -1,12 +1,42 @@
...
+type Watermark struct {
+	Id            uuid.UUID
+	OrgId         *string
+	Type          WatermarkType
 	WatermarkTime time.Time
-	WatermarkId   string
+	WatermarkId   *string
 	CreatedAt     time.Time
 	UpdatedAt     time.Time
+	Params        *json.RawMessage
```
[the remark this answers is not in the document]
leo-marble: I see, thanks :) 

image

repositories/decisions_repository.go
leo-marble: I will fix it in https://github.com/checkmarble/marble-backend/pull/1071 after rebase

models/offloading_watermark.go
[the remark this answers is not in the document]
leo-marble: I will remove the pointer in https://github.com/checkmarble/marble-backend/pull/1071 after rebase. 
Thanks

PR #1078, opened 2025-07-08
Work on audit trail

apognu:
This is preliminary work to make audit trail more useful and complete.

 * It adds the organization ID to the audit (with backfill) so it could be used as is.
 * Also, it sets the session settings used to create the trail in all DB executors (simple and transactional) so the only thing needed to add a trail is to create the trigger.

This PR will also contain new commits that adds new kinds of audit entries.

---

A few more triggers are created to add more audit trails, on top of the original custom list management recording, namely:

 * User and permissions management
 * API keys management
 * Changing the outcome of a case
 * Changing the review status of a decision
 * Changing the status of a screening
 * Committing a scenario iterations
 * Publishing a scenario iteration
 * Creating a rule snooze
 * Whitelisting a screening counterparty

### DIFF
--- a/repositories/migrations/20250710093900_better_audit_trails.sql
+++ b/repositories/migrations/20250710093900_better_audit_trails.sql
@@ -3,7 +3,8 @@
 
 alter table "audit"."audit_events"
     add column if not exists org_id uuid,
-    add column if not exists api_key_id uuid;
+    add column if not exists api_key_id uuid,
+    add column if not exists previous_data jsonb;
 
 update "audit"."audit_events" events
 set org_id = sq.organization_id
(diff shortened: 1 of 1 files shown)


repositories/utils.go
```diff
@@ -4,22 +4,29 @@ import (
...
-			fmt.Sprintf("SELECT SET_CONFIG('%s', $1, false)", postgres_audit_user_id_parameter),
-			*userId,
-		)
-		return err
+func injectDbSessionConfig(ctx context.Context, exec TransactionOrPool) (pgconn.CommandTag, error) {
+	if creds, ok := utils.CredentialsFromCtx(ctx); ok && creds.ActorIdentity.UserId != "" {
+		if tag, err := exec.Exec(ctx, "SELECT SET_CONFIG($1, $2, false)", postgres_audit_user_id_parameter, creds.ActorIdentity.UserId); err != nil {
+			return tag, err
+		}
+		if creds.OrganizationId != "" {
```
apognu, line 22: A user _should_ always have an organization, but they don't in integration tests. In any case, better safe than sorry since an error in the trigger function bubbles up.

repositories/migrations/20250708172300_add_org_id_to_audit_trail.sql
```diff
@@ -0,0 +1,115 @@
...
+    add column if not exists org_id uuid;
+
+update "audit"."audit_events" events
+set org_id = sq.organization_id
+from ( select id, organization_id from users ) sq
+where sq.id::text = events.user_id;
+
+create or replace function global_audit() returns trigger as $$
+begin
+    if current_setting('custom.current_user_id', true) is null then
```
apognu, line 14: This assumes we do not have intended audit trails that are not mapped to a user ID. Indirectly, this filters out changes through the API, but depending on what we want, we can do it differently.

apognu: Also, as a side effect compared to currently, it does not generate an audit entry when modifying the database directly.

repositories/db_executor_getter.go
```diff
@@ -77,6 +77,10 @@ func (g ExecutorGetter) Transaction(
 
 	execInTransaction := func() error {
 		return pgx.BeginFunc(ctx, pool, func(tx pgx.Tx) error {
+			if _, err := injectDbSessionConfig(ctx, tx); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
apognu: The three other are not for transactions, they're for regular connections, so we do need to have those for times, once at the beginning of transactions, and once for every type of query for regular connections.

Those three could have been joined into one in `NewExecutor()`, but that would require to pass in a context and change the bazillion call sites, which I think is a bit overkill.

repositories/utils.go
Pascal-Delange: and then there are marble admins, who don't have an organization, so that's fine

repositories/migrations/20250708172300_add_org_id_to_audit_trail.sql
Pascal-Delange: > Also, as a side effect compared to currently, it does not generate an audit entry when modifying the database directly.

That I think is fine, I don't want to start generating audit logs when we do DB migrations of any sort

repositories/db_executor_getter.go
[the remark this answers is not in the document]
leo-marble: Maybe we can add a comment to specify we have the same behavior as regular connections but it is injected in `func (g ExecutorGetter) Transaction()`

leo-marble (approved):

Pascal-Delange:
> Shouldn't we log when:
> 
> * scenario is created
> * scenario_iteration_rules is created, updated, deleted

To answer this:
- scenario is created: no, not very impactful (a newly created scenario is just an empty hull)
- scenario_iteration_rules is created, updated, deleted: updated: hmmm it's tricky - during edition it's a heavily updated object. Maybe we can do it but only if we don't store the detail of AST rule updates. Creation: not so impactful. Deletion: yes.

PR #1082, opened 2025-07-10
fix(docs): update database volume name in README

leo-marble:
Fix the command to reset the database


Pascal-Delange (commented):
I'm not 100% sure how this works, but I'm afraid it names the volume in function of the folder name you are running docker-compose in. In which case your PR works if we launched the backend's docker compose from the "api" submodule cloned with the marble repo, but but the old version works with the marble backend repo cloned directly.

leo-marble:
@Pascal-Delange Hum ... in that case we can fix the name of the volume like 
```
volumes:
  postgres-db:
    name: marble-api_postgres-db
    driver: local
```

But it create a new volume and doesn't transfer your old persisted data ...

apognu:
I thought we named the containers specifically to "fix" the directory name thing? Maybe we did it only for the Marble repo?

leo-marble:
> I thought we named the containers specifically to "fix" the directory name thing? Maybe we did it only for the Marble repo?

@apognu I don't think Docker use the container name in the volume name, but we can set a name to the volume

Pascal-Delange:
I'm fine with naming the volumes, then we can always copy the volume content to the new one if we don't want to lose our dev local data

Pascal-Delange (approved):
let's merge this 👍

PR #1092, opened 2025-07-21
Case auto-assignment

apognu:
This PR implements auto-assignments for cases. The exact details can be found in Linear, but overall:

 * A license entitlement enables case auto-assignment
 * An org can set a maximum size for the case backlog per user, per inbox
 * An inbox can enable auto-assignment
 * An inbox user can be added to an inbox rotation
 * A new endpoint can be used by a user to declare they are not available until a set date

Whenever elligible actions are performed, a task is enqueues that is going to run the algorithm to assign non-assigned, non-closed cases to available users, if their queue has slots. Those actions, for now are:

 * A case is created
 * A case is closed

The oldest case is assigned to the users with the least open, assigned cases, and so on until either all cases are assigned or there are no available slots in the available user's queues.

---

TODO:

 - [ ] Improve unavailability API: right now, it only asks for an end date and sets the unavailability to `[now()-end]`. I think we might want to do better than that, even if that is a good first use case. Changing the end date is possible until the period is passed, and cancelling the unavailability moves the end date to `now()`.
 - [ ] Add more hook points to trigger auto-assignment

### DIFF
--- a/repositories/auto_assignment_repository.go
+++ b/repositories/auto_assignment_repository.go
@@ -11,31 +11,46 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
+// This SQL query and the one below basically do the following
+//   - For an organization;
+//   - Select all active users that:
+//   - Configured as part of a rotation
+//   - Have less than X active cases assigned to them (through the lateral subquery)
+//   - Are currently available (through the `not exists` subquery)
 func (repo *MarbleDbRepository) FindAutoAssignableUsers(ctx context.Context, exec Executor, orgId string, limit int) ([]models.UserWithCaseCount, error) {
 	sql := `
-		select u.*, count(distinct c.id) filter (where c.status != 'closed') as case_count
+		select u.*, count(distinct c.id) as case_count
 		from inbox_users iu
 		inner join users u on
 		  u.organization_id = $1 and
 		  u.id = iu.user_id and
 		  u.deleted_at is null and
 		  iu.auto_assignable
-		left join cases c on c.assigned_to = u.id
+		  left join lateral (
+		    select c.id
+		    from cases c
+		    where
+		  	  c.org_id = u.organization_id and
+		  	  c.assigned_to = u.id and
+		  	  c.status != 'closed' and
+		  	  coalesce(c.snoozed_until, to_timestamp(0)) < now()
+			  limit $2
+		  ) c on true
 		where
 		  not exists (
 		    select 1
 		    from user_unavailabilities uu
 		    where
-				uu.org_id = $2 and
+				uu.org_id = $1 and
 				uu.user_id = iu.user_id and
 				now() between uu.from_date and uu.until_date
 	  	  )
 		group by u.id
-		having count(distinct c.id) filter (where c.status != 'closed') < $3
+		having count(distinct c.id) < $2
 		order by case_count
 	`
 
-	rows, err := exec.Query(ctx, sql, orgId, orgId, limit)
+	rows, err := exec.Query(ctx, sql, orgId, limit)
 
 	if err != nil {
 		return nil, err
@@ -58,7 +73,7 @@ func (repo *MarbleDbRepository) FindAutoAssignableUsers(ctx context.Context, exe
 
 func (repo *MarbleDbRepository) FindNextAutoAssignableUserForInbox(ctx context.Context, exec Executor, orgId string, inboxId uuid.UUID, limit int) (*models.User, error) {
 	sql := `
-		select u.*, count(distinct c.id) filter (where c.status != 'closed') as case_count
+		select u.*, count(distinct c.id) as case_count
 		from inbox_users iu
 		inner join users u on
 		  u.organization_id = $1 and
@@ -66,23 +81,32 @@ func (repo *MarbleDbRepository) FindNextAutoAssignableUserForInbox(ctx context.C
 		  u.deleted_at is null and
 		  iu.inbox_id = $2 and
 		  iu.auto_assignable
-		left join cases c on c.assigned_to = u.id
+		left join lateral (
+		  select c.id
+		  from cases c
+		  where
+			c.org_id = u.organization_id and
+			c.assigned_to = u.id and
+			c.status != 'closed' and
+		    coalesce(c.snoozed_until, to_timestamp(0)) < now()
+		  limit $3
+		) c on true
 		where
 		  not exists (
 		    select 1
 		    from user_unavailabilities uu
 		    where
-				uu.org_id = $3 and
+				uu.org_id = $1 and
 				uu.user_id = iu.user_id and
 				now() between uu.from_date and uu.until_date
 	  	  )
 		group by u.id
-		having count(distinct c.id) filter (where c.status != 'closed') < $4
+		having count(distinct c.id) < $3
 		order by case_count, id asc
 		limit 1
 	`
 
-	rows, err := exec.Query(ctx, sql, orgId, inboxId, orgId, limit)
+	rows, err := exec.Query(ctx, sql, orgId, inboxId, limit)
 	if err != nil {
 		return nil, err
 	}
@@ -100,16 +124,17 @@ func (repo *MarbleDbRepository) FindNextAutoAssignableUserForInbox(ctx context.C
 	return &user.User, nil
 }
 
-func (repo *MarbleDbRepository) FindAutoAssignableCases(ctx context.Context, exec Executor, limit int) ([]models.Case, error) {
+func (repo *MarbleDbRepository) FindAutoAssignableCases(ctx context.Context, exec Executor, orgId string, limit int) ([]models.Case, error) {
 	sql := NewQueryBuilder().
 		Select(columnsNames("c", dbmodels.SelectCaseColumn)...).
 		From(dbmodels.TABLE_CASES+" c").
 		InnerJoin(dbmodels.TABLE_INBOXES+" i on i.id = c.inbox_id").
 		Where(squirrel.Eq{
-			"i.auto_assign_enabled": true,
+			"c.org_id":              orgId,
 			"c.assigned_to":         nil,
-			"c.snoozed_until":       nil,
+			"i.auto_assign_enabled": true,
 		}).
+		Where("coalesce(c.snoozed_until, to_timestamp(0)) < now()").
 		Where("c.status != ?", models.CaseClosed).
 		OrderBy("c.created_at").
 		Limit(uint64(limit))
--- a/repositories/task_queue_repository.go
+++ b/repositories/task_queue_repository.go
@@ -252,7 +252,12 @@ func (r riverRepository) EnqueueAutoAssignmentTask(
 			InboxId: inboxId,
 		},
 		&river.InsertOpts{
-			Queue: orgId,
+			Queue:       orgId,
+			ScheduledAt: time.Now().Add(time.Minute),
+			UniqueOpts: river.UniqueOpts{
+				ByQueue:  true,
+				ByPeriod: 2 * time.Minute,
+			},
 		})
 
 	if err != nil {
--- a/usecases/auto_assignment_usecase.go
+++ b/usecases/auto_assignment_usecase.go
@@ -20,7 +20,7 @@ type autoAssignmentCaseRepository interface {
 type autoAssignmentRepository interface {
 	FindAutoAssignableUsers(ctx context.Context, exec repositories.Executor, orgId string, limit int) ([]models.UserWithCaseCount, error)
 	FindNextAutoAssignableUserForInbox(ctx context.Context, exec repositories.Executor, orgId string, inboxId uuid.UUID, limit int) (*models.User, error)
-	FindAutoAssignableCases(ctx context.Context, exec repositories.Executor, limit int) ([]models.Case, error)
+	FindAutoAssignableCases(ctx context.Context, exec repositories.Executor, orgId string, limit int) ([]models.Case, error)
 }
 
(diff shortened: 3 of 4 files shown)


apognu:
First light review for decision considerations welcome. 🙏

dto/license_dto.go
```diff
@@ -20,6 +20,7 @@ type LicenseEntitlements struct {
 	RuleSnoozes    bool `json:"rule_snoozes"`
 	TestRun        bool `json:"test_run"`
 	Sanctions      bool `json:"sanctions"`
+	AutoAssignment bool `json:"auto_assignment"`
```
Pascal-Delange, line 23: let's call it "case_auto_assignment" ?

repositories/migrations/20250721145200_create_auto_assignment.sql
```diff
@@ -0,0 +1,36 @@
...
+    from_date timestamp with time zone not null,
+    until_date timestamp with time zone not null,
+    created_at timestamp with time zone not null default now(),
+    updated_at timestamp with time zone,
+
+    constraint fk_org_id foreign key (org_id) references organizations (id) on delete cascade,
+    constraint fk_user_id foreign key (user_id) references users (id) on delete cascade
+);
+
+create index idx_user_avail_org_id_dates on user_unavailabilities (org_id, from_date, until_date);
```
leo-marble, line 25: I don't think the index on `org_id` is appropriate here. Your queries filter on `user_id`, not `org_id`.

models/user_settings.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/google/uuid"
+)
+
+type UserUnavailability struct {
+	Id        uuid.UUID
+	OrgId     uuid.UUID
+	UserId    uuid.UUID
+	FromDate  time.Time
+	UntilDate    time.Time
```
Pascal-Delange, line 14: this should have been formatted, weird 🤔

repositories/migrations/20250721145200_create_auto_assignment.sql
```diff
@@ -0,0 +1,36 @@
...
+
+alter table licenses
+    add column auto_assignment bool not null default false;
+
+-- +goose Down
+
+alter table organizations drop column auto_assign_queue_limit;
+alter table inboxes drop column auto_assign_enabled;
+alter table inbox_users drop column auto_assignable;
+
```
leo-marble, line 35: We should remove the `idx_user_avail_org_id_dates` index here.

usecases/auto_assignment_usecase.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	assignableUsers, err := uc.repository.FindAutoAssignableUsers(ctx, uc.executorFactory.NewExecutor(), inboxId)
+	if err != nil {
+		return errors.Wrap(err, "could not find assignable users")
+	}
+
+	slots := 0
+	for _, user := range assignableUsers {
+		slots += org.AutoAssignQueueLimit - user.CaseCount
```
leo-marble, line 51: I think it's possible for `CaseCount` to be greater than `AutoAssignQueueLimit`. In `FindAutoAssignableUsers`, we don't distinguish between cases that were auto-assigned and those that were manually assigned. I might be wrong, but I can imagine a scenario where a user already has `AutoAssignQueueLimit` cases assigned and then manually assigns themselves another case.

I think we should use `slots += int(math.Max(float64(org.AutoAssignQueueLimit - user.CaseCount), 0.0))` here.

Pascal-Delange (commented):
Review WIP, here are the comments other than on the core assignment logic

usecases/case_usecase.go
```diff
@@ -1634,3 +1647,25 @@ func (uc *CaseUseCase) PerformCaseActionSideEffects(ctx context.Context, tx repo
...
 	return nil
 }
+
+func (uc *CaseUseCase) triggerAutoAssignment(ctx context.Context, tx repositories.Transaction, orgId string, inboxId uuid.UUID) error {
+	features, err := uc.featureAccessReader.GetOrganizationFeatureAccess(ctx, orgId, nil)
+	if err != nil {
+		return errors.Wrap(err, "could not check feature access")
+	}
+
+	if features.AutoAssignment.IsAllowed() {
```
leo-marble, line 1657: I suggest to early return to improve readability

usecases/inboxes/inboxes.go
```diff
@@ -46,6 +46,16 @@ func (i *InboxReader) GetInboxById(ctx context.Context, inboxId uuid.UUID) (mode
...
+// We do not check for permissions here, since this is partly going to be executed through headless actions.
+func (i *InboxReader) GetAutoAssignmentEnabled(ctx context.Context, inboxId uuid.UUID) (bool, error) {
+	inbox, err := i.InboxRepository.GetInboxById(ctx, i.ExecutorFactory.NewExecutor(), inboxId)
+	if err != nil {
+		return false, err
+	}
+
+	return inbox.AutoAssignEnabled, nil
+}
+
```
leo-marble, line 58: Since this function is very simple and the `AutoAssignEnabled` field is public, do we really need this function? Wouldn't it be clearer to just call `GetInboxById()` and access `AutoAssignEnabled` directly?

api/handle_user_settings.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"time"
+
+	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/gin-gonic/gin"
+)
+
+type UnavailabilityDto struct {
+	Until time.Time `json:"until"`
+}
```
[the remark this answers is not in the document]
Pascal-Delange: not required so let's skip this option IMO

models/organization_feature_access.go
```diff
@@ -47,6 +48,9 @@ func (o OrganizationFeatureAccess) WithTestMode() OrganizationFeatureAccess {
 	if o.NameRecognition == Restricted {
 		o.NameRecognition = Test
 	}
+	if o.AutoAssignment == Restricted {
```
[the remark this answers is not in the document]
apognu: I asked this this morning, and he wants to be granular here.

repositories/auto_assignment_repository.go
```diff
@@ -0,0 +1,67 @@
...
+		select u.*, count(*) filter (where c.id is not null) as case_count
+		from inbox_users iu
+		inner join users u on u.id = iu.user_id
+		left join cases c on c.assigned_to = u.id
+		where
+		  auto_assignable = true and
+		  not exists (
+		    select 1
+		    from user_unavailabilities uu
+		    where uu.user_id = iu.user_id and now() between uu.from_date and uu.until_date
```
Pascal-Delange, line 23: this is a strong indicator that the index you created should not _add_ user_id but use it _instead_ of org_id

repositories/auto_assignment_repository.go
```diff
@@ -0,0 +1,67 @@
...
+			return nil, err
+		}
+
+		user, err := dbmodels.AdaptAssignableUserWithCaseCount(dbUser)
+		if err != nil {
+			return nil, err
+		}
+
+		users = append(users, user)
+	}
```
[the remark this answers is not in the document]
Pascal-Delange: though you are also in the conditions where the helper function SqlToListOfRow works I think ?

api/handle_user_settings.go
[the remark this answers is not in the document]
apognu: That's in the PR description:

> Improve unavailability API: right now, it only asks for an end date and sets the unavailability to [now()-end]. I think we might want to do better than that, even if that is a good first use case. Changing the end date is possible until the period is passed, and cancelling the unavailability moves the end date to now().

usecases/auto_assignment_usecase.go
Pascal-Delange: This seems correct

Pascal-Delange (commented):
Last series of comment from reading the rest of the files (nothing structural in there).

One remaining question, maybe the most important comment I have on this WIP PR: in this workflow, the trigger event "something happens in an inbox", and the action is "dispatch all possible cases that can be assigned in this inbox".

This is subtly different from "try to assign the oldest cases in the org as soon as someone has space available", because the greedy approach in one inbox means that even just one case review in an inbox will try to max out the inbox users's queue, ignoring possibly older cases in another inbox.
Of course I understand this makes the implementation of this usecase simpler, but in this case I think we want to stick closer to the spec even at the cost of some complexity.

repositories/user_settings_repositories.go
```diff
@@ -0,0 +1,53 @@
...
+func (repo *MarbleDbRepository) GetCurrentUnavailability(ctx context.Context, exec Executor, userId string) (*models.UserUnavailability, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+
+	sql := NewQueryBuilder().
+		Select(dbmodels.ColumnsSelectUserUnavailabilities...).
+		From(dbmodels.TABLE_USER_UNAVAILABILITIES).
+		Where(squirrel.Eq{"user_id": userId}).
+		Where(squirrel.Gt{"until_date": time.Now()})
```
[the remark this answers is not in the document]
apognu: Yes, see the description of the PR.

apognu:
I pushed a commit modifying the algorithm to select cases and users across inboxes. The way it works now:

 - Compute available slots across all auto-assignable users of the org (so we can `limit` the case query)
 - Find `limit` oldest cases across all inboxes where auto-assignment is enabled
 - For each case, find the user in its inbox with the least number of case (or in case or a draw, the lowest ID 🤷)
 - If a user was found, assign them to the case, otherwise, go to the next case

repositories/auto_assignment_repository.go
apognu: Yeah I am waiting until we settle on an algorithm to drill into the indices.

repositories/inbox_users_repository.go
```diff
@@ -93,6 +93,7 @@ func (repo *MarbleDbRepository) UpdateInboxUser(ctx context.Context, exec Execut
 		exec,
 		NewQueryBuilder().Update(dbmodels.TABLE_INBOX_USERS).
 			Set("role", role).
+			Set("auto_assignable", autoAssignable).
```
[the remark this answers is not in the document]
Pascal-Delange: 👍 
yes let's make them all like a proper PATCH with everything optional

Pascal-Delange (commented):
Now looks all functionally correct to me !
(except for detail points like reordering migrations, but should be done just before merging ideally).

The one remaining question I have is how those queries will behave at scale if run at every case closing, on every eligible case.
I think I'd like to re-read the queries about this "à tête reposée" later this week

apognu:
They were quite a few filter predicates missing from both SQL queries (omit deleted users and closed cases, for example). I amended them.

apognu:
I started iterating on the queries to analyze their execution of high case count, and propose indexes as comments in the code, for now.

repositories/auto_assignment_repository.go
```diff
@@ -0,0 +1,124 @@
...
+	//  - We filter on users.deleted_at and inbox_users.auto_assignable (though those tables should be quite small)
+	sql := `
+		select u.*, count(distinct c.id) as case_count
+		from inbox_users iu
+		inner join users u on
+		  u.organization_id = $1 and
+		  u.id = iu.user_id and
+		  u.deleted_at is null and
+		  iu.auto_assignable
+		inner join cases c on c.org_id = u.organization_id and c.assigned_to = u.id and c.status != 'closed'
```
Pascal-Delange, line 26: We also need to filter out snoozed cases

repositories/auto_assignment_repository.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
+	"github.com/checkmarble/marble-backend/repositories/dbmodels"
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+)
+
+func (repo *MarbleDbRepository) FindAutoAssignableUsers(ctx context.Context, exec Executor, orgId string, limit int) ([]models.UserWithCaseCount, error) {
+	// TODO: those might require more indices, notably:
+	//  - We join on cases.(org_id, assigned_to, c.status)
```
Pascal-Delange, line 16: I suppose we want an index on (org_id, assigned_to) including (snoozed_at) filtered on non-closed cases only (the filter condition should ensure that the index remains small)

repositories/task_queue_repository.go
```diff
@@ -232,3 +238,29 @@ func (r riverRepository) EnqueueCaseReviewTask(
...
+	orgId string, inboxId uuid.UUID,
+) error {
+	res, err := r.client.InsertTx(
+		ctx,
+		tx.RawTx(),
+		models.AutoAssignmentArgs{
+			OrgId:   orgId,
+			InboxId: inboxId,
+		},
+		&river.InsertOpts{
```
Pascal-Delange, line 254: Last comment: as discussed, let's insert with unicity condition, by queue and with by period (2 min ?), and always insert a little bit in the future (1min ?)

repositories/auto_assignment_repository.go
apognu: As pointed out in the comment at the end of the file (updated now), this index should help:

```(org_id, status != 'closed', assigned_to, coalesce(snoozed_until, to_timestamp(0))) include (id)```

Pascal-Delange (approved):
Looks good, let's merge ! 
Last thing to fix in one query and perhaps a comment

PR #1098, opened 2025-07-23
WIP backfill org id for sanction_checks table

leo-marble:


leo-marble:
I tested on local with 20_000_000 rows, it took around 27min. 
I could use the app without any blocking issue. 

But I expected to see the progression during the migration by fetch rows with an `org_id` but the select query return 0 each time and I needed to wait the end of the migration. 
I removed the transaction generated by goose, I expect each UPDATE query auto commit.

repositories/migrations/20250723112605_backfill_sanction_checks_org_id.sql
```diff
@@ -0,0 +1,55 @@
...
+
+SET log_min_messages = notice;
+
+DO $$
+DECLARE
+    batch_size CONSTANT INTEGER := 1000;  -- Process 1000 records at a time
+    updated_rows INTEGER;
+    total_updated INTEGER := 0;
+    batch_count INTEGER := 0;
+BEGIN
```
apognu, line 16: I think there is a simpler (or at least less bloated path) from `sanction_checks` to an `org_id`: `sanction_checks -> sanction_check_configs -> scenario_iterations`. The two tables to join to are on the smaller side, they should have a limited numbers of rows, which could be way more efficient than whatever we do with `decisions`.

Something along those lines:

```sql
with orgs as (
  select scc.id as scc_id, si.org_id si_org_id
  from sanction_check_configs scc
  inner join scenario_iterations si on si.id = scc.scenario_iteration_id
)
update sanction_checks sc
set org_id = orgs.si_org_id
from orgs
where sc.sanction_check_config_id = orgs.scc_id;
```

Also, are we really worried about load here, how many records are we expecting? How many screenings do we have in production (or what is the volume for our self-hosted customers)? Asking myself if the

apognu: I think we still need to discuss if we really need such a convoluted PR for this amount of data. I don't have numbers, so I'll refer to @Pascal-Delange .

leo-marble: Hum ... my initial thought was to not block too long some rows in case we want to update it (status for example). I was thinking to add a `order by created_at` to update older one first (less chance to be modified by a user). 

Ok I simplified it!

leo-marble:
I simplified the query because we can't have transaction for each batch in a DO block to avoid blocking rows during all migrations

leo-marble:
Tested on my local env, 20_030_035 rows updated in 9min

repositories/migrations/20250723112605_backfill_sanction_checks_org_id.sql
```diff
@@ -0,0 +1,22 @@
...
+    from sanction_check_configs scc
+    inner join scenario_iterations si on si.id = scc.scenario_iteration_id
+)
+UPDATE sanction_checks 
+SET org_id = orgs.si_org_id
+FROM orgs
+WHERE sanction_checks.sanction_check_config_id = orgs.scc_id
+AND sanction_checks.org_id IS NULL;
+
+ALTER TABLE sanction_checks ALTER COLUMN org_id SET NOT NULL;
```
Pascal-Delange, line 15: Just heads up that this will take a (presumably short, but depends on nb of rows) lock on the table and might make other things time out. 
(But it's ok, let's still do it like this)

Pascal-Delange (approved):

PR #1107, opened 2025-07-31
Filter allowed networks

apognu:
This PR implements IP whitelisting to access the backend API endpoints. It introduces a new middleware that retrieves both the organization's whitelist as well as the client IP address and check that they match. It comes in two flavors:

 * As a post-handler middleware: when using it to guard the `/token` endpoint, we need to perform the authentication before we are able to determine the user's organization. Here, we need to buffer the response body until the middleware runs so it is not sent to the client early.
 * As a pre-handler middleware: for most other endpoints, we already know the principal's idendity, so we can perform the check before the handler is run and guard its execution.

This PR assumes the content of `x-real-ip` is safe and that the upstream reverse proxy prevents clients from spoofing it. It will be quite easy to implement on our managed platform, but a disclaimer will need to be provided to customers.

Some items still need to be decided upon, notably for some of the failure conditions: so we fail open or close?

---

The management API has been implemented as a separate endpoint to the update organization one (felt this was critical enough to not be lost in the same endpoint).

Something to keep in mind when we make them is to only allow management of whitelists **if** the user is within the new list of subnets, so they don't cut themselves off.

---

**Note:** as a bonus, this implement a short-circuit in all authentication middlewares that finally return a `404` if the requested route does not exist. No more `401`s on random routes.

### DIFF
--- a/usecases/allowed_networks_usecase.go
+++ b/usecases/allowed_networks_usecase.go
@@ -11,23 +11,23 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-type IpWhitelistUse int
+type AllowedNetworksUse int
 
 const (
-	IpWhitelistLogin IpWhitelistUse = iota
-	IpWhitelistOther
+	AllowedNetworksLogin AllowedNetworksUse = iota
+	AllowedNetworksOther
 )
 
 type ipWhitelistRepository interface {
-	GetOrganizationSubnets(ctx context.Context, exec repositories.Executor, orgId string) ([]net.IPNet, error)
+	GetOrganizationAllowedNetworks(ctx context.Context, exec repositories.Executor, orgId string) ([]net.IPNet, error)
 }
 
-type IpWhitelistUsecase struct {
+type AllowedNetworksUsecase struct {
 	executorFactory executor_factory.ExecutorFactory
 	repository      ipWhitelistRepository
 }
 
-func (uc IpWhitelistUsecase) Guard(use IpWhitelistUse) gin.HandlerFunc {
+func (uc AllowedNetworksUsecase) Guard(use AllowedNetworksUse) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		var buf *utils.BufferResponseWriter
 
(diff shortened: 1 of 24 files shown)


utils/auth.go
```diff
@@ -46,6 +46,13 @@ type Authentication struct {
 
 func (a *Authentication) AuthedBy(methods ...AuthType) gin.HandlerFunc {
 	return func(c *gin.Context) {
+		// If the middleware is executing, but no route was matched, we
+		// short-circuit with a 404.
+		if c.FullPath() == "" {
```
apognu, line 51: This allows to not have 401s when the client requested an unknown route.

usecases/allowed_networks_usecase.go
```diff
@@ -0,0 +1,113 @@
...
+		// proxy, so we fail open if it is not set.
+		if clientIp == nil {
+			if use == AllowedNetworksLogin {
+				buf.Restore(c)
+			}
+
+			return
+		}
+
+		// TODO: we need to discuss about this: even when whitelisting is
```
apognu, line 73: This is an important one to discuss.

I want, before letting the admin configure its allowed networks, to check that they are currently on one of those networks (so they don't cut themselves off). To do that, I would need to rely on the "real IP" header from being present.

That implies the user's infrastructure was properly set up to add that header **and** to block values sent by the client.

The line being at this location has an impact on logs: the content of that header will be the IP address written to the logs, so if infrastructure does not filter client-sent data, it will be spoofable.

---

One solution would be to gate IP whitelisting being (yet another...) configuration flag, which would solve the issue but it not amazing UX-wise.

apognu:
Do we need a special exemption for `MARBLE_ADMIN`s, since those will be called from outside the allowed networks?

apognu:
For reference, we could be using the more standard `x-forwarded-for` header, but this adds the question of how do we let users configure the list / length of trusted proxies, which adds a layer of complexity to the configuration, but is required to make it flexible and secure.

leo-marble:
You need to reorder your migration file before merging into `main`

PR #1108, opened 2025-07-31
feat(case_review): enhance CaseReviewV1 with proofs

leo-marble:
Add a `proofs` object to the case review, enabling the AI agent to reference supporting objects during the review process. 
Include an `ID` field in the case and decision DTOs to allow the agent to precisely identify and refer to these objects.
Use `reaction` in DB to store the user feedback about a review. Handle with 2 different values so far, `ko` and `ok`. This field can be `null` which means no reaction. 

Link to the new prompt can be found in the linear ticket

Link to this PR for frontend https://github.com/checkmarble/marble-frontend/pull/1004


usecases/ai_agent/ai_agent_usecase.go
```diff
@@ -667,7 +667,18 @@ func (uc *AiAgentUsecase) CreateCaseReviewSync(ctx context.Context, caseId strin
...
 		return nil, errors.Wrap(err, "could not prepare case review request")
 	}
-	requestCaseReview, err := llm_adapter.NewUntypedRequest().
+
+	type caseReviewOutput struct {
+		CaseReview string `json:"case_review" jsonschema_description:"The case review analysis including all the information needed to make the review"`
+		Proofs     []struct {
+			Id          string `json:"id" jsonschema_description:"The id of the object used as proof, for organization data model you have an object_id"`
+			Type        string `json:"type" jsonschema_description:"The type of the object used as proof, could be decision or case, for data model from organization, take the trigger_object_type"`
+			IsDataModel bool   `json:"is_data_model" jsonschema_description:"Whether the proof is from organization data model. Decision and case are not data model"`
```
leo-marble, line 676: Not sure about this field. 
For now, data can come from `Marble` or from `Organisation` data model, should I change the type from bool to an enum?

apognu: I think it would be more extendable if this was indeed an enum string. The second we want to add another source of proof (internal or external), this is going to bite us in the backside.

dto/agent_dto/case_review.go
```diff
@@ -36,3 +49,95 @@ func UnmarshalCaseReviewDto(version string, payload io.Reader) (AiCaseReviewDto,
...
 }
+
+// Extend the AiCaseReviewDto with the id and reaction fields for our API endpoints
+// Didn't add in aiCaseReviewDto to avoid adding none related fields of AI agent analysis, also aiCaseReviewDto is
+// stored in blob storage, so we don't need to add it here.
+type AiCaseReviewOutputDto struct {
+	Id       uuid.UUID `json:"id"`
+	Reaction *string   `json:"reaction"`
+
+	AiCaseReviewDto
```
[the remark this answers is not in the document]
apognu: Is this all to avoid having a nested `review` in the root object? This is an internal API and the right moment to break it if necessary, especially if it can avoir having to reimplement JSON marshalling and tags parsing ourselves, in my opinion.

dto/agent_dto/case_review.go
```diff
@@ -15,12 +19,39 @@ type AiCaseReviewDto interface {
...
-	Ok          bool   `json:"ok"`
-	Output      string `json:"output"`
-	SanityCheck string `json:"sanity_check"`
-	Thought     string `json:"thought"`
-	Version     string `json:"version"`
+	Ok          bool              `json:"ok"`
+	Output      string            `json:"output"`
+	SanityCheck string            `json:"sanity_check"`
+	Thought     string            `json:"thought"`
+	Version     string            `json:"version"`
```
apognu, line 53: Let's not serialize the version inside the review itself, since it is present one level up (unless this DTO can be sent on its own).

apognu (approved):

dto/agent_dto/case.go
```diff
@@ -93,10 +95,10 @@ func AdaptCaseEventDto(caseEvent models.CaseEvent, users []models.User) CaseEven
 		EventType: string(caseEvent.EventType),
 		// TODO: Commented out for now because we don't want the AI agent to "cheat" by using the human review to generate a review
 		// while we iterate on the prompts. Final behavior may change but is still undetermined.
-		// AdditionalNote: caseEvent.AdditionalNote,
-		AdditionalNote: "Redacted",
-		NewValue:       caseEvent.NewValue,
-		ResourceType:   string(caseEvent.ResourceType),
+		AdditionalNote: caseEvent.AdditionalNote,
+		// AdditionalNote: "Redacted",
```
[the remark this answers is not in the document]
leo-marble: ping @Pascal-Delange

PR #1116, opened 2025-08-07
chore(db): Remove no more used scenario workflow cols

siiick:
Following the PR #1115 :
Drop old workflow related column from scenario table


apognu (changes requested):
You should change the target orf this PR to be #1115 so the tests pass.

PR #1117, opened 2025-08-08
Custom screening lists support

apognu:
This PR introduces two changes:

 * A configuration knob to configure which scope is used when performing match request. It defaults to `default`, including official lists, but can be changed to, for example, point to a superset scope containing the official and custom lists.
 * Consideration of tags placed directly on local datasets, so custom lists can be classified.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -96,6 +96,10 @@ func RunServer(config CompiledConfig) error {
 		utils.GetEnv("OPENSANCTIONS_API_KEY", ""),
 	)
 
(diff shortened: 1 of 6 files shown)


leo-marble:
It might be helpful to check at startup whether the specified scope exists, and exit if it doesn't (due to a typo or misconfiguration). I'm not sure if that's feasible.

How do we add a custom list to Yente, is it done through Marble or directly in Yente?

apognu:
Let's not make the whole Marble fail just because of a screening misconfiguration.

Everything related to lists is done on Yente.

PR #1122, opened 2025-08-12
Add custom OTEL sampler based on route prefix.

apognu:
This PR implements a probabilistic sampler for our tracing setup, it applies a default rate of `0.3` based on the Trace ID that is overriden by:
 * Statically defined rules based on HTTP route prefix.
 * Static analysis of the SQL queries (to remove `set_config` and duplicate `prepare` statements)

The goal here is to not be overwhelmed by traces from heavily called endpoints (decisions, ingestion, etc.) and to completely disable traces for infrastructure endpoints (such as /liveness).

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -131,6 +131,7 @@ func RunServer(config CompiledConfig) error {
 		sentryDsn                        string
 		transferCheckEnrichmentBucketUrl string
 		firebaseEmulatorHost             string
+		telemetryExporter                string
 		otelSamplingRates                string
 	}{
 		batchIngestionMaxSize:            utils.GetEnv("BATCH_INGESTION_MAX_SIZE", 0),
@@ -143,6 +144,7 @@ func RunServer(config CompiledConfig) error {
 		sentryDsn:                        utils.GetEnv("SENTRY_DSN", ""),
 		transferCheckEnrichmentBucketUrl: utils.GetEnv("TRANSFER_CHECK_ENRICHMENT_BUCKET_URL", ""), // required for transfercheck
 		firebaseEmulatorHost:             utils.GetEnv("FIREBASE_AUTH_EMULATOR_HOST", ""),
+		telemetryExporter:                utils.GetEnv("TRACING_EXPORTER", "gcp"),
 		otelSamplingRates:                utils.GetEnv("TRACING_SAMPLING_RATES", ""),
 	}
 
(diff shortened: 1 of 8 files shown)


leo-marble (approved):

apognu:
The system now takes an optional JSON file that can set sampling ratio according to span names if those do not match one of the other rules (HTTP paths or SQL queries).

cmd/server.go
```diff
@@ -179,6 +183,8 @@ func RunServer(config CompiledConfig) error {
 		ApplicationName: apiConfig.AppName,
 		Enabled:         gcpConfig.EnableTracing,
 		ProjectID:       gcpConfig.ProjectId,
+		Exporter:        serverConfig.telemetryExporter,
+		SamplingMap:     infra.NewTelemetrySamplingMap(ctx, serverConfig.otelSamplingRates),
```
Pascal-Delange, line 187: Let's add the same on the worker please :-)

cmd/server.go
```diff
@@ -142,6 +144,8 @@ func RunServer(config CompiledConfig) error {
 		sentryDsn:                        utils.GetEnv("SENTRY_DSN", ""),
 		transferCheckEnrichmentBucketUrl: utils.GetEnv("TRANSFER_CHECK_ENRICHMENT_BUCKET_URL", ""), // required for transfercheck
 		firebaseEmulatorHost:             utils.GetEnv("FIREBASE_AUTH_EMULATOR_HOST", ""),
+		telemetryExporter:                utils.GetEnv("TRACING_EXPORTER", "gcp"),
```
Pascal-Delange, line 147: Since you're adding it, we might as well make it OTLP by default, and set it to gcp in our hosted environment ?

infra/tracing.go
```diff
@@ -72,3 +93,110 @@ func InitTelemetry(configuration TelemetryConfiguration, apiVersion string) (Tel
...
+
+	// This span should not be sampled if the parent is not. Except for the root
+	// span ID (the one that does not have a trace ID).
+	if psc.HasTraceID() && !psc.IsSampled() {
+		return sdktrace.NeverSample().ShouldSample(p)
+	}
+
+	for _, attr := range p.Attributes {
+		if attr.Key == semconv.HTTPRouteKey {
+			kind = SpanHttpIngress
```
Pascal-Delange, line 148: Similarly, we also have job spans for river jobs (we have a middleware that creates spans)

infra/tracing.go
```diff
@@ -72,3 +93,110 @@ func InitTelemetry(configuration TelemetryConfiguration, apiVersion string) (Tel
...
+		"/health":           0.0,
+		"/liveness":         0.0,
+		"/version":          0.0,
+		"/config":           0.0,
+		"/is-sso-available": 0.0,
+		"/signup-status":    0.0,
+		"/decisions":        0.05,
+		"/v1/decisions":     0.05,
+		"/ingestion":        0.05,
+		"/v1/ingest":        0.05,
```
Pascal-Delange, line 118: Let's also add the high volume river jobs here, for instance "async_decision" is a high volume span name, as are "match_enrichment" and "test_run_summary".

apognu: Added a separate default map that filters on span names instead of HTTP paths. For symmetry, I added the same two options to the sampling config file.

Pascal-Delange (approved):
very good, thanks 👍

PR #1125, opened 2025-08-13
feat(case_review): failure mechanism, resume at the last succeed step

ghost:
A failure mechanism has been added to prevent unnecessary LLM calls. 

If the `CreateCaseReviewSync` method fails, all LLM results are stored in a new `CaseReviewContext` struct, which is passed by reference. The caller is responsible for creating, populating, and storing the contents of this struct in a file to be fetch in case it need to use it.

The process for creating an `AiCaseReviewFile` has been refactored: previously, the database resource was only created after a successful job iteration. Now, the row is created when the job is enqueued, and the `AiCaseReviewFile.Id` is passed so the job knows which data to process. 

Additionally, a new property has been added to reference a temporary file containing the `CaseReviewContext` data.

If the case review job fails, the worker stores the data in a file. The next iteration can proceed by reusing the same arguments provided by the caller.


usecases/ai_agent/async_case_review.go
```diff
@@ -74,56 +84,80 @@ func (w *CaseReviewWorker) Timeout(job *river.Job[models.CaseReviewArgs]) time.D
...
+	if err != nil {
+		errUpdate := w.repository.UpdateCaseReviewFile(ctx, w.executorFactory.NewExecutor(),
+			aiCaseReview.Id, models.UpdateAiCaseReview{
+				Status: models.AiCaseReviewStatusFailed,
+			})
+		if errUpdate != nil {
+			return errors.Join(
+				errors.Wrap(errUpdate, "Error while updating case review file status"),
+				errors.Wrap(err, "Error while generating case review"))
+		}
```
leo-marble, line 119: I think I can extract into a function, I repeated too many time

apognu (approved):

PR #1130, opened 2025-08-20
feat(migrations): add migration to set file_temp_reference as NOT NULL

leo-marble:
fix: https://marble-eu.sentry.io/issues/59004707/?alert_rule_id=193450&alert_type=issue&notification_uuid=94bec797-d449-4b8b-a91e-136ea703cf8e&project=4509310347640912&referrer=slack 

`file_temp_reference` column was set as nullable column but the data model in code expect a non null value. 

I changed the nullable setting of this column and fill a default value for existing row. 

We don't have many rows so the migration should be ok and fast


siiick (approved):
👍

PR #1198, opened 2025-09-16
Offloading: concurrent GCS upload to improve performance.

apognu:
We have seen poor performance in the initial version of the offloading job when writing to GCS serially. This PR batches decision rules to offload into batches and write each of those concurrently up to the save point size.

### DIFF
--- a/repositories/blob_repository.go
+++ b/repositories/blob_repository.go
@@ -65,10 +65,10 @@ func (repository *blobRepository) openBlobBucket(ctx context.Context, bucketUrl 
 	)
 	defer span.End()
 
-	if repository.buckets[bucketUrl] == nil {
-		repository.m.Lock()
-		defer repository.m.Unlock()
+	repository.m.Lock()
+	defer repository.m.Unlock()
 
+	if repository.buckets[bucketUrl] == nil {
 		var bucket *blob.Bucket
 		// adapt bucket url with additional values from env variables in the GCS case
 		url, err := url.Parse(bucketUrl)
--- a/usecases/scheduled_execution/offloading_job.go
+++ b/usecases/scheduled_execution/offloading_job.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/storage"
@@ -153,6 +154,7 @@ func (w OffloadingWorker) Work(ctx context.Context, job *river.Job[models.Offloa
 		// Slice of pointers to not have to keep track of which decision were skipped or not, the
 		// query ignores NULL IDs.
 		offloadedIds := make([]*string, w.config.SavepointEvery)
+		rulesToOffload := make([]models.OffloadableDecisionRule, w.config.SavepointEvery)
 		idx := 0
 
 		var lastOfBatch *models.OffloadableDecisionRule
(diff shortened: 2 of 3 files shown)


repositories/blob_repository.go
```diff
@@ -65,10 +65,10 @@ func (repository *blobRepository) openBlobBucket(ctx context.Context, bucketUrl
 	)
 	defer span.End()
 
-	if repository.buckets[bucketUrl] == nil {
-		repository.m.Lock()
-		defer repository.m.Unlock()
+	repository.m.Lock()
```
apognu, line 68: A quick note on this change.

I noticed lock contention when multiple goroutine were trying to open a bucket that had not been initialized before. In our case, 200 goroutine tried to acquire the lock **after** having checked that it was `nil`, repeating the costly initialization setup down there.

For offloading, it meant that the first batch would take up to 15 seconds instead of 1.

I am not entirely happy with this, since now we get contention of every access, even though we safeguard against initializing everything again (all except the first one should be short enough, as opposed to the previous iteration, where all concurrent 200 would be slow and serialized).

usecases/scheduled_execution/offloading_job.go
```diff
@@ -261,10 +230,17 @@ func (w OffloadingWorker) Work(ctx context.Context, job *river.Job[models.Offloa
...
 		// items to persist.
 		if idx%w.config.SavepointEvery > 0 {
-			remainingItems := offloadedIds[:idx%w.config.SavepointEvery]
+			span.AddEvent("partial_savepoint_reached")
+
+			remainingItemIds := offloadedIds[:idx%w.config.SavepointEvery]
+			remainingRules := rulesToOffload[:idx%w.config.SavepointEvery]
+
+			if err := w.writeBatchToBlobStorage(ctx, req, remainingRules, remainingItemIds); err != nil {
+				return stopChannelAndReturn(err)
```
Pascal-Delange, line 239:
```suggestion
				return err
```
Just this should be good here (my "fix" on channeling errors back to the rule streaming goroutine I think only needs to be used for errors within the loop on rules

Pascal-Delange: (though I don't think this panics or anything)

usecases/scheduled_execution/offloading_job.go
```diff
@@ -307,3 +283,86 @@ func (w OffloadingWorker) Work(ctx context.Context, job *river.Job[models.Offloa
 		span.End()
 	}
 }
+
+// This will write all objects within the save point to GCS concurrently.
+// Encountering an error will not immediately stop the batch and will continue
+// writing the rest of the objects. The last error to be encountered will be
+// returned.
```
Pascal-Delange, line 290: for what it's worth, we could cancel the context if any error happens ? Or is that overkill ?

Pascal-Delange (approved):

usecases/scheduled_execution/offloading_job.go
apognu: Technically, we could. Now, all the writes happen within the same second, and will most probably all have started/finished already whe the error happens. We could cancel the context, but that isn't going to be a meaningful gain over finishing the writes.

I would have the opposite opinion if we were transferring large files, where cancelling early could be a net gain in terms of time or pricing, but I feel that it will only add complexity, here.

Happy to discuss.

PR #1199, opened 2025-09-17
Code cleanup

Pascal-Delange:
Remove the deprecated entrypoints - leaves only --migrations, --server and --worker.
This is a breaking change, though not one that should impact anyone.

Also rename api/server.go to api/http_server.go so I don't get to it every time I try to "jump to server" in my IDE.


apognu (approved):
Image

PR #1213, opened 2025-09-24
feat: add DecisionWorkflowsWorker

OrriMandarin:
- Introduced a new worker for processing decision workflows
- Updated task queue repository to support enqueuing decision workflow tasks.
- Implemented logic for handling decision workflows in the usecases layer by replacing the process by enqueue function
- The behavior for creating decisions asynchronously in batch mode remains unchanged.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -39,7 +39,7 @@
       "args": [
         "--worker",
         "--worker-only=${input:jobName}",
-        "--worker-args=${input:jobArgs}",
+        "--worker-args=${input:jobArgs}"
       ],
       "console": "integratedTerminal"
     }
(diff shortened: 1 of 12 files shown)


OrriMandarin:
~~The unit test failed~~ The test fails on my machine, likely because I have multiple river workers running, whereas CI only uses one—since the case is now created asynchronously after the decision. To work around this, I added a time.Sleep after creating the decision and then retrieved the updated decision data from the database.

However, there's still an issue with this refactor (similar to Peepal): if multiple decisions are created at the same time, the workflow doesn't detect cases created by other decisions and can't associate a decision with a case that has the same pivot.

The transition to using River jobs works, but the logic for determining whether a case should be created in the worker needs to be revised.

I recommend reviewing this rework as it stands. I'll address the race condition related to case creation in an other Pull Request based on this PR.

apognu (approved):

PR #1215, opened 2025-09-25
fix: use Serializable isolation when executing workflow

OrriMandarin:
- For async decisions, workflow execution now occurs in a separate river task. This centralizes the logic and avoids repeatedly setting the transaction isolation level when processing decision workflows.
- The transaction attempt mechanism is now accurate: the retry attempt value reflects the total number of executions, not retries. For instance, `Attempt = 1` means the function runs only once, even if an error occurs.
- When handling decision workflows, transactions use the Serializable isolation level to prevent race conditions—especially for the `add to case if possible` action. The race condition happens when the system reads for a matching case and creates a new one if none exists.

With Serializable isolation, if two workflows try to create the same case at the same time in `add to case if possible` mode, only the first succeeds; the second fails then retry the function and attaches the decision to the existing case.


OrriMandarin:
Hum ... 

I attempted a batch decision execution, but some decision workflows failed with the following error:
```
error processing decision workflows: Error executing transaction: error processing decision workflows: error while executing workflow action: error adding decision to open case: error updating case: error executing sql query: UPDATE decisions SET case_id = $1 WHERE id IN ($2): ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
```

It is fine because we use river and a transaction retry mechanism, but it does mean some tasks may fail.

repositories/db_executor_getter.go
```diff
@@ -93,7 +93,7 @@ func (g ExecutorGetter) Transaction(
 	}
 	// Retry retryable commit errors once, immediately after failure.
 	err = retry.Do(execInTransaction,
-		retry.Attempts(1),
+		retry.Attempts(2),
```
Pascal-Delange, line 96: let's close this PR but still keep this fix ?

OrriMandarin:
Not the right solution, lot of fails due to this isolation level (tight relation between case and decision). 
I close this PR since we have an other solution to fix the root cause: https://github.com/checkmarble/marble-backend/pull/1216

PR #1216, opened 2025-09-26
feat: implement advisory lock functionality in db_lock.go

OrriMandarin:
- Added a new file `db_lock.go` to manage PostgreSQL advisory locks.
- Implemented functions to acquire and release both transaction-level and session-level advisory locks.
- Updated `actions_case.go` to utilize advisory locks for preventing race conditions when processing decisions with the same pivot value.


usecases/decision_workflows/actions_case.go
```diff
@@ -94,6 +96,20 @@ func (d DecisionsWorkflows) AutomaticDecisionToCase(
 	case models.WorkflowCreateCase:
 		return createNewCaseForDecision(ctx)
 	case models.WorkflowAddToCaseIfPossible:
+		// Get an advisory lock on pivot value to prevent race conditions when multiple decisions
+		// for the same entity (user, account, etc.) arrive simultaneously. This ensures that
+		// AddToCaseIfPossible operations are serialized and only one case gets created per pivot value.
+		// If pivot value is nil, no lock is needed since we'll always create a new case.
+		if decision.PivotValue != nil {
```
apognu, line 103: Shouldn't we take the lock on `(pivot_id, pivot_value)` instead? Two decision can have the same pivot value without being related at all (they could even be from another org).

usecases/scheduled_execution/async_decision_job.go
```diff
@@ -398,19 +388,12 @@ func (w *AsyncDecisionWorker) createSingleDecisionForObjectId(
 	}
 	sendWebhookEventId = append(sendWebhookEventId, webhookEventId)
 
-	workflowRules, err := w.repository.ListWorkflowsForScenario(ctx, tx, uuid.MustParse(scenario.Id))
+	err = w.taskQueueRepository.EnqueueDecisionWorkflowTask(ctx, tx,
```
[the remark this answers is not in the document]
OrriMandarin: I didn’t migrate the async decision flow because it already runs in an async task. Now this change is mainly to isolate that path and align its behavior with the new locking approach.

usecases/decision_workflows/actions_case.go
OrriMandarin: Done, I make the lock key with pivot_value, pivot_id and inbox_id (with a default value if the inbox_id is not provided (AnyInbox)

PR #1219, opened 2025-09-29
Harden analytics filter detection, prevent error on empty set.

apognu:
An error was encountered when a table had exported with no exported fields (unpivot would error out on not having any column to pivot on), so this PR adds the dummy `id` column that is always present).

At the same time, adding a sanity check on array slicing (even if if should always be the case).


usecases/analytics_metadata_usecase.go
```diff
@@ -50,15 +50,20 @@ func (uc AnalyticsMetadataUsecase) GetAvailableFilters(ctx context.Context, req
 	//
 	// Each of those columns is then joined with the table definition to add the
 	// column type.
+	//
+	// It requires that the targetted table has an "id" columns to be used as a
```
Pascal-Delange, line 54:
```suggestion
	// It requires that the targeted table has an "id" column to be used as a
```

Pascal-Delange (approved):

PR #1230, opened 2025-10-01
fix: add missing field in public openapi

OrriMandarin:
This pull request updates the OpenAPI schema to include missing fields: **nationality** for Person and **country** for Company. These fields are already supported by the endpoint.


apognu (approved):
Let's check with Pascal, but this is probably a good candidate to be cherry-picked on the hotfix branch.

PR #1231, opened 2025-10-01
Case mass-update endpoint.

apognu:
This PR implements an API endpoint to update select case attributes for a large number of cases. So far, the supported actions are:

 * Close cases
 * Reopen cases
 * Assign cases to user
 * Change cases' inbox

This is a work in progress. Notably, some heavy optmization must be done form this first draft to avoid hammering the database for all cases in the request.

### DIFF
--- a/usecases/case_usecase.go
+++ b/usecases/case_usecase.go
@@ -72,6 +72,7 @@ type CaseUseCaseRepository interface {
 
 	UserById(ctx context.Context, exec repositories.Executor, userId string) (models.User, error)
 
+	GetMassCasesByIds(ctx context.Context, exec repositories.Executor, caseIds []uuid.UUID) ([]models.Case, error)
 	CaseMassChangeStatus(ctx context.Context, tx repositories.Transaction, caseIds []uuid.UUID, status models.CaseStatus) ([]uuid.UUID, error)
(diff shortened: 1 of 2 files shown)


usecases/case_usecase.go
```diff
@@ -1663,3 +1669,143 @@ func (uc *CaseUseCase) triggerAutoAssignment(ctx context.Context, tx repositorie
...
+
+		newAssignee, err = uc.repository.UserById(ctx, exec, req.Assign.AssigneeId.String())
+		if err != nil {
+			return errors.Wrap(err, "target user for assignment not found")
+		}
+	}
+
+	// For all cases in the mass update, we need to check the current user can manage them.
+	for _, caseId := range req.CaseIds {
+		// TODO: that is harsh, refactor GetCaseById into a single SQL query to
```
Pascal-Delange, line 1699: agreed let's add that 👍

Pascal-Delange (commented):
Then there is the nastier question of case update side effects.
Currently those include:
- assign to self (if unassigned)
- unboost (if boosted)
- unsnooze (if snoozed)
- set status to investigating (if pending)

In this context (batch action, most likely from an admin, in any case not by someone who is specifically trying to work on "that one case"), I think there is an argument for doing 2 and 3 but not 1 and 4.

apognu:
Oh yes, the side effects are on my to-do, pending discussion on what we want to do.

apognu:
As settled on Slack, for now, the only side effect we want is unboosting the affected cases. This is done.

PR #1264, opened 2025-10-17
update linter config

Pascal-Delange:
I was getting annoyed by having loads of non-blocking linter warnings in my cursor (because Cursor seems to not properly pick up some details of our .golangci.yml configuration).
So I tried to fix them, and going down the rabbit hole also made some changes to the lint config. Let me know if you find some of those too opinionated.

Also make CI lint output verbose to try and see why it's so damn slow

### DIFF
--- a/.github/workflows/backend_test_workflow.yaml
+++ b/.github/workflows/backend_test_workflow.yaml
@@ -46,6 +46,9 @@ jobs:
         with:
           go-version-file: ./go.mod
 
+      - name: Build
+        run: go build -v ./...
+
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:


models/workflows.go
```diff
@@ -30,11 +30,11 @@ type WorkflowConditionType string
 
 const (
 	WorkflowConditionUnknown   WorkflowConditionType = "unknown"
-	WorkflowConditionAlways                          = "always"
```
Pascal-Delange: while this was not fundamentally broken (the code compiled and worked), the custom type WorkflowConditionType was only used for the first object.

apognu:
I don't really have a strong opinion (other than verbose changes like my `uc` to `usecase` and such which I find cluttery 😅), my main concern, not for this PR but overall is that I don't use Cursor, so I won't get those "lints" in the future, so those will most probably appear again.

usecases/data_model_usecase.go
```diff
@@ -38,7 +38,7 @@ type dataModelUsecaseIngestedDataReadRepo interface {
 		orgId string) ([]models.FieldStatistics, error)
 }
 
-type DataModelUseCase struct {
+type usecase struct {
```
apognu: This is a bit surprising. I guess it's the only usecase in the package, but it's really not explicit.

apognu (commented):
Added one comment, and an overall impression.

Surprised it did not report changing all the `interface{}` to `any`, which pop up in my editor's

But after this is merged, I'll try to figure out a way for it to be editor-agnostic (lint and format) , so we can guarantee a unique style.

usecases/data_model_usecase.go
Pascal-Delange: oh, that looks like a accidental refactoring mishap. Definitely not intended.

Pascal-Delange:
What is the thing about interface{} vs any ? 
Is this something just your IDE is showing ?

PR #1266, opened 2025-10-17
Pascal/faster write read

Pascal-Delange:
## Goals
- avoid losing time and wasting DB ressources in the critical path:
   - only one reading of decision rules and screenings
   - or, actually no reading of decision rules and screenings, we know them before writing them ?
- make the duration logs make sense again in the "many decisions" case
- some minor tracing & logging changes

### DIFF
--- a/usecases/screening_usecase_test.go
+++ b/usecases/screening_usecase_test.go
@@ -46,21 +46,6 @@ func TestListScreeningOnDecision(t *testing.T) {
 	uc, exec := buildScreeningUsecaseMock()
 	sccId := uuid.NewString()
 
-	sccRow, mockSccRow := utils.FakeStruct[dbmodels.DBScreeningConfigs](
-		ops.WithCustomFieldProvider("Id", func() (interface{}, error) {
-			return sccId, nil
-		}),
-		ops.WithCustomFieldProvider("TriggerRule", func() (interface{}, error) {
-			return []byte(`{}`), nil
-		}),
-		ops.WithCustomFieldProvider("Query", func() (interface{}, error) {
-			return []byte(`{}`), nil
-		}),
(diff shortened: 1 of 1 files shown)


usecases/decision_usecase.go
```diff
@@ -549,9 +552,17 @@ func (usecase *DecisionUsecase) CreateAllDecisions(
 			usecase.scenarioEvaluator.EvalScenario(ctx, evaluationParameters)
 		switch {
 		case err != nil:
-			return nil, 0, errors.Wrap(err, "error evaluating scenario in CreateAllDecisions")
+			return nil, 0, errors.Wrapf(err, "error evaluating scenario \"%s\" in CreateAllDecisions", scenario.Name)
```
apognu: Change the string to delimited by backticks instead, way more readable and less prone to escaping issues.

apognu (approved):

models/decision.go
```diff
@@ -140,8 +142,23 @@ func AdaptScenarExecToDecision(scenarioExecution ScenarioExecution, clientObject
 			ScheduledExecutionId: scheduledExecutionId,
 			Score:                scenarioExecution.Score,
 		},
-		RuleExecutions:      scenarioExecution.RuleExecutions,
-		ScreeningExecutions: scenarioExecution.ScreeningExecutions,
+		RuleExecutions: scenarioExecution.RuleExecutions,
+		ScreeningExecutions: pure_utils.Map(scenarioExecution.ScreeningExecutions,
+			MergeScreeningExecWithDefaults(decisionId, scenarioExecution.OrganizationId)),
+	}
```
OrriMandarin: Instead of setting defaults in code, what about using a SQL INSERT with a returning clause to populate the struct with the actual persisted values? I don’t expect returning to be expensive.

PR #1268, opened 2025-10-21
return some new fields on screenings, change some error handling

Pascal-Delange:
## Contents
- exposes the screening rule name and the error codes (if any) in the screenings public API
- modify some error handling details during screening execution: see comment inline

### DIFF
--- a/pubapi/openapi/v1.yml
+++ b/pubapi/openapi/v1.yml
@@ -1023,6 +1023,7 @@ components:
         - status
         - query
         - match_count
+        - name
         - created_at
         - updated_at
       properties:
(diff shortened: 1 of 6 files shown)


usecases/decision_usecase.go
```diff
@@ -600,7 +600,7 @@ func (usecase *DecisionUsecase) CreateAllDecisions(
 				Observe(time.Since(decisionStart).Seconds())
 
 			for _, sce := range item.decision.ScreeningExecutions {
-				err := usecase.screeningRepository.InsertScreening(ctx, tx, sce)
+				err = usecase.screeningRepository.InsertScreening(ctx, tx, sce)
```
apognu: Is that an effective functional change? Because from what I see, it doesn't change anything (it is also redefined in an inner scope and after the current scope.

apognu (approved):

OrriMandarin (approved):

PR #1276, opened 2025-10-24
Add Redis executor.

apognu:
This PR adds a Redis executor to our usecases. Here's the usage:

```go
cache := usecase.redisClient.NewExecutor(decision.OrganizationId.String())

// Set a scalar value (or really execute any Redis op).
err := cache.Exec(func(c *redis.Client) error {
	return c.Set(ctx, cache.Key("user"), "apognu", 0).Err()
})

// Obtain a scalar value. Return type is generic.
// `cache.Key(string)` scopes a key to an organization
username, err := repositories.RedisQuery(cache, func(c *redis.Client) (string, error) {
	return c.Get(ctx, redisExecutor.Key("user")).Result()
})

// Load a JSON-marshaled model
if cachedDecision, err := repositories.RedisLoadModel[models.DecisionWithRuleExecutions](ctx, cache, redisExecutor.Key("thedecision")); err == nil {
	return cachedDecision, nil
}

// business logic

// Save a JSON-marshaled model
if err := cache.SaveModel(ctx, cache.Key(decision.id), decision, 0); err != nil {
	return models.DecisionWithRuleExecutions{}, err
}

return decision, nil
```

There are other data structures that I would like to explore, depending on use cases:

* Redis lists (for custom lists for example)
* Redis hashes (useful when we retrieve exactly one column from a table row, such as "is this entitlement enabled?", even we would need to change the way we use it)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/TwiN/deepmerge v0.2.2
 	github.com/adhocore/gronx v1.19.6
 	github.com/adrg/strutil v0.3.1
+	github.com/alicebob/miniredis/v2 v2.36.1
 	github.com/authenticvision/rgeo v1.4.0
 	github.com/avast/retry-go/v4 v4.7.0
 	github.com/biter777/countries v1.7.5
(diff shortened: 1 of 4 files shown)


Pascal-Delange (approved):

Pascal-Delange (approved):
let's try this 👍

PR #1283, opened 2025-10-30
Add table to store monitored ingested data

OrriMandarin:
Create a per-data model table in the client schema storing `object_id` and `config_id` to enable monitoring based on its associated configuration.

Support two monitoring modes:
- Providing the object ID.
- Providing a payload that is ingested first, then added to the monitoring list.

### DIFF
--- a/dto/screening_monitoring_config.go
+++ b/dto/screening_monitoring_config.go
@@ -61,7 +61,7 @@ func (dto CreateScreeningMonitoringConfigDto) Validate() error {
 	if dto.MatchLimit < 1 {
 		return errors.Wrap(
 			models.BadParameterError,
-			"match limit must be greater than or equal to 0",
+			"match limit must be at least 1",
 		)
 	}
 
@@ -98,7 +98,7 @@ func (dto UpdateScreeningMonitoringConfigDto) Validate() error {
 	if dto.MatchLimit != nil && *dto.MatchLimit < 0 {
 		return errors.Wrap(
 			models.BadParameterError,
-			"match limit must be greater than or equal to 0",
+			"match limit must be at least 1",
 		)
 	}
 
--- a/usecases/screening_monitoring/object_monitoring_operation.go
+++ b/usecases/screening_monitoring/object_monitoring_operation.go
@@ -105,13 +105,14 @@ func (uc *ScreeningMonitoringUsecase) InsertScreeningMonitoringObject(
 		)
 	})
 
+	if repositories.IsUniqueViolationError(err) && ignoreConflictError {
+		return nil
+	}
 	if repositories.IsUniqueViolationError(err) {
-		// If the object already exists in the screening monitoring table, we can ignore the conflict error
-		// in case of ingestion. Consider the object as a new one and force the screening on the updated object.
-		if ignoreConflictError {
-			return nil
-		}
-		return models.ConflictError
+		return errors.Wrap(
+			models.ConflictError,
+			"object already exists in screening monitored objects table",
+		)
 	}
 	return err
 }
--- a/usecases/screening_monitoring/object_monitoring_operation_test.go
+++ b/usecases/screening_monitoring/object_monitoring_operation_test.go
@@ -487,7 +487,7 @@ func (suite *ScreeningMonitoringUsecaseTestSuite) TestInsertScreeningMonitoringO
 
 	// Assert - should error when ignoreConflictError is false and unique violation occurs
 	suite.Error(err)
-	suite.Equal(models.ConflictError, err)
+	suite.Contains(err.Error(), "object already exists in screening monitored objects table")
 	suite.AssertExpectations()
 }
 


dto/screening_monitoring_config.go
```diff
@@ -102,3 +103,17 @@ func AdaptUpdateScreeningMonitoringConfigDtoToModel(dto UpdateScreeningMonitorin
 		Enabled:        dto.Enabled,
 	}
 }
+
+type InsertScreeningMonitoringObjectDto struct {
+	TableId  string    `json:"table_id" binding:"required"`
```
OrriMandarin: Should we reference the table by its ID from ‎`data_model_tables`, or use the table name, which is expected to remain unchanged after creation?

api/routes.go
```diff
@@ -194,6 +194,7 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 	router.PATCH("/screening-monitoring/configs/:config_id", tom,
 		handleUpdateScreeningMonitoringConfig(uc),
 	)
+	router.POST("/screening-monitoring/objects", tom, handleInsertScreeningMonitoringObject(uc))
```
Pascal-Delange: we can keep it here as long as we are developing, but in the end this should go to public api v1, with the proper conventions used

api/handle_screening_monitoring.go
```diff
@@ -116,3 +116,30 @@ func handleUpdateScreeningMonitoringConfig(uc usecases.Usecases) func(c *gin.Con
...
+	return func(c *gin.Context) {
+		ctx := c.Request.Context()
+
+		var input dto.InsertScreeningMonitoringObjectDto
+		if err := c.ShouldBindJSON(&input); err != nil {
+			presentError(ctx, c, errors.Wrap(models.BadParameterError, err.Error()))
+			return
+		}
+		if err := input.Validate(); err != nil {
+			presentError(ctx, c, errors.Wrap(models.BadParameterError, err.Error()))
```
Pascal-Delange: minor thing, but wrapping with BadParamError IMO should rather be done where we create the error than in the handler

repositories/screening_monitoring_client_repository.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+)
+
+const TABLE_INTERNAL_SCREENING_MONITORING = "_screening_monitoring_objects"
```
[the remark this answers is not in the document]
Pascal-Delange: It is in fact validated on the backend server. Not its length, however.

repositories/screening_monitoring_client_repository.go
```diff
@@ -0,0 +1,77 @@
...
+
+func tableNameWithPrefix(tableName string) string {
+	return fmt.Sprintf("%s_%s", TABLE_INTERNAL_SCREENING_MONITORING, tableName)
```
Pascal-Delange: Postgres max table name length is 63, this may break in unexpected ways in practice.
I'm afraid we need something a bit more cumbersome.
Maybe `{table_name_first_10_chars}_{nb}_{suffix}` where "nb" is a number (incremental) used for tie-breaking in case of name clashing.
It's a PITA but then again it would probably be one method encapsulating all of this logic.

api/routes.go
OrriMandarin: Yes, I plan to open a separate PR for the public API once the implementation is complete.

usecases/screening_monitoring/object_monitoring_operation.go
```diff
@@ -0,0 +1,157 @@
...
+		ignoreConflictError = true
+	} else if input.ObjectId != nil {
+		objectId = *input.ObjectId
+	} else {
+		// Should never happen if the input is validated
+		return errors.New("object_id or object_payload is required")
+	}
+
+	// Check if the object exists in ingested data then insert it into screening monitoring table
+	// Create if not exists the screening monitoring table and index
```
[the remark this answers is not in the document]
OrriMandarin: We could also avoid ingesting the data ourselves and require the user to handle ingestion before adding the object to the monitored list.

The concern is that if something fails mid-process, we won’t know whether ingestion completed or which version ended up in the marble DB.

repositories/screening_monitoring_client_repository.go
```diff
@@ -0,0 +1,77 @@
...
+// uniq_idx_config_id_object_id: UNIQUE INDEX, config_id and object_id must be unique
+// TODO: Should we add the index creation with the table creation?
+func (repo *ClientDbRepository) CreateInternalScreeningMonitoringIndex(ctx context.Context, exec Executor, tableName string) error {
+	if err := validateClientDbExecutor(exec); err != nil {
+		return err
+	}
+
+	// Unique index to have a unique object_id for a given config_id
+	sql := fmt.Sprintf(
+		"CREATE UNIQUE INDEX IF NOT EXISTS uniq_idx%s_config_id_object_id ON %s (config_id, object_id)",
```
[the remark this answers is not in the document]
OrriMandarin: But I don't enforce the transaction in the function

usecases/screening_monitoring/object_monitoring_operation.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if repositories.IsUniqueViolationError(err) {
+		// If the object already exists in the screening monitoring table, we can ignore the conflict error
+		// in case of ingestion. Consider the object as a new one and force the screening on the updated object.
+		if ignoreConflictError {
+			return nil
+		}
+		return models.ConflictError
+	}
+	return err
```
Pascal-Delange:
```suggestion
	if repositories.IsUniqueViolationError(err) && ignoreConflictError{
		// If the object already exists in the screening monitoring table, we can ignore the conflict error
		// in case of ingestion. Consider the object as a new one and force the screening on the updated object.
		return nil
	}
	return errors.Wrap(models.ConflictError, err.Error())
```

usecases/screening_monitoring/object_monitoring_operation.go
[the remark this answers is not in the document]
Pascal-Delange: As discussed on Wed, I still think ingestion should be handled here, in any case not doing it does not make things any simpler since the list of monitored objects is still maintained in the client DB.

But we can do no in another PR.

Pascal-Delange (approved):
This PR looks ok to me to merge, with remaining comments that are probably better suited to be integrated as follow-up PRs with smaller scope.

PR #1289, opened 2025-11-02
Disable seqscan on DuckDB's Postgres connections directly.

apognu:
DuckDB manages its own connection pooling to Postgres, meaning its own sessions. We therefore cannot set local variables when initiating the first connection (they were not inherited by subsequent sessions).

This PR sets the relevant session variables (notably `enable_seqscan=0`) on the client directly (by amending the DSN).


Pascal-Delange (approved):
:bingo: as I like to say

PR #1294, opened 2025-11-04
fix query: can again export custom fields for filters from decision r…

Pascal-Delange:
Tentative fix for slow export queries. With this, on the cases I have tested we do not need to force "no seqscan, no hash join". I'm still leaving the "no seqscan" for now, out of prudence.
Also, with those settings I have tested that we can reliably query decision rules (as per the analytics export query) with batch size of 100k+, perhaps even up to 500k (on decisions, so that's more decision rules).

Tested on the S&P prod org.


Pascal-Delange:
Actually, it should not block I think 🤔  especially if this is the last migration in the release

PR #1295, opened 2025-11-04
Add whether the backend runs an outdated version in /config.

apognu:
This fetched the list of Github releases for the backend in the background and populates whether we are outdated according to the following heuristic:

 - The current and latest releases are more than two minors apart AND
 - The current and latest releases are more than a month apart.

---

Let's discuss what we consider is an outdated version to pinpoint the lofic.


apognu:
Out of completeness, I checked the rate limits on the GitHub API's public API, and it's 60 requests per hour per IP address. So even imagining that a customer deploys all it's instances behind the same IP (very possible), it could scale to 60 instances of the backend.

In any case, this PR, right now, fails gracefully by not reporting itself as outdated on errors.

PR #1305, opened 2025-11-07
Add an entrypoint to run a standalone analytics server.

apognu:
This server will initialize the same way as the backend, but will only mount misc. routes (healthchecks, version) and the analytics routes.

A new environment variable, on the regular backend, can indicate the base URL to that analytics server, and will proxy analytics requests to that URL. All relevant elements from the source requests are copied (path, query, headers). The reponse will experience the same treatment, where the status, header and body will be copied as is.


Pascal-Delange (approved):
looks good 👍

PR #1307, opened 2025-11-07
Add case queries for analytics dashboard (PG).

apognu:
For a first version, the analytics queries for cases are performed on PostgreSQL (instead of DuckDB) with a hardcoded timeframe of 10 days.

It adds two queries:

 - case_status_by_date
 - case_status_by_inbox

And returns the count of cases in four categories:

 - pending
 - investigating
 - snoozed
 - closed

Both queries are implicitly scoped to the inboxes available to the user.

---

> [!NOTE]
> Adds Postgres-backed `case_status_by_date` and `case_status_by_inbox` analytics, scoped by org and available inboxes, and allows empty request bodies for analytics queries.
> 
> - **API**
>   - `handleAnalyticsQuery`: accept empty JSON body; extract org ID and set `filters.OrgId`; wire new queries `case_status_by_date` and `case_status_by_inbox`.
> - **Usecases**
>   - Implement `AnalyticsQueryUsecase.CaseStatusByDate` and `CaseStatusByInbox` using Postgres with a 10-day window, timezone offset handling, and inbox scoping via `InboxReader`.
>   - Add `getAvailableInboxIds`; inject `inboxReader` in `NewAnalyticsQueryUsecase`.
> - **DTO/Models**
>   - Extend `dto.AnalyticsQueryFilters` with `OrgId`.
>   - Add result structs `analytics.CaseStatusByDate` and `analytics.CaseStatusByInbox`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2e6124fc2b4104b79afeba9542a020685c1d1e04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/usecases/analytics_query_usecase.go
+++ b/usecases/analytics_query_usecase.go
@@ -378,38 +378,33 @@ func (uc AnalyticsQueryUsecase) CaseStatusByInbox(ctx context.Context,
 		return nil, err
 	}
 
-	cte := repositories.WithCtes("case_statuses", func(b squirrel.StatementBuilderType) squirrel.SelectBuilder {
-		return b.Select(
-			"i.name as inbox",
-			"c.status",
-			"snoozed_until is not null and snoozed_until > now() and boost = 'unsnoozed' as snoozed",
-		).
-			From(dbmodels.TABLE_CASES+" c").
-			InnerJoin(dbmodels.TABLE_INBOXES+" i on i.id = c.inbox_id").
-			Where("org_id = ?", filters.OrgId).
-			Where(squirrel.Eq{"inbox_id": inboxes}).
-			Where("c.created_at >= (now() - interval '10 days')::date")
-	})
+	sql := `
+		with case_statuses as (
+			select
+				i.id as inbox_id,
+				c.status,
+				snoozed_until is not null and snoozed_until > now() as snoozed
+            from cases c
+            inner join inboxes i on i.id = c.inbox_id
+            where
+            	org_id = $1
+             	and i.id = any($2)
+             	and c.created_at >= (now() - interval '10 days')::date
+        )
+        select
+        	i.name as inbox,
+	        count(*) filter (where snoozed) as snoozed,
+	        count(*) filter (where not snoozed and cs.status = 'pending') as pending,
+	        count(*) filter (where not snoozed and cs.status = 'investigating') as investigating,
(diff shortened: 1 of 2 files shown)


usecases/analytics_query_usecase.go
```diff
@@ -226,6 +231,123 @@ func (uc AnalyticsQueryUsecase) ScreeningHits(ctx context.Context,
...
+			From(dbmodels.TABLE_CASES).
+			Where("org_id = ?", filters.OrgId).
+			Where(squirrel.Eq{"inbox_id": inboxes}).
+			Where("created_at >= (now() - interval '10 days')::date")
+	})
+
+	query := squirrel.
+		Select(
+			"created_at::date as date",
+			"count(*) filter (where snoozed) as snoozed",
```
Pascal-Delange: Pure curiosity here, do not feel obligated to change it: would you think that grouping first, and pivoting next, gives faster queries ?

usecases/analytics_query_usecase.go
```diff
@@ -226,6 +231,123 @@ func (uc AnalyticsQueryUsecase) ScreeningHits(ctx context.Context,
...
+			"count(*) filter (where not snoozed  and status = 'pending') as pending",
+			"count(*) filter (where not snoozed and status = 'investigating') as investigating",
+			"count(*) filter (where not snoozed  and status = 'closed') as closed",
+		).
+		PrefixExpr(cte).
+		From("case_statuses").
+		GroupBy("created_at::date").
+		OrderBy("created_at::date")
```
Pascal-Delange: For this query (on the other one it's less important, though it might still be good for consistency?), I believe it would be best if we apply logic - either within the SQL, or perhaps more conveniently in the go code - to actually make sure we return only the _last 10 days_ of data (localized in the current timezone). As it is, if I'm not too tired and understand this `Where("created_at >= (now() - interval '10 days')::date")` correctly, I'm afraid the first day in chronological order would almost always be incomplete, and hence would show up as "smaller" on the graph (Of course the flipside of this is that the current day also shows up as partial, but that is fine).

usecases/analytics_query_usecase.go
apognu: This is probably what this does, since the aggregate functions in the `select` clause (the pivoting) must be executed on each group (so after the grouping).

PR #1316, opened 2025-11-14
cases count in inbox only counts open, non snoozed cases

Pascal-Delange:
So far the count of cases per inbox was only used in the inboxes settings page to show the number and decide if the inbox can be (soft) deleted. It counted all cases.
Now that we use it to display a case count in the main inbox page, it should show a count that is more aligned with what people care about, which is "how many active cases there are".


apognu:
The plan you pasted is just for one instance of the subquery, right? How does it fare when it has to be executed one time per inbox on a large number of inboxes?

apognu (approved):

PR #1328, opened 2025-11-19
Expose rule false positive ratio to rule hit table.

apognu:
The false positive ratio must be computed from the latest outcome of closed cases where a rule appears, this required adding the org ID to case events, ingesting some of the case events, and performing a separate query.

A migration adds the org ID, an appropriate index and backfills the data for the `case_events` table.

Not tested on production data yet.

---

> [!NOTE]
> Adds false-positive ratio to the rule hit table by exporting latest closed-case outcomes per rule to `rule_hit_outcomes`, requiring `org_id` on `case_events` and updating event creation paths.
> 
> - **Analytics queries**:
>   - Extend `analytics.RuleHitTable` with `rule_id` and `false_positive_ratio`; add `FalsePositiveRatio` type.
>   - `RuleHitTable` endpoint now joins aggregated false positives from `rule_hit_outcomes` (latest outcome per case).
> - **Data export pipeline**:
>   - New `AnalyticsCopyCaseEvents` to copy latest `outcome_updated` events for closed cases into `rule_hit_outcomes` (partitioned by org/year/month/trigger_object_type).
>   - Export job now runs `exportCaseEvents` in parallel with other exports and watermarks via `AnalyticsGetLatestRow`.
>   - Add `WithCtesRaw` to build CTEs with raw placeholders.
> - **Models/DB**:
>   - Add `OrgId uuid.UUID` to `models.CaseEvent`, `CreateCaseEventAttributes`, and DB adapter; batch insert now writes `org_id`.
>   - Migrations: add `org_id` column to `case_events`, backfill from `cases`, and create partial index `idx_case_events_by_org` on `(org_id, event_type, created_at)`.
> - **Usecases/Repositories**:
>   - Update all case-event creations to populate `OrgId` (case create/update, assign/unassign, snooze/unsnooze, comments/tags/files, decision review, SAR, screenings, rule snooze, escalate, mass update).
>   - Change `UpdateDecisionsWithEvents` to accept `orgId` and propagate across callers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30d0fe4af4ce3b5d7ecab35a9538f526a1903447. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


usecases/analytics_query_usecase.go
```diff
@@ -114,7 +116,60 @@ func (uc AnalyticsQueryUsecase) RuleHitTable(ctx context.Context, filters dto.An
...
 
-	return repositories.AnalyticsScanStruct[analytics.RuleHitTable](ctx, exec, query)
+	ruleHits, err := repositories.AnalyticsScanStruct[analytics.RuleHitTable](ctx, exec, query)
+	if err != nil {
+		return nil, err
+	}
+
+	// Fetch false positives from another query.
+	//
+	// In case of error (which could happen, for example, because of the
```
Pascal-Delange: I'm not sure I understand, in which case would this error typically occur ?

repositories/migrations/20251118115800_add_case_event_index.sql
```diff
@@ -0,0 +1,25 @@
...
+where c.id = case_events.case_id;
+
+alter table case_events
+    alter column org_id set not null;
+
+-- +goose StatementEnd
+
+create index concurrently if not exists idx_case_events_by_org
+    on case_events (org_id, event_type, created_at)
+    where event_type in ('outcome_updated');
```
Pascal-Delange: I created the index manually (⚠️  and created the org_id column and backfilled it for only outcome_updated events, so need to roll this back before deploying prod !!! ⚠️ ), so we can now test the query in prod.

repositories/migrations/20251118115800_add_case_event_index.sql
```diff
@@ -0,0 +1,25 @@
...
+
+alter table case_events
+    add column org_id uuid;
+
+update case_events
+set org_id = c.org_id
+from cases c
+where c.id = case_events.case_id;
+
+alter table case_events
```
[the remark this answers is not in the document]
apognu: I agree with the overall premise of your comments, except for the "solution" in your last paragraph.

Where it would work on our SaaS, we have no way to make sure that our self-hosted customer will not skip the first version and directly apply both migrations, which would incur the same problem (and past data shows they usually don't upgrade promptly).

We can reduce the possibility of this by waiting X releases between the two migrations, but the goalpost is just moved.

Maybe we should create the  non-null column with a default value of the nil UUID, backfill the values. And in a later migration, backfill the missing data again. It still doesn't solve all issues (the events created as the second migration runs will still get their default value instead of the org ID, but that's bad data instead of an error, and we can delay this?

And yes, I can move the index in a separate file so we can control the transaction better.

usecases/analytics_query_usecase.go
apognu: Any error that can happen on this query and not the other (network error, etc.). More to the point, and the one example I give below, the two sets of data are ingested separately (decision rule, then case events). In the case where the query happens when the case events are ingested and the case events are not (which will happen before the first export job runs), this query will error out as file not found.

PR #1360, opened 2025-11-28
Change shape of screening export query.

apognu:
Do not merge yet, ~this query is better but still has a sequential scan we want to try to avoid through column statistics.~

Let's be careful and check the query.


Pascal-Delange (approved):
Looks correct to me 👍

PR #1675, opened 2026-04-09
create migration to populate field/table metadata with DM options

OrriMandarin:
This pull request introduces a migration script to move configuration data from the `data_model_options` table into the `metadata` columns of `data_model_fields` and `data_model_tables`. This change helps consolidate field visibility and ordering preferences directly into the metadata, simplifying future queries and maintenance.

Migration of field visibility and order settings:

* For each table, fields not listed in `displayed_fields` (when set) are marked as hidden by updating their `metadata` with `{"hidden": true}`. The `object_id` field is always excluded from this update.
* The `field_order` array is translated from UUIDs to field names and stored in the table's `metadata` as a `fieldOrder` array, preserving the original order. Frontend use the field name in the ordering list

## Summary by CodeRabbit

* **Chores**
  * Performed database migration to consolidate field visibility preferences and field ordering settings into the system's metadata structure.

### DIFF
--- a/repositories/migrations/20260409200003_migrate_data_model_options_to_metadata.sql
+++ b/repositories/migrations/20260409200003_migrate_data_model_options_to_metadata.sql
@@ -1,0 +1,32 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- 1. Migrate displayed_fields → field metadata {"hidden": true}
+-- For each table with data_model_options, fields NOT in displayed_fields get hidden=true.
+-- Only apply when displayed_fields is not empty (empty means no preference was set).
+-- object_id is excluded because it is never part of displayed_fields/field_order options.
+UPDATE data_model_fields f
+SET metadata = COALESCE(f.metadata, '{}'::jsonb) || '{"hidden": true}'::jsonb
+FROM data_model_options o
+WHERE f.table_id = o.table_id
+  AND array_length(o.displayed_fields, 1) IS NOT NULL
+  AND f.id != ALL(o.displayed_fields)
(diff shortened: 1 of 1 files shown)


OrriMandarin:
@coderabbitai full review

repositories/migrations/20260409200003_migrate_data_model_options_to_metadata.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- 1. Migrate displayed_fields → field metadata {"hidden": true}
+-- For each table with data_model_options, fields NOT in displayed_fields get hidden=true.
+-- Only apply when displayed_fields is not empty (empty means no preference was set).
+-- object_id is excluded because it is never part of displayed_fields/field_order options.
```
Pascal-Delange, line 7: hmm why not, though ?

Pascal-Delange (approved):

repositories/migrations/20260409200003_migrate_data_model_options_to_metadata.sql
OrriMandarin: The existing Data model options didn’t include the `object_id` field. 
As a result, the `object_id` is never hidden by the frontend. 

The migration will simply update the hidden flag on the field that the frontend currently manages. 

I excluded the `object_id` field because the hidden setting is not configured for this field. 
However, with the new settings based on the `metadata` field, we can now hide the `object_id` field on the frontend.

PR #1700, opened 2026-04-22
fix: has past alert rule take into account the pivot on same table

OrriMandarin:
The new datamodel, by default, defines a pivot on the table itself (using the `object_id` field) if the table doesn’t have a link `belongs_to` defined. However, this rule didn’t take into account this type of pivot.

This pull request updates the logic in the `ObjectHasConfirmedRisks` method in `case_repository.go` to improve how related pivots are found. The main change ensures that pivots defined directly on a table's field (not just those linked via a path) are also considered when checking for confirmed risks.

**Query logic improvements:**

* The SQL now includes pivots that are defined directly on a field of the target table (using `field_id`), in addition to those found via path links, making the risk detection more comprehensive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated case evaluation logic to more accurately identify cases with confirmed risks by expanding the detection of applicable pivots for evaluated objects.

### DIFF
--- a/repositories/case_repository.go
+++ b/repositories/case_repository.go
@@ -731,26 +731,34 @@ func (repo *MarbleDbRepository) GetCasesRelatedToObject(ctx context.Context, exe
 
 // This query does the following:
 //  1. Find all links where this table is the arrival table
-//  2. Find all pivots whose path end with one of those links, effectively finding the pivots to that table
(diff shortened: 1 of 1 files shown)


Pascal-Delange (approved):
👍

PR #1708, opened 2026-04-26
fix: refuse percentiles above 1 (should be 0-1) and downscale legacy > 1

Pascal-Delange:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved percentile argument validation in aggregator evaluation. Dry-run mode now fails fast for invalid percentile values outside the 0-1 range. Runtime mode automatically normalizes legacy percentile inputs (values greater than 1) by dividing by 100.

### DIFF
--- a/usecases/ast_eval/evaluate/evaluate_aggregator.go
+++ b/usecases/ast_eval/evaluate/evaluate_aggregator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/checkmarble/marble-backend/models/ast"
 	"github.com/checkmarble/marble-backend/repositories"
 	"github.com/checkmarble/marble-backend/usecases/executor_factory"
+	"github.com/checkmarble/marble-backend/utils"
 )
 
 type AggregatorEvaluator struct {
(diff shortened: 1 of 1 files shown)


usecases/ast_eval/evaluate/evaluate_aggregator.go
```diff
@@ -126,6 +126,17 @@ func (a AggregatorEvaluator) Evaluate(ctx context.Context, arguments ast.Argumen
 				"missing or invalid value for percentile"))
 		}
 
+		if a.ReturnFakeValue {
+			// Dry-run / scenario validation: reject out-of-range so the builder surfaces a clear error.
+			if arg < 0 || arg > 1 {
+				return MakeEvaluateError(errors.Wrapf(ast.NewNamedArgumentError("percentile"),
+					"percentile value %v must be between 0 and 1", arg))
+			}
+		} else if arg > 1 {
```
apognu, line 140: Check negative values again here and normalize to 0? Should not happen (as > 1 should not happen), but let's make sure?

apognu (approved):
Commented. But pre-approved do it can be merged

usecases/ast_eval/evaluate/evaluate_aggregator.go
Pascal-Delange: yep, makes sense 👍

PR #1709, opened 2026-04-27
fix: ingestion with continuous screening, delta tack was ignored

OrriMandarin:
## Problem

When objects were added to monitoring via the ingestion endpoint, two bugs existed:

1. **Wrong delta track operation**: new objects always produced an \`Update\` delta track instead of an \`Add\` delta track in the screening worker (only when initial screening was requested).
2. **Missing delta track for non-screened objects**: new objects with \`ShouldScreen=false\` were inserted into the monitoring table but never got a delta track, making them invisible to future dataset rebuilds.

The root cause: the ingestion path mixed monitoring registration into the ingestion transaction, then handed off everything to \`DoScreeningWorker\`, which was designed for ongoing updates — not initial registration.

A concurrent race condition also existed: when an object with multiple continuous screening configurations was ingested, two \`RegisterObjectWorker\` jobs ran simultaneously for the same \`object_id\`. Both saw an empty monitored objects table and both created an \`Add\` delta track, resulting in duplicate entries.

## Solution

Extract object registration from the ingestion transaction into a dedicated \`RegisterObjectWorker\`. The worker owns the full registration lifecycle for objects added via ingestion:

1. **Registration** (client DB transaction): insert into \`_monitored_object\` and create an audit entry with caller identity (\`userId\`/\`apiKeyId\`). Idempotent — a duplicate job detects the existing entry and skips reinsertion.
2. **Delta track** (marble DB): record an \`Add\` operation independently of screening. Always attempted — \`ON CONFLICT DO NOTHING\` makes it idempotent, so retries and concurrent jobs for the same object are handled by the unique constraint on \`(org_id, object_type, object_id, object_internal_id)\` for \`add\`/\`update\` operations.
3. **Screening** (marble DB transaction, optional): query OpenSanctions, persist the result, enqueue match enrichment, and create a case if in-review. Always attempted even on retry, so a failed screening transaction does not leave the object permanently unscreened.

The unique constraint on the delta track table enforces the business invariant at the database level: two concurrent jobs racing to create the first \`Add\` delta track for the same object version will have one silently no-op via \`ON CONFLICT DO NOTHING\`. A migration deduplicates any existing duplicate rows before creating the index.

\`DoScreeningWorker\` becomes update-only, which makes its hardcoded \`DeltaTrackOperationUpdate\` always correct.

## Changes

- \`ingestion_usecase.go\`: remove \`InsertContinuousScreeningObject\` from the ingestion transaction; detect new \`(objectId, configId)\` pairs after ingestion and enqueue \`RegisterObjectWorker\` jobs instead.
- \`worker_register_object.go\`: new worker implementing the two-step registration flow above. Simplified from the original design — no verify enqueue needed since the delta track insert is idempotent.
- \`worker_verify_delta_track_existence.go\`: safety-net worker for the manual add endpoint path, repairs missing delta tracks from cross-DB atomicity failures (~5 min delayed).
- \`continuous_screening_repository.go\`: \`CreateContinuousScreeningDeltaTrack\` now uses \`ON CONFLICT DO NOTHING\`.
- \`migrations/20260427160000\`: deduplicates existing duplicate delta tracks, then adds the unique index on \`(org_id, object_type, object_id, object_internal_id)\` for \`add\`/\`update\` operations.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous background job processing for registering objects to continuous screening and verifying delta track status.
  * Enabled batch registration of multiple monitored objects for continuous screening.

* **Bug Fixes**
  * Made continuous screening registration idempotent; registering an already-monitored object no longer returns a conflict error.

* **Documentation**
  * Updated API documentation to reflect idempotent registration behavior and improved response descriptions.

### DIFF
--- a/repositories/continuous_screening_client_repository.go
+++ b/repositories/continuous_screening_client_repository.go
@@ -121,25 +121,21 @@ func (repo *ClientDbRepository) InsertContinuousScreeningObject(
 	objectType string,
 	objectId string,
 	configStableId uuid.UUID,
-	ignoreConflicts bool,
-) error {
+) (uuid.UUID, error) {
 	if err := validateClientDbExecutor(exec); err != nil {
-		return err
-	}
-
-	sql := "INSERT INTO %s (id, object_type, object_id, config_stable_id) VALUES ($1, $2, $3, $4)"
-
-	if ignoreConflicts {
-		sql += " ON CONFLICT DO NOTHING"
+		return uuid.Nil, err
 	}
 
 	query := fmt.Sprintf(
-		sql,
+		"INSERT INTO %s (id, object_type, object_id, config_stable_id) VALUES ($1, $2, $3, $4)",
 		sanitizedTableName(exec, dbmodels.TABLE_CONTINUOUS_SCREENING_MONITORED_OBJECTS),
 	)
 
-	_, err := exec.Exec(ctx, query, pure_utils.NewId(), objectType, objectId, configStableId)
-	return err
+	id := pure_utils.NewId()
+	if _, err := exec.Exec(ctx, query, id, objectType, objectId, configStableId); err != nil {
+		return uuid.Nil, err
+	}
+	return id, nil
 }
 
 // Table schema
--- a/repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
+++ b/repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
@@ -1,0 +1,21 @@
+-- +goose Up
+-- +goose StatementBegin
+DELETE FROM continuous_screening_delta_tracks a
+USING continuous_screening_delta_tracks b
+WHERE a.operation IN ('add', 'update')
+  AND b.operation IN ('add', 'update')
+  AND a.org_id = b.org_id
+  AND a.object_type = b.object_type
+  AND a.object_id = b.object_id
+  AND a.object_internal_id = b.object_internal_id
+  AND (a.created_at < b.created_at OR (a.created_at = b.created_at AND a.id < b.id));
+
+CREATE UNIQUE INDEX idx_cs_delta_tracks_unique_add_update
+ON continuous_screening_delta_tracks (org_id, object_type, object_id, object_internal_id)
+WHERE operation IN ('add', 'update');
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP INDEX idx_cs_delta_tracks_unique_add_update;
+-- +goose StatementEnd
--- a/repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
+++ b/repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
@@ -1,5 +1,7 @@
 -- +goose Up
 -- +goose StatementBegin
+LOCK TABLE continuous_screening_delta_tracks IN SHARE MODE;
+
 DELETE FROM continuous_screening_delta_tracks a
 USING continuous_screening_delta_tracks b
 WHERE a.operation IN ('add', 'update')
--- a/usecases/continuous_screening/worker_register_object.go
+++ b/usecases/continuous_screening/worker_register_object.go
@@ -39,6 +39,10 @@ type registerObjectWorkerTaskQueueRepository interface {
 		organizationId uuid.UUID,
 		continuousScreeningId uuid.UUID,
 	) error
+	EnqueueContinuousScreeningVerifyDeltaTrackExistenceTask(
+		ctx context.Context,
+		args models.ContinuousScreeningVerifyDeltaTrackExistenceArgs,
+	) error
 }
 
 type registerObjectWorkerClientDbRepository interface {
@@ -48,8 +52,7 @@ type registerObjectWorkerClientDbRepository interface {
 		objectType string,
 		objectId string,
 		configStableId uuid.UUID,
-		ignoreConflicts bool,
-	) error
+	) (uuid.UUID, error)
 	InsertContinuousScreeningAudit(
 		ctx context.Context,
 		exec repositories.Executor,
@@ -141,11 +144,15 @@ func (w *RegisterObjectWorker) Timeout(_ *river.Job[models.ContinuousScreeningRe
 // Work registers a newly ingested object under continuous screening monitoring.
 // The flow has three independent steps, each of which is safe to retry:
 //  1. Registration (client DB transaction): if the (objectId, configId) pair is not already
-//     in the monitoring table, insert it and create an audit entry. Idempotent — a second run
-//     detects the existing entry and skips to step 3.
+//     in the monitoring table, insert it and create an audit entry. The same client-DB
+//     transaction also enqueues a delayed verification job (against the marble pool) — see
+//     VerifyDeltaTrackExistenceWorker — so a failure of step 2 below can be self-healed.
+//     The verify enqueue runs before the client-DB commit: if the enqueue fails, the
+//     monitored_object insertion is rolled back and the worker errors for River to retry.
 //  2. Delta track (marble DB): record an Add operation so the object is included in the next
 //     dataset rebuild. Skipped on retry (alreadyRegistered) and skipped when the object is
-//     already tracked by another config (no new entry needed in the index).
+//     already tracked by another config (no new entry needed in the index). If this step
+//     fails, the queued verify job (5 min later) creates the missing track.
 //  3. Screening (marble DB transaction, only if ShouldScreen): query OpenSanctions, persist
 //     the result, enqueue match enrichment, and create a case if the result is in-review.
 //     This step is always attempted — even on retry — so a failed screening transaction does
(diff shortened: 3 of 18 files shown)


usecases/continuous_screening/worker_register_object.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	var apiKeyId *uuid.UUID
+	if job.Args.ApiKeyId != nil {
+		parsed, err := uuid.Parse(*job.Args.ApiKeyId)
+		if err != nil {
+			logger.WarnContext(ctx, "could not parse api_key_id", "api_key", *job.Args.ApiKeyId)
+			return nil
+		}
+		apiKeyId = &parsed
+	}
```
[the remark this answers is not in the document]
OrriMandarin: The issue arises when returning an error, causing the job to retry multiple times, which is unnecessary.

usecases/continuous_screening/worker_register_object.go
```diff
@@ -0,0 +1,317 @@
...
+	newObjectData, err := w.ingestedDataReader.QueryIngestedObjectByInternalId(
+		ctx, clientDbExec, table, newObjectInternalId, "id", "valid_from",
+	)
+	if err != nil {
+		if errors.Is(err, models.NotFoundError) {
+			logger.WarnContext(ctx, "Continuous Screening - ingested object not found, skipping registration",
+				"new_internal_id", newObjectInternalId)
+		}
+		return err
+	}
```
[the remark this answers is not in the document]
OrriMandarin: I rewrite the comment bu deleting the skipping info. 

Want to retry in case the object hasn't been ingested but ingest after

usecases/continuous_screening/worker_register_object.go
```diff
@@ -0,0 +1,317 @@
...
+			OrgId:            config.OrgId,
+			ObjectType:       job.Args.ObjectType,
+			ObjectId:         job.Args.ObjectId,
+			ObjectInternalId: &newObjectInternalId,
+			EntityId:         pure_utils.MarbleEntityIdBuilder(job.Args.ObjectType, job.Args.ObjectId),
+			Operation:        models.DeltaTrackOperationAdd,
+		}); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
OrriMandarin: I will add an other job to check if the delta is correctly created after the monitored object insertion

usecases/continuous_screening/worker_register_object.go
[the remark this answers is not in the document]
OrriMandarin: I think we need to find a way based on the log to have an alert. In this case, we need to fix the params and retry the job with a correct params.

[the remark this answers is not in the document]
OrriMandarin: Thanks I will use JobCancel :D

OrriMandarin:
@coderabbit full review

repositories/continuous_screening_client_repository.go
OrriMandarin, line 1: Can't ignore the conflict. There is no path where it is acceptable.

OrriMandarin:
I tested the following scenarios:



- [x] **Adding an existing object under monitoring:**
   - The object was successfully added to the `_monitored_object` and `_monitored_object_audit` tables.
   - A delta track was created for the object.
   - The delta track job was scheduled and executed without any errors.

- [x] **Adding an existing object under monitoring with failed screening:**
   - The object was added to the `_monitored_object` and `_monitored_object_audit` tables.
   - A delta track was not created because the insertion occurred after the screening process.
   - The delta track job was scheduled and executed without errors with inserting the missing data
   - An error message was returned, but the object was still added under monitoring. This case is already present and can be reworked.

- [x] **Adding a non-existing object under monitoring:**
   - An error was encountered as expected.

- [x] **Adding an object that is already under monitoring:**
   - A conflict error was generated, and no changes were made.

- [x] **Ingesting a new object with `monitor=true&configuration_ids=….`:**
   - The object was successfully added to the `_monitored_object` and `_monitored_object_audit` tables.
   - A delta track was created for the object.
   - The delta track job was scheduled and executed without any errors.
   - A case was created.

- [x] **Ingesting a new object with multiple configurations:**
   - The object was successfully added to the `_monitored_object` and `_monitored_object_audit` tables.
   - Only one delta track was created for the object.
   - The delta track job was scheduled and executed without any errors.
   - A case was created.

- [x] **Ingesting a new object without screening:**
   - The object was successfully added to the `_monitored_object` and `_monitored_object_audit` tables.
   - A delta track was created for the object.
   - The delta track job was scheduled and executed without any errors.
   - No case was created.

- [x] **Ingesting a new object without monitoring:**
   - The object was successfully ingested.
   - No continuous or case were created.

repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
```diff
@@ -0,0 +1,21 @@
...
+  AND b.operation IN ('add', 'update')
+  AND a.org_id = b.org_id
+  AND a.object_type = b.object_type
+  AND a.object_id = b.object_id
+  AND a.object_internal_id = b.object_internal_id
+  AND (a.created_at < b.created_at OR (a.created_at = b.created_at AND a.id < b.id));
+
+CREATE UNIQUE INDEX idx_cs_delta_tracks_unique_add_update
+ON continuous_screening_delta_tracks (org_id, object_type, object_id, object_internal_id)
+WHERE operation IN ('add', 'update');
```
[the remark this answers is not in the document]
OrriMandarin: It’s okay since no client is using it for now.

repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
```diff
@@ -0,0 +1,21 @@
...
+USING continuous_screening_delta_tracks b
+WHERE a.operation IN ('add', 'update')
+  AND b.operation IN ('add', 'update')
+  AND a.org_id = b.org_id
+  AND a.object_type = b.object_type
+  AND a.object_id = b.object_id
+  AND a.object_internal_id = b.object_internal_id
+  AND (a.created_at < b.created_at OR (a.created_at = b.created_at AND a.id < b.id));
+
+CREATE UNIQUE INDEX idx_cs_delta_tracks_unique_add_update
```
Pascal-Delange, line 15: Heads up that this will lock the table for the duration. It's ok because it's all async.

repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose Up
+-- +goose StatementBegin
+DELETE FROM continuous_screening_delta_tracks a
```
Pascal-Delange, line 5: Just note that, if the code that avoids creating duplicates in the first place is deployed by this PR, depending on timing, this migration may be tricky to run (step 1 completes, but new duplicates created by the time you run the migration, rinse and repeat)

usecases/ingestion_usecase.go
Pascal-Delange, line 737: Hum, should this be done outside of the big transaction ?

usecases/continuous_screening/worker_register_object.go
```diff
@@ -0,0 +1,299 @@
...
+
+	if err := w.usecase.CheckFeatureAccess(ctx, job.Args.OrgId); err != nil {
+		logger.WarnContext(ctx, "Continuous Screening - feature access not allowed, skipping registration", "error", err)
+		return nil
+	}
+
+	newObjectInternalId, err := uuid.Parse(job.Args.NewInternalId)
+	if err != nil {
+		return river.JobCancel(errors.New("could not parse new internal id"))
```
Pascal-Delange, line 181: @OrriMandarin On top of coderabbits comment, an interesting thing to check (I don't know what it would do): what happens on sentry if we river.jobCancel.

OrriMandarin: I suppose the error handler is essentially the same as returning an error directly without retrying the job and immediately set the job status as canceled. 
Our `LoggerMiddleware` will handle the error and report it to Sentry.

usecases/ingestion_usecase.go
OrriMandarin: Which transaction?

repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
OrriMandarin: As discussed, the table is currently only used by a worker. I will lock the table for the migration since locking it won’t affect time-sensitive operations.

usecases/ingestion_usecase.go
Pascal-Delange: hmm nevermind

OrriMandarin:
@coderabbit full review

repositories/migrations/20260427160000_add_unique_constraint_cs_delta_tracks.sql
OrriMandarin: Done

OrriMandarin:
@Pascal-Delange,

I’ve split the change into several commits, each with a single main change. You can review them.

- Renamed the worker.
- Checked if the configuration is compatible with the object type when ingesting.
- Deleted the delta track creation in the continuous screening endpoint and relied on the worker.
- Continuous screening is now idempotent. Calling it multiple times will redo the screening and case creation, but it won’t raise a conflict error.

Pascal-Delange (approved):
Great, thanks !

PR #1733, opened 2026-05-05
feat: add free form table to store search history

OrriMandarin:
Create a free form table to store each free form search made by a user, whether from the frontend or API. This table will be used to display the search history and also for the metric collector to count the number of searches made by providers.
Persist every freeform search usage

Currently, only Opensanctions is available for free form searches. By default, we store the results with this provider. 
In the future, we can change the provider and store the appropriate one in the table. 

## New Features
* Added freeform search functionality for screening operations
* Added OpenSanctions provider support
* Search requests are now recorded

### DIFF
--- a/models/screening_providers.go
+++ b/models/screening_providers.go
@@ -1,0 +1,4 @@
+package models
+
+// TODO: TBD confirm provider name
+const ScreeningProviderOpenSanctions = "opensanctions"
--- a/repositories/dbmodels/db_freeform_search.go
+++ b/repositories/dbmodels/db_freeform_search.go
@@ -1,0 +1,38 @@
+package dbmodels
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/utils"
+	"github.com/google/uuid"
+)
+
+const TABLE_FREEFORM_SEARCHES = "screening_freeform_searches"
+
+var SelectFreeformSearchColumn = utils.ColumnList[DBFreeformSearch]()
+
+type DBFreeformSearch struct {
+	Id          uuid.UUID       `db:"id"`
+	OrgId       uuid.UUID       `db:"org_id"`
+	UserId      *uuid.UUID      `db:"user_id"`
+	ApiKeyId    *uuid.UUID      `db:"api_key_id"`
+	Provider    string          `db:"provider"`
+	CreatedAt   time.Time       `db:"created_at"`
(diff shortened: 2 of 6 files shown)


models/screening_providers.go
```diff
@@ -0,0 +1,4 @@
+package models
+
+// TODO: TBD confirm provider name
+const ScreeningProviderOpenSanctions = "opensanctions"
```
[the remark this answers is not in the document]
apognu: NOP.

repositories/dbmodels/db_freeform_search.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/utils"
+	"github.com/google/uuid"
+)
+
+const TABLE_FREEFORM_SEARCH = "screening_freeform_search"
```
apognu: Nitpick, but let's keep plural names for tables (`screening_freeform_searches`).

repositories/freeform_search_repository.go
```diff
@@ -0,0 +1,39 @@
...
+
+	sql := NewQueryBuilder().
+		Insert(dbmodels.TABLE_FREEFORM_SEARCH).
+		Columns(
+			"id",
+			"org_id",
+			"user_id",
+			"api_key_id",
+			"provider",
+			"search_input",
```
apognu, line 27: Are we not persisting the results? There is a column in the table but do not pass any value into it.

repositories/dbmodels/db_freeform_search.go
OrriMandarin: Ah right, thx

repositories/freeform_search_repository.go
OrriMandarin: Oh, I forgot to mention the result behavior in my comment.
The user will have the option to save or not save the result (an opt-in feature).

[Based on the design I saw](https://www.figma.com/design/a9n93l5twcEpmthKw6cBz5/Marble—Wireframes?node-id=6129-83693&t=Wqgk8LEC0fbLQkJw-0), we will have a button to save the result, which means an endpoint will handle the persistence of the result. That why this field is nullable.

This PR doesn’t include this feature at the moment, but I can add it if it makes sense to include it now.

repositories/dbmodels/db_freeform_search.go
OrriMandarin: Done

apognu (approved):

PR #1734, opened 2026-05-06
feat: add monitored object active gauge

OrriMandarin:
## Summary

This PR adds a new metrics collector that counts the number of **distinct monitored objects active during the current billing year** per organization.

### What is an \"active\" monitored object?

A monitored object (identified by its `object_type` + `object_id` pair, regardless of which `config_stable_id` it is linked to) is considered **active for a given year** if it was being monitored at any point during that year, even if it was later removed.

This metric is used for billing purposes: if an organization monitored an object at any point during the year, it counts toward their yearly quota. Removing an object mid-year does not cancel it from the year's count, but it will not count for the following year if not re-added.

### Query design

The metric is computed with a two-part `UNION` query against the client DB:

```sql
SELECT COUNT(*) FROM (
    SELECT DISTINCT object_type, object_id FROM _monitored_objects
    UNION
    SELECT DISTINCT object_type, object_id FROM _monitored_objects_audit
    WHERE action = 'remove'
        AND created_at >= <year_start>
        AND created_at < <year_end>
)
```

- **First branch** — the live `_monitored_objects` table: captures all objects currently being monitored, regardless of when they were added or how many configs they are linked to.
- **Second branch** — the `_monitored_objects_audit` table filtered to `remove` actions during the year: captures objects that were active at some point during the year but have since been removed.
- **`UNION`** deduplicates by `(object_type, object_id)`, so the same object appearing under multiple `config_stable_id` values or in both branches is counted only once.

### Scenario coverage

| Scenario | In live table | In audit removes this year | Counted? | Expected |
|---|---|---|---|---|
| Added before year Y, still active | ✓ | — | ✓ | ✓ |
| Added during year Y, still active | ✓ | — | ✓ | ✓ |
| Added before year Y, removed during year Y | — | ✓ | ✓ | ✓ |
| Added during year Y, removed during year Y | — | ✓ | ✓ | ✓ |
| Added before year Y, removed before year Y, never re-added | — | — | ✗ | ✓ correct exclusion |
| Added before year Y, removed before year Y, re-added before year Y, still active | ✓ | — | ✓ | ✓ |
| Added before year Y, removed before year Y, re-added before year Y, removed during year Y | — | ✓ | ✓ | ✓ |
| Added before year Y, removed before year Y, re-added during year Y, still active | ✓ | — | ✓ | ✓ |
| Removed during year Y, then re-added during year Y, still active | ✓ | ✓ | ✓ (UNION deduplicates) | ✓ |
| 2 configs, both active | ✓ ✓ | — | ✓ (DISTINCT collapses) | ✓ |
| 2 configs, one active, one removed during year Y | ✓ | ✓ | ✓ (UNION deduplicates) | ✓ |
| 2 configs, both removed during year Y | — | ✓ ✓ | ✓ (DISTINCT collapses) | ✓ |
| 2 configs, both removed before year Y | — | — | ✗ | ✓ correct exclusion |

### Changes

- New metric name: `monitored_objects_active.gauge`
- New repository method `CountActiveMonitoredObjectsForYear` on `ClientDbRepository`
- New collector `MonitoredObjectActiveCollector` — skips orgs without continuous screening set up, emits one gauge per org with `From = Jan 1` of the current year and `To = collection time`
- Registered in `NewCollectorsV1`
- Unit tests covering: active objects, no objects, org not set up (skipped)

## Summary by CodeRabbit

* **New Features**
  * Metrics collection now reports active monitored objects per organization on a year-bounded basis.

* **Refactoring**
  * Replaced the previous continuous-screening collector with a year-based monitored-object active collector and updated the collectors pipeline accordingly.

* **Tests**
  * Updated tests to validate the new year-bounded collection behavior and expected per-organization metrics.

* **Removed**
  * Deprecated the prior continuous-screening counting approach and related collection behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/checkmarble/marble-backend/pull/1734?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


OrriMandarin:
@coderabbit full review

repositories/continuous_screening_client_repository.go
```diff
@@ -397,6 +398,45 @@ func (repo *ClientDbRepository) ListMonitoredObjects(
...
+		) AS t
+	`, monitoredObjectsTable, auditTable)
+
+	var count int
+	if err := exec.QueryRow(ctx, sql, yearStart, yearEnd).Scan(&count); err != nil {
+		return 0, err
+	}
+
+	return count, nil
+}
```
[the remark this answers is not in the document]
OrriMandarin: Wrong, This function is designed to get the gauge metric of active monitored objects for the **current year** and can not be used to get historical data for previous years because it is based on the `_monitored_objects` table. objects removed YearX will not be present in the table to compute gauge for YearX-1, try to recompute the metric of YearX-1 using this function will return incorrect value.
For historical data, you should based the count on the audit table only and on the `action` field

usecases/metrics_collection/constant.go
```diff
@@ -1,11 +1,12 @@
...
-	CheckLicenseMetricName       = "check_license.count"
-	CSMonitoredObjectsMetricName = "monitored_objects.gauge"
+	AiCaseReviewCountMetricName        = "ai_case_reviews.count"
+	AppVersionMetricName               = "app_version"
+	CaseCountMetricName                = "cases.count"
+	DecisionCountMetricName            = "decisions.count"
+	ScreeningCountMetricName           = "screenings.count"
+	CheckLicenseMetricName             = "check_license.count"
+	CSMonitoredObjectsMetricName       = "monitored_objects.gauge"
+	CSMonitoredObjectsActiveMetricName = "monitored_objects_active.gauge"
```
OrriMandarin: In fact, instead of introducing a new metric and deprecating `monitored_objects.gauge`, I can reuse it.

The existing metric isn’t currently used for billing and will always be equal to or less than the `monitored_objects_active.gauge` because it doesn’t account for the removed object.

OrriMandarin: I will remove the new one, I will change the worker implementation instead of creating a new one

OrriMandarin:
⚠️ The query doesn’t consider the possibility of having a disabled continuous screening configuration. In such a scenario, an object linked to a disabled configuration is counted as an active object.

Currently, the product doesn’t provide users with the ability to disable a configuration. I’m not sure if we need to address this possibility.

apognu (approved):

Pascal-Delange (approved):
LGTM

PR #1767, opened 2026-06-01
add stable rule id (and screening rule name) to decision rule, screen…

Pascal-Delange:
…ing DTO in public API

## Summary by CodeRabbit

* **New Features**
  * API responses for decision rules and screenings now include a stable rule identifier (UUID) that persists across scenario versions.
  * Screening results now include the rule name.


pubapi/openapi/v1.yml
```diff
@@ -990,12 +990,17 @@ components:
...
       required:
         - name
+        - stable_rule_id
         - outcome
         - score_modifier
       properties:
         name:
           description: Name of the rule in the scenario version
           type: string
+        stable_rule_id:
```
apognu: Not a big fan of having internal implementation details (here, the fact that that the ID is stable) in the Publix API. I believe it's too much information and misleading.

"The other IDs are unstable? 😰" — Someone, somewhere.

The same way I renamed `stable_config_id` to `monitoring_config_id`. The only rule ID a user needs to interact with is this one, so I prefer `rule_id`.

Pascal-Delange: Agreed 👍 
Let me change that

apognu (approved):

PR #1778, opened 2026-06-05
add filters for freeform searches

Pascal-Delange:
- Implement filters & pagination when listing freeform searches
- Add an endpoint to save a freeform search's results. Idempotent. Returns a 409 if the set of entity_ids that motiva returns has changed since the initial freeform search was done. In this case, there is nothing to do, the search is just presumably too old for it to make sense to save the results.
- And another endpoint that reads a freeform search with its detailed results (unenriched, however)
- delete all past screening_freeform_searches entries as the format has changed (and there is not a lot of useful past data anyway)


utils/validator.go
```diff
@@ -12,6 +9,9 @@ import (
 	"github.com/go-playground/validator/v10"
 )
 
+// Authored by Antoine Popineau
+// MIT license
```
apognu, line 13: ❤️

apognu (approved):

utils/validator.go
Pascal-Delange: Credit where it's due :D

PR #1794, opened 2026-06-12
tentative: changes to CI/CD to push a release candidate tag on prerel…

Pascal-Delange:
Switch the production deploy trigger from tag-push to GitHub Release events, so pre-releases publish under their suffixed tag instead of a clean stable one.

- **Trigger:** `push:` on `v[0-9]+.[0-9]+.[0-9]+` tags → `release: [published]`
- **Harden against shell injection:** the release tag name (`inputs.version`) is (internal) attacker-influenceable, so it's now referenced as `$VERSION_IMAGE` (job-level `env`) in all `run:` blocks instead of being interpolated via `${{ }}` — the latter pastes the value into the script as code; the former passes it as inert data. (per CodeRabbit)

The build → `:<sha>` → retag → deploy pipeline and `git describe` version baking are unchanged. Every release (rc and stable) builds fresh.

## Process note

CI no longer gates on tag shape, so the suffix discipline is on us: cut pre-releases with a `-rc.1` tag (not a clean `v1.2.0`). Standard tooling (Dependabot) then excludes them by default. Promote by publishing a **separate** clean `v1.2.0` release on the same commit — don't un-check the pre-release box (that fires `released`/`edited`, not `published`).

## Summary by CodeRabbit

* **Chores**
  * Updated production deployment workflow to support both stable and release-candidate version releases.
  * Enhanced backend deployment process with improved image tagging and deployment handling across services.


.github/workflows/backend_deploy_workflow.yaml
```diff
@@ -54,23 +53,38 @@ jobs:
...
           password: ${{ steps.auth.outputs.access_token }}
 
-      - name: Extract version from tag
+      # BUILD_IMAGE is the artifact we promote from. Releases key it by
+      # (version + sha) so a pre-release and its promoted stable release share
+      # one artifact, while never colliding with the staging :<sha> build.
+      - name: Compute versions
         id: version
-        run: echo "MARBLE_VERSION=$(git describe --tags)" >> "$GITHUB_OUTPUT"
+        run: |
```
apognu: I don't understand this whole thing.

Why do we need to change the "commit SHA" image into this "truncated tag + commit SHA".

Also, as far as I can see it, we will lose any indication in the SaaS production UI, on exactly which version is deployed, all RCs and base version will be scrambled together.

Same with the tagged image, we remove the rc suffix and add the commit SHA instead, which is arguably worse to determine at first glance which is which.

.github/workflows/deploy_production.yaml
```diff
@@ -1,12 +1,11 @@
 name: Deploy Production
 
 on:
```
apognu, line 3: Any reason we don't use either a regex match of `/v\d+\.\d+\.\d+(?:-rc\.\d+)?/` or two patterns with and without the `rc` suffix?

Pascal-Delange: No strong reason, I can push that

apognu (approved):

PR #1797, opened 2026-06-15
Save delta record last_change to Redis to prevent reprocessing the sa…

apognu:
…me entity over and over again.

## Summary by CodeRabbit

* **New Features**
  * Added last change timestamp tracking for entity updates in delta file processing

* **Performance Improvements**
  * Optimized delta file processing efficiency


Pascal-Delange (approved):
Let's just hope (95% of confidence here) that the entity "updated at" received from LN is really updated _every time_ an entity is touched in one way or another

PR #1809, opened 2026-06-17
Support screening entities with timestamp fields.

apognu:
## Release Notes

**New Features**
- Timestamp handling now accepts date-only input (`YYYY-MM-DD`) for screening query filters and CSV/payload timestamp parsing, in addition to existing full datetime formats.
- Screening query formulas can now evaluate to either `string` or `time.Time`; nil/zero/empty values are treated as empty fields.
- When a time value falls exactly at midnight, it’s represented as a date-only value (`YYYY-MM-DD`) for consistent formatting.

**Tests**
- Added coverage for additional date-only timestamp cases (including a new `date_of_birth` scenario).


usecases/payload_parser/payload.go
```diff
@@ -241,6 +241,9 @@ func NewParser(opts ...ParserOpt) *Parser {
 			if t, err := time.Parse("2006-01-02T15:04:05", result.String()); err == nil {
 				return t.UTC(), nil
 			}
+			if t, err := time.Parse("2006-01-02", result.String()); err == nil {
+				return t, nil
+			}
```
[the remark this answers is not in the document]
apognu: I believe this is on purpose, we are going from the assumption that when something only makes sense as a raw date (without a time portion), they need to be expressed without it, therefore being understood as UTC.

usecases/scenarios/scenario_validation.go
```diff
@@ -165,11 +165,20 @@ func (self *ValidateScenarioIterationImpl) Validate(ctx context.Context,
...
+				returnTypeIsValid := false
+
+				if _, ok := queryNameValidation.RuleEvaluation.ReturnValue.(string); ok {
+					returnTypeIsValid = true
+				}
+				if _, ok := queryNameValidation.RuleEvaluation.ReturnValue.(time.Time); ok {
+					returnTypeIsValid = true
+				}
+
+				if !returnTypeIsValid {
```
OrriMandarin, line 177: What about something like 

```
switch queryNameValidation.RuleEvaluation.ReturnValue.(type) {
case string, time.Time:
default:
    queryNameValidation.Errors = append(...)
}
```

OrriMandarin (approved):

usecases/payload_parser/payload.go
```diff
@@ -241,6 +241,9 @@ func NewParser(opts ...ParserOpt) *Parser {
 			if t, err := time.Parse("2006-01-02T15:04:05", result.String()); err == nil {
 				return t.UTC(), nil
 			}
+			if t, err := time.Parse("2006-01-02", result.String()); err == nil {
```
Pascal-Delange, line 249: Watch out, there is another implementation of parsing for reading from a csv file, in ingestion_usecase.go/parseStringValuesToMap 🙏

Pascal-Delange (changes requested):
Looks ok except see comment about csv ingestion

apognu:
I also added special paths for `updated_at`, since we still want those to not be provided a civil dates.

Pascal-Delange (approved):
Thanks!

PR #1815, opened 2026-06-18
feat: restrict screening provider changes to Marble admins on managed deployments

OrriMandarin:
## What & why

Organizations choose a screening provider (OpenSanctions / LexisNexis) per feature via the `screening_providers` map, edited through `PATCH /organizations/:organization_id`. Today any org `ADMIN` (who holds `ORGANIZATIONS_UPDATE`) can change it.

This restricts that on **managed (SaaS)** deployments: only the **`MARBLE_ADMIN`** role may change the screening provider. On **self-hosted** deployments the org admin keeps the ability (still subject to the existing LexisNexis license-entitlement check).

## Approach

The rule is **mode-conditional** (MARBLE_ADMIN-only *on managed*; org admin allowed *on self-hosted*). A static role→permission table can't express that — a permission granted only to `MARBLE_ADMIN` would imply org admins never can, which is false for self-hosted. So it's a direct role check in a new `EnforceSecurityOrganization.EditScreeningProvider` method, matching the established Marble-admin pattern in `enforce_org_import_security.go`.

- Deployment mode (`IsManagedMarble`) is passed into the usecase **as a bool fact**, keeping `usecases/security/` free of any license/config dependency. This is the first authz check in that layer to branch on deployment mode.
- The check fires only when a provider value **actually changes**, so idempotent PATCHes that echo the current map, and edits to other org fields (timezone, thresholds, auto-assign), are unaffected.
- The self-hosted branch delegates to the existing `EditOrganization(org)`, making the method self-contained (both checks are pure in-memory).

## Behavior

| Deployment | Caller | Change provider? |
|---|---|---|
| Managed | org `ADMIN` | ❌ 403 |
| Managed | `MARBLE_ADMIN` | ✅ |
| Self-hosted | org `ADMIN` | ✅ |
| Any | no actual provider change | unaffected |

## Test plan

- New test covers managed/self-hosted × MARBLE_ADMIN / ADMIN / VIEWER

### DIFF
--- a/mocks/enforce_security.go
+++ b/mocks/enforce_security.go
@@ -175,6 +175,11 @@ func (e *EnforceSecurity) EditOrganization(org models.Organization) error {
 	return args.Error(0)
 }
 
+func (e *EnforceSecurity) EditOrganizationScreeningProvider(org models.Organization, isManagedMarble bool) error {
+	args := e.Called(org, isManagedMarble)
+	return args.Error(0)
+}
+
 func (e *EnforceSecurity) DeleteOrganization() error {
 	args := e.Called()
 	return args.Error(0)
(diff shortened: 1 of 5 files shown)


OrriMandarin:
@coderabbit full review

apognu (approved):

Pascal-Delange (approved):

PR #1818, opened 2026-06-22
chore: use copy() for slice assignment in decision screening

OrriMandarin:
Address linting warning by using the built-in `copy()` function instead of a manual loop for slice assignment.

## Summary by CodeRabbit

* **Refactor**
  * Improved how decision screening results are copied into the final response, maintaining the same output while streamlining internal handling.

### DIFF
--- a/usecases/decision_usecase.go
+++ b/usecases/decision_usecase.go
@@ -132,10 +132,7 @@ func (usecase *DecisionUsecase) GetDecision(ctx context.Context, decisionId stri
 	}
 
 	decision.ScreeningExecutions = make([]models.ScreeningWithMatches, len(scs))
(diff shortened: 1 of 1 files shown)


apognu (approved):
I think we know of those, but Pascal left them on purpose because they are clearer (at least I remember him adding a comment in that regard).

No beef merging this, but we need to all agree (there are other places with this same lint violation).

cc @Pascal-Delange

Pascal-Delange:
Thanks,
Yea as Antoine said. I don't have a religion on this and I'm flexible, but generally speaking I find that using copy for those cases is useless obfuscation and hurts readability more than it helps with anything else.

PR #1819, opened 2026-06-22
fix: continuous screening use the entire dataset to screen new entity scope by object type

OrriMandarin:
## Problem

When two tables (person_A, person_B) are both mapped to the same FTM entity type (Person) and each has its own screening config with a separate inbox, the dataset-update jobs mixed matches across tables. Alerts in inbox A would surface hits on entities that originated from person_B, and vice-versa.

Root cause: all org entities are indexed into a single Yente dataset (`marble_org_{orgId}`) with no per-table discriminator, and the dataset-update query had no filter to restrict results to the tables a given config actually monitors.

## Solution

Tag every indexed Marble entity with its source table name in the programId FTM property (a keyword-type field, exact hard-filterable in ES). Then, when a per-config dataset-update job builds its search query, inject a hard filter on `properties.programId` using the config's ObjectTypes.

- models/opensanctions.go — add `ContinuousScreeningObjectTypeProperty` constant and ObjectTypes []string field on OpenSanctionsQuery
- worker_create_full_dataset.go — write programId = [objectType] on every entity in full and delta dataset files
- worker_apply_delta_file.go — strip the incoming record's own programId from scoring properties (prevents noise); set ObjectTypes from config.ObjectTypes on the query
- provider_opensanctions.go — emit filters["properties.programId"] per subquery when ObjectTypes is non-empty; gated so all other OpenSanctions callers are unaffected

The filter uses the same hard-filter mechanism already used by the LexisNexis provider. No job-level redesign needed — jobs were already per-config.

## Notes

Existing indexed dataset files do not carry programId and will need a reindex script (separate follow-up) before the fix takes full effect in production.

## Summary by CodeRabbit

* **New Features**
  * Continuous screening queries can now filter by object type, enabling restriction to specific source tables.
  * Screening entities are now tagged with their source object type for improved data organization.

* **Tests**
  * Added tests validating object type filtering and entity source tagging functionality.

### DIFF
--- a/mocks/continuous_screening_repository.go
+++ b/mocks/continuous_screening_repository.go
@@ -446,6 +446,52 @@ func (m *ContinuousScreeningRepository) SetContinuousScreeningMatchEnriched(
 	return args.Error(0)
 }
 
+func (m *ContinuousScreeningRepository) GetEnrichedContinuousScreeningUpdateJob(
+	ctx context.Context,
+	exec repositories.Executor,
+	updateId uuid.UUID,
+) (models.EnrichedContinuousScreeningUpdateJob, error) {
+	args := m.Called(ctx, exec, updateId)
+	return args.Get(0).(models.EnrichedContinuousScreeningUpdateJob), args.Error(1)
+}
+
+func (m *ContinuousScreeningRepository) UpdateContinuousScreeningUpdateJob(
+	ctx context.Context,
+	exec repositories.Executor,
+	updateId uuid.UUID,
+	status models.ContinuousScreeningUpdateJobStatus,
+) error {
+	args := m.Called(ctx, exec, updateId, status)
+	return args.Error(0)
+}
+
+func (m *ContinuousScreeningRepository) GetContinuousScreeningJobOffset(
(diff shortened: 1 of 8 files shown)


OrriMandarin:
I tested the functionality locally and it works as expected.

- UserA and UserB tables
- CS config A is configured with UserA
- CS config B is configured with UserB
- CS config A+B is configured with both UserA and UserB

For config A, there is only one match from UserA.
For config B, there is only one match from UserB.
For config A+B, there are two matches, one from UserA and one from UserB.

In my screenshot, UserA is represented as `user` and UserB as `user_bis`. Please verify the entity ID to ensure that we are using the correct entity with the appropriate case.

Capture d’écran 2026-06-22 à 18 59 15
Capture d’écran 2026-06-22 à 18 59 07
Capture d’écran 2026-06-22 à 18 58 54

apognu (approved):

Pascal-Delange (approved):
Thanks!

Pascal-Delange:
P.S.: we'll want to do a hotfix on v1.3 for this

PR #1830, opened 2026-06-26
WIP: more reliable batch exec

Pascal-Delange:
## Intent
Today's batch execution breaks at larger batch sizes.
A few elements:
- on a batch exec on 1M transactions, inserting as many decisions_to_create and river_jobs is a real tax on DB performance while those commit, which causes slowdowns for other workflows.
- there is no fast way to cancel the X00k remaining river jobs if for whatever reason the "global" job is stopped / marked as failed
- significant overhead of getting basic data out of the database, on every individual decision
- counting remaining decisions efficiently is complicated
- throughput is tightly linked to the setting of river workers per queue in the org, because every decision is a job
- object_ids matching the trigger condition filters (those that can be pre-evaluated, so simple filters on field values) are all held in memory, which causes memory usage spikes where we could easily stream

## Main changes
Introduces an alternative pathway for batch job execution (for now, feature flagger per organization) that foregoes the main inconvenients described above.
- streams the target object_ids to a bucket file
- starts a river job that processes decisions, using only one queue slot
- the job walks the file by batches of (currently) 50 objects, evaluating decisions (currently with max 10 concurrency), before writing progress advance and taking the next 50
- a global max duration for all decisions to be computed, at 24h

In terms of sequencing:
- the job has a 10min deadline, after which it snoozes itself briefly to rerun
- the relatively short duration is so that a stuck job can be quickly recovered by the river master if it becomes stuck (e.g. worker restart)
- in a 10min window, the job has time to advance multiple slices of decisions
- In case of errors, retry within the loop with exponential backoff (lightweight circuit breaker)

## Limitations to explore
- Every iteration of 50 decisions has 1min to complete. This should be plenty with 10 max concurrency, but could be blocking in very marginal cases? The value can still be discussed.
- Contrary to before, decisions are now created strictly sequentially. There is no more "peaking ahead" in case a given decision consistently fails.
   - We had early discussions about a dead letter queue, but there was no obvious idea to keep the simplicity of "one file listing decisions remaining to create" + transparent retries with circuit breaking in case of unexpected (possibly transient) errors + retry failing decisions last after all the others have been attempted
   - My impression is that we can only choose 2 out of 3: "no tx per tx bookkeeping of remaining decisions", "transparent retries on transient errors", "complete as many decisions as possible then come back for missing ones" - but we can discuss it further

### DIFF
--- a/usecases/worker_jobs/batch_execution_coordinator.go
+++ b/usecases/worker_jobs/batch_execution_coordinator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/netip"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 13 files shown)


usecases/worker_jobs/batch_execution_coordinator.go
```diff
@@ -0,0 +1,692 @@
...
+// the number of bytes consumed, so the caller can advance its byte offset. Each manifest line
+// is a Go-quoted id (see StreamAllObjectIdsFromTable); reads stop at EOF or batchSize.
+func readManifestBatch(reader *bufio.Reader, batchSize int) (ids []string, consumed int64, err error) {
+	ids = make([]string, 0, batchSize)
+	for len(ids) < batchSize {
+		line, readErr := reader.ReadString('\n')
+		if len(line) > 0 {
+			consumed += int64(len(line))
+			if trimmed := strings.TrimRight(line, "\n"); trimmed != "" {
+				id, unquoteErr := strconv.Unquote(trimmed)
```
apognu, line 430: Why do we quote/unquote the `object_id`s? Is it sure someone doesn't add a `0x20` in their IDs?

usecases/worker_jobs/run_scheduled_execution.go
```diff
@@ -219,6 +244,91 @@ func (usecase *RunScheduledExecution) ExecuteScheduledExecutionById(
 	return nil
 }
 
+// batchExecV2EnabledForOrg reports whether an organization uses manifest-based batch execution.
+// The allowlist is read from SCHEDULED_EXEC_V2_ORGS: comma-separated org UUIDs, or "*" for all.
+func batchExecV2EnabledForOrg(orgId uuid.UUID) bool {
```
apognu, line 249: FYI, we got `infra.HasFeatureFlag(flag, orgId) bool` already for this.

apognu (approved):

usecases/worker_jobs/batch_execution_coordinator.go
Pascal-Delange: Hum, my intention was, in the odd case that someone puts a line break in their IDs (which would be weird, but still)

usecases/worker_jobs/run_scheduled_execution.go
Pascal-Delange: Oh, right. I'll use that

usecases/worker_jobs/batch_execution_coordinator.go
Pascal-Delange: So I did double check, ingestion & decisions & batch exec work if the object_ids contain spaces, trailing spaces, line breaks, and other weird characters.

Pascal-Delange: On the other hand, line breaks don't show up in the decision's UI

PR #1831, opened 2026-06-26
feat: Truncate screening search with the limit

OrriMandarin:
## Problem

The Lexis Nexis provider and the Vehicle abstract type (which fans out to Airplane + Vessel) issue multiple subqueries in a single match request. OpenSanctions applies the limit per subquery, so with limit = 10 and N subqueries the merged result set could contain up to N×10 unique entities. The frontend therefore shows a non-homogeneous count (sometimes 3, sometimes 19) even though the configured limit is 10.

## Solution

The merge, dedup, and deterministic sort (score desc, EntityId asc tiebreak) were already implemented in AdaptOpenSanctionsResult. The missing step was truncating the merged list back to the configured limit.

The OpenSanctions /match response echoes back the effective limit at the top level (sibling of responses). We now read that field and slice the sorted result list down to it, so we always return at most limit entities regardless of how many subqueries were issued.

Design assumption (in code comment): all subqueries within a single match request serve the same screening purpose — they are linked variants of the same lookup. Merging and truncating their results into one list is intentional. If two independent lookups need separate result lists, they must be issued as two distinct match requests.

Load More is unaffected: it already overrides the limit to MaxScreeningCandidates (500), so the cap simply becomes 500 — no special branch needed.

## Open questions for reviewers

1. First-seen score on duplicates
When the same entity appears in multiple subquery responses (e.g. it matches both a sanctions topic and a PEP topic), we keep the score from whichever subquery happens to be iterated first. We assume the score is identical for the same query params regardless of which topic filter was applied. Is that assumption always valid in the screening service, or can the same entity receive different scores from different subqueries within the same request?

## Summary by CodeRabbit

* **Bug Fixes**
  * Screening results now respect the limit returned by OpenSanctions, so only the top-ranked matches are shown.
  * Multi-query OpenSanctions responses are now merged and trimmed correctly, keeping result counts and displayed matches aligned.

* **Tests**
  * Added coverage for multi-query OpenSanctions responses to verify truncation and ordering behavior.


Pascal-Delange:
> We assume the score is identical for the same query params regardless of which topic filter was applied. Is that assumption always valid in the screening service, or can the same entity receive different scores from different subqueries within the same request?

valid

Pascal-Delange (approved):
👍

PR #1841, opened 2026-07-02
feat: marble server can serve prompt to other marble instance

OrriMandarin:
## Summary

Adds a new endpoint so Marble SaaS can serve the private AI prompt bundle to self-hosted premium Marble instances over HTTP, replacing manual prompt-file setup for self-hosted deployments. Prompts stay private (SaaS-only endpoint, license-gated) while self-hosted clients can fetch exactly the version compatible with their own release.

Companion change (GCS upload side): https://github.com/checkmarble/marble-llm-prompt/pull/38

## GCS directory structure

Prompt bundles are published into GCS under a folder per product **Major.Minor** version:

```
<bucket>/
  v1.3/
    case_review/...
    kyc_enrichment/...
    rule/...
    screening_hit_suggestion/...
    system.md
    ai_agent_models.json
  v1.4/
    ...same layout, this Major.Minor's content...
  case_review/          <- flat, unversioned (see below)
  system.md
  ai_agent_models.json
  ...
```

- One folder per Major.Minor (not per patch): every CI publish targeting the same Major.Minor **overwrites** that folder, so "latest for the Major.Minor" is automatic — there's no patch level to track or compare.
- Versioning follows the *product's* own release tags (truncated to Major.Minor by the prompts-repo CI), not an independent numbering. Not every release needs a new prompt folder — if prompt content didn't change, a version can be skipped entirely.
- The **flat files at the bucket root are untouched** by this change — they're read directly by Marble's own SaaS backend for its own internal AI feature usage, a separate concern from this endpoint. A later PR will migrate that internal usage onto the versioned layout too and retire the flat files.
- This endpoint **never reads the flat files**, only versioned folders — see resolution logic below.

## `PROMPTS_DIR` and the GCS mount

`PROMPTS_DIR` (new env var, wired via `usecases.WithPromptsDir`) tells the server where to read prompts from. On Marble SaaS, this points at the **GCS bucket mounted directly into the container** (GCS FUSE / a Cloud Run volume mount) — not something fetched over the network per request. `PromptServingUsecase` just walks that mounted directory tree with regular filesystem calls (`os.ReadDir`, `filepath.WalkDir`), so serving a request is a cheap local read, not a GCS API call.

If `PROMPTS_DIR` isn't configured at all, every request fails with `models.MissingRequirement` — this is treated as a Marble-side deployment misconfiguration (not a per-client condition), since it means the SaaS forgot to mount/configure its own prompts bucket.

## Endpoint & resolution

```
GET /ai-prompts/download?prompts_version=<major.minor>
Authorization: Bearer <license-key>
```

- SaaS-only: registered inside the existing `infra.IsMarbleSaasProject()`-gated block in `api/routes.go`.
- Auth: license key via `Authorization: Bearer`, validated and required to carry the `CaseAiAssist` entitlement — a single flag gates the whole prompt bundle, no per-resource granularity.
- `prompts_version` is **required**, always the caller's own product Major.Minor (no manual pin/override — self-hosted clients auto-detect their own build version). Resolution (`PromptServingUsecase.resolvePromptsBase`):
  - Exact Major.Minor match → served.
  - No exact match → the **nearest earlier published** Major.Minor is used instead (a backward search; can cross a major boundary, since a product major bump doesn't imply a prompt-breaking change). Never resolves *forward* to something newer than requested.
  - Nothing available `<=` the requested version → `404`.

## Response

`Content-Type: application/zip`. The body contains **everything found under the resolved version folder, as-is** — there's no fixed resource list to keep in sync with what the prompts repo publishes, so a new version can introduce a brand-new folder or file and it's included automatically without any backend code change. No manifest or version metadata is embedded in the response: the caller already knows exactly which version it asked for.

## How a Marble client should use this (self-hosted, client-side work tracked separately)

The self-hosted client integration isn't built in this PR, but the intended usage is:

1. On every process start (server and worker each do this independently), call this endpoint with the client's own detected product Major.Minor.
2. The response is a zip byte stream — unzip it **directly into memory**. Go's `archive/zip.Reader`, constructed over the downloaded bytes, is itself an `fs.FS`, so no temp files or disk writes are needed at all.
3. Use that in-memory `fs.FS` as the prompt source for the lifetime of that process. No local caching/persistence across restarts by design — the payload is small and the read is cheap, so simplicity wins over avoiding a repeat fetch on the next start.

## Testing

- `usecases/prompt_serving_usecase_test.go`: license/entitlement gating, exact-match and backward-search version resolution (including crossing a major boundary and failing closed when nothing qualifies), path-traversal/malformed-input rejection, and that the zip includes whatever is actually present in the resolved directory (including a brand-new, previously-unlisted folder).
- `go build`, `go vet`, `golangci-lint run`, full `go test ./...` all pass.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated endpoint to download AI prompt bundles as a streamed ZIP (`GET /ai-prompts/download`), requiring a `prompts_version` query parameter.
  * Made prompt bundle storage configurable at runtime via `PROMPTS_DIR`.
* **Bug Fixes**
  * Enforced license entitlement checks before serving downloads.
  * Updated version handling to Major.Minor semantics (rejects patch versions) with nearest-earlier selection and no fallback to legacy layouts; added safer validation for invalid/unsafe versions.
* **Tests**
  * Added coverage for authorization handling, version resolution rules, and ZIP contents.

### DIFF
--- a/api/handle_ai_prompts.go
+++ b/api/handle_ai_prompts.go
@@ -51,7 +51,6 @@ func handleAiPromptsDownload(uc usecases.Usecases) gin.HandlerFunc {
 		}
 
 		c.Header("Content-Type", "application/zip")
-		c.Header("Content-Disposition", `attachment; filename="ai-prompts.zip"`)
 		if _, err := io.Copy(c.Writer, zipReader); err != nil {
 			presentError(ctx, c, err)
 			return
--- a/api/handle_ai_prompts.go
+++ b/api/handle_ai_prompts.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/cockroachdb/errors"
 	"github.com/gin-gonic/gin"
 
 	"github.com/checkmarble/marble-backend/models"
@@ -11,8 +12,7 @@ import (
 	"github.com/checkmarble/marble-backend/utils"
 )
 
-// aiPromptsLicenseKey extracts the caller's license key from the "Authorization: Bearer <key>"
-// header.
+// Extract the caller's license
 func aiPromptsLicenseKey(c *gin.Context) (string, error) {
 	key, err := utils.ParseAuthorizationBearerHeader(c.Request.Header)
 	if err != nil {
@@ -25,11 +25,10 @@ func aiPromptsLicenseKey(c *gin.Context) (string, error) {
 }
 
 // handleAiPromptsDownload streams a zip of the whole AI prompt bundle the caller's license is
-// entitled to, for the exact prompts_version requested (the caller always asks for one precise
-// version — its own hardcoded default, or an operator-configured pin — never "give me the
-// latest"; see PromptServingUsecase for the resolution rules). The zip contains only the
-// prompt files themselves — no manifest or version is reported back, since the caller already
-// knows which version it asked for. Only reachable on Marble SaaS.
+// entitled to, resolved for the caller's prompts_version (its own product Major.Minor).
+// Resolution is a backward search: exact Major.Minor
+// match wins, else the nearest earlier published one; never a newer one. The zip contains only the prompt files themselves.
+// Only reachable on Marble SaaS.
 func handleAiPromptsDownload(uc usecases.Usecases) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		ctx := c.Request.Context()
@@ -40,6 +39,10 @@ func handleAiPromptsDownload(uc usecases.Usecases) gin.HandlerFunc {
 		}
 
 		version := c.Query("prompts_version")
+		if version == "" {
+			presentError(ctx, c, errors.Wrap(models.BadParameterError, "prompts_version in query is required"))
+			return
+		}
 
 		usecase := uc.NewPromptServingUsecase()
 		zipReader, err := usecase.DownloadPrompts(ctx, licenseKey, version)
--- a/api/handle_ai_prompts.go
+++ b/api/handle_ai_prompts.go
@@ -1,0 +1,58 @@
+package api
+
+import (
+	"io"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/usecases"
+	"github.com/checkmarble/marble-backend/utils"
+)
+
+// aiPromptsLicenseKey extracts the caller's license key from the "Authorization: Bearer <key>"
+// header.
+func aiPromptsLicenseKey(c *gin.Context) (string, error) {
+	key, err := utils.ParseAuthorizationBearerHeader(c.Request.Header)
+	if err != nil {
+		return "", err
+	}
+	if key == "" {
+		return "", models.UnAuthorizedError
+	}
+	return key, nil
+}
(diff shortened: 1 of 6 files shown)


apognu:
> On Marble SaaS, this points at the GCS bucket mounted directly into the container (GCS FUSE / a Cloud Run volume mount) — not something fetched over the network per request. PromptServingUsecase just walks that mounted directory tree with regular filesystem calls (os.ReadDir, filepath.WalkDir), so serving a request is a cheap local read, not a GCS API call.

Just to be clear, any operation on the GCS fuse mount point **is** one (or several) network call(s), the fuse driver simply translates the syscall into network packets. They're basically the same thing, just wrapped differently.

That means that, potentially, "simple" filesystem operations can become costly when performed here. It shouldn't be too impactful for this usecase where there are not a lot of files. But `os.ReadDir` and `filepath.WalkDir` are examples of things that can get unwieldy because they can perform **a lot** of network requests to get all the files in a large bucket.

api/handle_ai_prompts.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		usecase := uc.NewPromptServingUsecase()
+		zipReader, err := usecase.DownloadPrompts(ctx, licenseKey, version)
+		if presentError(ctx, c, err) {
+			return
+		}
+
+		c.Header("Content-Type", "application/zip")
+		c.Header("Content-Disposition", `attachment; filename="ai-prompts.zip"`)
```
apognu: Why do we need `content-disposition`? This is not going to be used by a browser, is it?

usecases/prompt_serving_usecase.go
```diff
@@ -0,0 +1,206 @@
...
+	baseDir, err := uc.resolvePromptsBase(version)
+	if err != nil {
+		return nil, err
+	}
+
+	var buf bytes.Buffer
+	zw := zip.NewWriter(&buf)
+
+	if err := addDirToZip(zw, baseDir, ""); err != nil {
+		return nil, err
```
apognu, line 93: The zip writer and its buffer might leak if an error happens here. The `Close()` method should be called in a defer statement.

usecases/prompt_serving_usecase.go
```diff
@@ -0,0 +1,206 @@
...
+
+	entries, err := os.ReadDir(uc.promptsDir)
+	if err != nil {
+		return "", errors.Wrapf(err, "could not list prompts directory %s", uc.promptsDir)
+	}
+
+	bestName := ""
+	bestMajor, bestMinor := -1, -1
+	for _, e := range entries {
+		if !e.IsDir() || !promptVersionDirPattern.MatchString(e.Name()) {
```
apognu: Callback to my general comment about how fuse works: did you try with an actual `gcsfuse` mount? I know it does not implement all of POSIX, but not sure which part of the standard it does implement.

Let's give it a run locally with an actual production setup to check if everything we do is supported?

OrriMandarin: I tested the code locally by installing `macFuse` to mount the bucket. 
I also added some logs to capture the output of the `isDir()` function, which confirms that the directories are being correctly identified.

apognu (approved):

PR #1843, opened 2026-07-03
continuous screening observability

william-schlegel:
# implement observability endpoints for continuous screening workers

## feat: add endpoint to list continuous screening dataset updates

- Implemented a new handler for listing continuous screening dataset updates with pagination and sorting.
- Added corresponding route in the API for accessing the dataset updates.
- Introduced a new DTO for continuous screening dataset updates to structure the response data.
- Updated repository and use case layers to support fetching dataset updates from the database.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated continuous-screening API endpoints for dataset updates, update job status, and client data indexing.
  * Introduced new enriched response payloads including timing/config details, totals, optional item counts, and structured job errors.
  * Implemented validated cursor/keyset pagination with reliable next-page detection, plus catalog-driven enrichment of dataset version/index metadata.
* **Bug Fixes**
  * Ensured update jobs are marked as **Processing** for observability when not already in that state.


usecases/continuous_screening/dataset_update.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/pure_utils"
+)
+
+func (uc *ContinuousScreeningUsecase) ListContinuousScreeningDatasetUpdates(
+	ctx context.Context,
+	pagination models.PaginationAndSorting,
+) ([]models.ContinuousScreeningDatasetUpdateSummary, error) {
+	if err := models.ValidatePagination(pagination); err != nil {
```
apognu, line 22: We usually get one more item than requested so we can know whether or not there is a next page, then strip it out of the slice. We then use this DTO to convey the information to the frontend:

https://github.com/checkmarble/marble-backend/blob/f8f50a911158e50e4edc5c34929d53146110bf3b/dto/pagination.go#L21-L24

models/continuous_screening.go
```diff
@@ -176,6 +176,16 @@ type ContinuousScreeningDatasetUpdate struct {
 	CreatedAt     time.Time
 }
 
+// ContinuousScreeningDatasetUpdateSummary is a trimmed view of a dataset update,
+// exposing only the fields relevant to a monitoring/history listing.
+type ContinuousScreeningDatasetUpdateSummary struct {
```
OrriMandarin: I wonder if we need to define this struct since we already have a similar one called `ContinuousScreeningDatasetUpdate`. when adapting to the DTO, we can simply ignore the `DeltaFilePath` field.

OrriMandarin:
I’m thinking about our SaaS client. 
In our SaaS platform, we manage LexisNexis and OpenSanctions datasets. 
However, perhaps an organization configured with OpenSanctions might not need to see when the LexisNexis dataset update is detected.

I’m uncertain whether we should implement a filter to select the specific dataset name we want to return or to base on the organization screening provider config

models/continuous_screening.go
william-schlegel: ok, if the good practice is to avoid close structures when it is possible, I will change it

william-schlegel:
> I’m thinking about our SaaS client. In our SaaS platform, we manage LexisNexis and OpenSanctions datasets. However, perhaps an organization configured with OpenSanctions might not need to see when the LexisNexis dataset update is detected.
> 
> I’m uncertain whether we should implement a filter to select the specific dataset name we want to return or to base on the organization screening provider config

there is a provider field in the jobs, the filter should be easy to apply

usecases/continuous_screening/usecase.go
```diff
@@ -67,6 +67,23 @@ type ContinuousScreeningUsecaseRepository interface {
...
+		exec repositories.Executor,
+		orgId uuid.UUID,
+		pagination models.PaginationAndSorting,
+	) ([]models.ContinuousScreeningUpdateJobSummary, error)
+	ListContinuousScreeningClientDataIndexing(
+		ctx context.Context,
+		exec repositories.Executor,
+		orgId uuid.UUID,
+		pagination models.PaginationAndSorting,
+	) ([]models.ContinuousScreeningClientDataIndexingSummary, error)
```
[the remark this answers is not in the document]
william-schlegel: No column to scope on — the table is unique (dataset_name, version) with no org_id or provider column, unlike its siblings' continuous_screening_update_jobs table (which has both, hence those endpoints scope by org).

usecases/continuous_screening/dataset_update.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, name := range catalog.Current {
+		current[name] = struct{}{}
+	}
+	for i := range updates {
+		if dataset, ok := catalog.Datasets[updates[i].DatasetName]; ok {
+			updates[i].Title = dataset.Title
+			updates[i].LiveVersion = dataset.Version
+			_, updates[i].IsCurrent = current[updates[i].DatasetName]
+		}
+	}
```
OrriMandarin, line 66: There might be an error if the organisation has dataset updates from LexisNexis and OpenSanctions in the same paginated results.  Since you only get the catalogs for one provider, updates from another provider won’t have the same enriched information.

OrriMandarin:
For a given dataset update row, an organisation can have multiple update jobs (one per configuration).

Here are a few notes:
- I believe it would be more efficient to fetch all update jobs (for all configurations) rather than just those linked to a specific dataset update. This way users could easily see the status of each CS configuration for a dataset version.
- Not linked to this PR: When a new dataset version is detected, we launch a worker if the organisation has an active configuration.  We need to check if the update should proceed for the organisation based on the org provider setting.  It’s not necessary to process an opensanctions dataset if the organisation only uses `lexis nexis`.  fyi @Pascal-Delange

api/routes.go
```diff
@@ -215,6 +215,12 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 		handleDeleteContinuousScreeningObject(uc))
 	router.PATCH("/continuous-screenings/matches/:id", tom,
 		handleUpdateContinuousScreeningMatchStatus(uc))
+	router.GET("/continuous-screenings/dataset-updates", tom,
+		handleListContinuousScreeningDatasetUpdates(uc))
+	router.GET("/continuous-screenings/update-jobs", tom,
+		handleListContinuousScreeningUpdateJobs(uc))
+	router.GET("/continuous-screenings/client-data-indexing", tom,
+		handleListContinuousScreeningClientDataIndexing(uc))
```
OrriMandarin, line 222: I didn't understand the difference between these two endpoints except that the last one returns a trimmed version.

OrriMandarin: I mean, using the first one should be enough, did we need the second one?

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,359 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+// The job status is left-joined via a LATERAL aggregate scoped to the dataset update, org and
+// provider: a dataset update may have several jobs (one per config), and a plain join would
+// multiply rows and break the keyset pagination window.
+func (repo *MarbleDbRepository) ListContinuousScreeningDatasetUpdates(
+	ctx context.Context,
+	exec Executor,
+	orgId uuid.UUID,
+	provider models.ScreeningProvider,
+	pagination models.PaginationAndSorting,
+) ([]models.ContinuousScreeningDatasetUpdateEnriched, error) {
```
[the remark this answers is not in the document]
william-schlegel: is it something common in other files from this package?

repositories/migrations/20260720160000_add_provider_to_continuous_screening_dataset_files.sql
```diff
@@ -0,0 +1,20 @@
+-- +goose Up
+
+alter table continuous_screening_dataset_files
```
Pascal-Delange: Hmm no this is wrong, the provider should be on the continuous_screening_dataset_updates table (where it currently isn't, only on update_jobs).
continuous_screening_dataset_files is the table that contains delta files generated from udpated Marble data to index them.

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+			"MIN(dt.id::text)::uuid AS id",
+			"df.created_at AS job_date",
+			"COUNT(*) AS total_items",
+			"df.version AS version",
+			"dt.object_type AS object_type",
+		).
+		From(dbmodels.TABLE_CONTINUOUS_SCREENING_DELTA_TRACKS+" AS dt").
+		InnerJoin(dbmodels.TABLE_CONTINUOUS_SCREENING_DATASET_FILES+
+			" AS df ON df.id = dt.dataset_file_id AND df.org_id = dt.org_id"+
+			" AND df.file_type = ? AND df.provider = ?",
```
Pascal-Delange: No need for provider here, cf other comment

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+	query := NewQueryBuilder().
+		Select(
+			"MIN(dt.id::text)::uuid AS id",
+			"df.created_at AS job_date",
+			"COUNT(*) AS total_items",
+			"df.version AS version",
+			"dt.object_type AS object_type",
+		).
+		From(dbmodels.TABLE_CONTINUOUS_SCREENING_DELTA_TRACKS+" AS dt").
+		InnerJoin(dbmodels.TABLE_CONTINUOUS_SCREENING_DATASET_FILES+
```
Pascal-Delange: Ok, this would work for now. However:
IMO we should handle the delta tracks table as a table that can be cleaned up eventually, so it would really be better if we'd store a number of entries per delta file in the table rather than need to join the delta tracks (cc @OrriMandarin ).

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+
+	var count int
+	if err := exec.QueryRow(ctx, sql, args...).Scan(&count); err != nil {
+		return 0, errors.Wrap(err, "error counting pending client data indexing items")
+	}
+
+	return count, nil
+}
+
+func continuousScreeningClientDataIndexingPendingQuery(
```
Pascal-Delange, line 1190: That method is unnecessary indirection IMO (and there are maybe a few other examples of this in this file)

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+	orgId uuid.UUID,
+	provider models.ScreeningProvider,
+	indexVersion *string,
+) squirrel.SelectBuilder {
+	query := NewQueryBuilder().
+		Select("COUNT(*)").
+		From(dbmodels.TABLE_CONTINUOUS_SCREENING_DELTA_TRACKS+" AS dt").
+		LeftJoin(dbmodels.TABLE_CONTINUOUS_SCREENING_DATASET_FILES+
+			" AS df ON df.id = dt.dataset_file_id AND df.org_id = dt.org_id"+
+			" AND df.file_type = ? AND df.provider = ?",
```
Pascal-Delange: see remark about provider

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+}
+
+func continuousScreeningClientDataIndexingPendingQuery(
+	orgId uuid.UUID,
+	provider models.ScreeningProvider,
+	indexVersion *string,
+) squirrel.SelectBuilder {
+	query := NewQueryBuilder().
+		Select("COUNT(*)").
+		From(dbmodels.TABLE_CONTINUOUS_SCREENING_DELTA_TRACKS+" AS dt").
```
Pascal-Delange, line 1197: Actually: if this is the query used to display what I'm thinking of, the goal here was rather to count unprocessed delta tracks, so delta tracks without a dataset_file_id. And doing a "limit 1000" or so in a subquery so we don't risk to run a very large count query on every page load.

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,430 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+			"ucs.org_id":   orgId,
+			"ucs.provider": provider,
+		})
+}
+
+type continuousScreeningPaginationOffset struct {
+	Value time.Time `db:"offset_value"`
+}
+
+func adaptContinuousScreeningPaginationOffset(
```
Pascal-Delange: could be an inline function

usecases/continuous_screening/worker_create_full_dataset.go
```diff
@@ -208,7 +208,8 @@ func (w *CreateFullDatasetWorker) Work(ctx context.Context,
 	}()
 
 	// Check if the org has a continuous screening config
-	configs, err := w.repo.GetContinuousScreeningConfigsByOrgId(ctx, exec, orgId, org.GetScreeningProviderFor(models.ScreeningFeatureContinuousMonitoring))
+	provider := org.GetScreeningProviderFor(models.ScreeningFeatureContinuousMonitoring)
```
Pascal-Delange, line 211: See remark in other places about provider

Pascal-Delange (changes requested):
Not totally done reading, but here is some intermediate feedback. Nothing very critical but may take a bit of time to fix, so actually let's not push this into this release (sorry again, my fault)

usecases/continuous_screening/dataset_update.go
Pascal-Delange: Hmm that seems correct.
However we would really want to filter to get only one provider.

usecases/continuous_screening/dataset_update.go
Pascal-Delange: I think that is done now

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,434 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+		), 0)::int AS items_processed`,
+		"(COUNT(ucs.id) * ds.total_items)::int AS items_total",
+		`CASE
+			WHEN COUNT(*) FILTER (WHERE ucs.status = 'failed') > 0 THEN 'failed'
+			WHEN COUNT(*) FILTER (WHERE ucs.status = 'pending') > 0 THEN 'pending'
+			WHEN COUNT(*) FILTER (WHERE ucs.status = 'processing') > 0 THEN 'processing'
+			WHEN COUNT(*) FILTER (WHERE ucs.status = 'completed') > 0 THEN 'completed'
+			ELSE 'pending'
+		END AS status`,
+	).
```
OrriMandarin, line 888: I wonder if we can avoid using count a second time in the query. 
May be we can use the `completed_count`/`processing_count`/pending_count` in the outer query and have something like 

CASE
  job.completed_count > 0 ...
  job.processing_count > 0 ...

dto/continuous_screening_dto.go
```diff
@@ -9,6 +9,144 @@ import (
...
+		Details:   e.Details,
+		CreatedAt: e.CreatedAt,
+	}
+}
+
+type ContinuousScreeningUpdateJobDto struct {
+	Id             uuid.UUID                        `json:"id"`
+	Status         string                           `json:"status"`
+	JobStart       time.Time                        `json:"job_start"`
+	JobEnd         time.Time                        `json:"job_end"`
```
OrriMandarin: Do we really need to rename the `created_at` and `updated_at` fields to `job_start` and `job_end`?

The `job_end` field will always have a value, but it doesn’t necessarily represent the end of the job (it depends on the status). I believe we can retain the `created_at` and `updated_at` fields, but we can use different terms on the front end to represent them.

repositories/continuous_screening_repository.go
OrriMandarin: It depends on whether we want to keep a record of which delta files my entity has been indexed or a proof of how we generate an indexation file for the organization.

If not, yes, we can consider this table as a cleanable table and need to store the number of entries we have in each file.

And it avoids a join :)

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,434 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+	)
+}
+
+func (repo *MarbleDbRepository) ListContinuousScreeningClientDataIndexing(
+	ctx context.Context,
+	exec Executor,
+	orgId uuid.UUID,
+	indexVersion *string,
+	pagination models.PaginationAndSorting,
+) (models.ContinuousScreeningClientDataIndexing, error) {
```
OrriMandarin, line 1080: I’m not a fan of our current listing method. We use delta tracks as the entry point for pagination (the pagination ID is a delta track, and we use a MIN(ID) trick to paginate, etc.).

This  should be anchored on `continuous_screening_dataset_files`, not `continuous_screening_delta_tracks`. The endpoint lists delta files generated — the file is the entity being paginated, delta tracks are just what's being counted.

The endpoint's job is to list the **delta files that were generated**, with how many items each contains — the file is the entity being paginated, the delta tracks are just what we're counting. Using `dt` as the `FROM` and joining to `df` gets the relationship backwards
It also forces the query into an awkward corner: since `dt` is the anchor, there's no natural "id" for the aggregated row, so it currently fakes one via `MIN(dt.id::text)::uuid AS id`, picking an arbitrary delta-track id to stand in for the file. That's more machinery than needed once the file is the anchor: `df.id` is already the real, stable identifier.

usecases/continuous_screening/worker_create_full_dataset.go
```diff
@@ -246,7 +247,7 @@ func (w *CreateFullDatasetWorker) Work(ctx context.Context,
 // Create the first full dataset for the org.
 // The flow is simpler than the patch dataset and we don't need to create a delta dataset file.
 func (w *CreateFullDatasetWorker) handleFirstFullDataset(ctx context.Context,
-	exec repositories.Executor, orgId uuid.UUID,
+	exec repositories.Executor, orgId uuid.UUID, provider models.ScreeningProvider,
```
OrriMandarin: provider is not used

usecases/continuous_screening/worker_apply_delta_file.go
```diff
@@ -232,6 +232,18 @@ func (w *ApplyDeltaFileWorker) Work(ctx context.Context, job *river.Job[models.C
...
+		if err := w.repository.UpdateContinuousScreeningUpdateJob(
+			ctx,
+			exec,
+			updateJob.Id,
+			models.ContinuousScreeningUpdateJobStatusProcessing,
+		); err != nil {
+			return err
+		}
+	}
+
```
OrriMandarin, line 246: Why do you need to change the status?

repositories/continuous_screening_repository.go
```diff
@@ -812,6 +812,434 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+	exec Executor,
+	orgId uuid.UUID,
+	provider models.ScreeningProvider,
+	pagination models.PaginationAndSorting,
+) ([]models.ContinuousScreeningDatasetUpdateEnriched, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, err
+	}
+
+	query, err := continuousScreeningDatasetUpdatesQuery(orgId, provider, pagination)
```
OrriMandarin, line 835: I remembered mentioning the need to filter dataset updates by provider. However, I believe this approach is not suitable. Instead, we should consider all dataset updates that have at least one update job assigned to the organization. Filtering by provider is unnecessary because assigning an update job implies that the organization has already processed that version.

Apologies for that call.

I also recall noting a bug in the `worker_scan_dataset_update` function. We haven’t ensured that the organization configuration provider allows the organization to process the update and that we only spawn a job if the update should be treated for that organization. I’ll try to fix this issue. This means that for a given organization and dataset version, we shouldn’t have any jobs linked to it.

OrriMandarin:
I recalled that we want to add the `provider` information in the `dataset_updates` table, I didn't see this change, have you changed your mind?

repositories/continuous_screening_repository.go
OrriMandarin: The fix: https://github.com/checkmarble/marble-backend/pull/1894

repositories/continuous_screening_repository.go
Pascal-Delange: I wouldn't worry too much about it, the pg query engine knows how to optimize this (in principle)

dto/continuous_screening_dto.go
william-schlegel: I used the same terminology than in the front end to make a clear match between the API response and the consumer

usecases/continuous_screening/worker_create_full_dataset.go
william-schlegel: I guess I missed this one

usecases/continuous_screening/worker_apply_delta_file.go
william-schlegel: because it is displayed in the front end, and sometimes it's not accurate (when the number of records to process is below the batch size I think, so there is no update at all

OrriMandarin: Ah!!! I defined this status but I never use it ahah. Nice !

dto/continuous_screening_dto.go
OrriMandarin: The issue here is that the `updated_at` timestamp doesn’t always signify the completion of a job. When changing the status, it updates the `updated_at` timestamp, but the status itself doesn’t necessarily indicate a final status.

william-schlegel: Ok, I will add job start and job end columns in the table. job start is set when the job process starts end job end is set when it is finished
if the job is skipped, there will be no start, but an end (not processed and finished anyway)

repositories/continuous_screening_repository.go
william-schlegel: The pagination is now common for both dataset updates and dataset per config updates
not sure it is relevant to modify it for now

repositories/migrations/20260727120000_add_continuous_screening_job_timings.sql
```diff
@@ -0,0 +1,28 @@
...
+ADD COLUMN finished_at TIMESTAMP WITH TIME ZONE;
+
+-- Backfill: created_at over-estimates the start by however long the job sat queued, but it is the
+-- only signal left for rows that already reached a terminal status. finished_at is only meaningful
+-- for terminal rows: for a row still in 'processing', updated_at is the moment it *started*, so
+-- writing it as an end would make an in-flight job render as finished.
+UPDATE continuous_screening_update_jobs
+SET started_at = created_at,
+    finished_at = CASE WHEN status IN ('completed', 'failed', 'skipped') THEN updated_at END
+WHERE status <> 'pending';
```
OrriMandarin, line 17: I believe we only launch this job once a day from decembre 2025, so we shouldn’t have a lot of rows to update. Okay.

repositories/continuous_screening_repository.go
```diff
@@ -908,13 +1337,39 @@ func (repo *MarbleDbRepository) UpdateContinuousScreeningUpdateJob(
 		return err
 	}
 
+	return ExecBuilder(ctx, exec, continuousScreeningUpdateJobStatusQuery(updateId, status))
+}
+
+// continuousScreeningUpdateJobStatusQuery derives the job timing columns from the status being
+// written: a job starts when it enters `processing` and ends when it reaches a terminal status.
+// Neither can be inferred from created_at (enqueue time) or updated_at (bumped on every write).
+func continuousScreeningUpdateJobStatusQuery(
```
OrriMandarin: Don't think we need to create a new function for the query instead writing the query directly in the `UpdateContinuousScreeningUpdateJob` method

repositories/continuous_screening_repository.go
```diff
@@ -814,6 +814,435 @@ func (repo *MarbleDbRepository) GetLastProcessedVersion(
...
+
+	return models.ContinuousScreeningClientDataIndexing{
+		PendingItems: pendingItems,
+		Items: models.Paginated[models.ContinuousScreeningClientDataIndexingSummary]{
+			Items: items,
+		},
+	}, nil
+}
+
+func continuousScreeningClientDataIndexingAggregateQuery(
```
OrriMandarin: I still believe that we should use the `TABLE_CONTINUOUS_SCREENING_DATASET_FILES` as the reference table to list the indexed data of the organization. Instead of referring to `TABLE_CONTINUOUS_SCREENING_DELTA_TRACKS`, which requires a trick with the `min(dt.id)` to obtain the cursor ID, we should use the `TABLE_CONTINUOUS_SCREENING_DATASET_FILES`.

OrriMandarin (approved):
Ok for me if it works like it is.
The SQL queries are very difficult to read 🫪

repositories/continuous_screening_repository.go
william-schlegel: it was there to be testable, but it is maybe not necessary. I simplify this

repositories/continuous_screening_repository.go
william-schlegel: ok, I can change that

Pascal-Delange (approved):
Have not re reviewed but good for me if it is for Leo !

PR #1856, opened 2026-07-07
Batch upload UX improvements (errors and public API)

apognu:
* We need to present CSV input errors when asynchronously uploading a file for ingestion. This commit separate internal errors (still stored, but not exposed) and input errors that will be returned.
* Also, add an endpoint to the public API to list ingestion logs.

## Summary by CodeRabbit

* **New Features**
  * Added a new upload/ingestion log view with filtering and pagination.
  * Upload log entries now include both processing progress and input-validation error details.

* **Bug Fixes**
  * Improved progress counts so reported rows processed are more accurate.
  * Fixed identifier handling to use UUIDs consistently, reducing mismatches across ingestion flows.

### DIFF
--- a/dto/upload_log_dto.go
+++ b/dto/upload_log_dto.go
@@ -21,11 +21,13 @@ func AdaptUploadLogDto(log models.UploadLog) UploadLogDto {
 		error = *log.Error
 	}
 
+	linesProcessed := max(log.LinesProcessed, log.RowsIngested)
+
 	return UploadLogDto{
 		Status:          string(log.UploadStatus),
 		StartedAt:       log.StartedAt,
 		FinishedAt:      log.FinishedAt,
-		LinesProcessed:  log.LinesProcessed,
+		LinesProcessed:  linesProcessed,
 		NumRowsIngested: log.RowsIngested,
 		Error:           error,
 	}
--- a/dto/upload_log_dto.go
+++ b/dto/upload_log_dto.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/utils"
 )
 
 type UploadLogDto struct {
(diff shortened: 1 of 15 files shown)


dto/upload_log_dto.go
```diff
@@ -18,14 +18,16 @@ type UploadLogDto struct {
 func AdaptUploadLogDto(log models.UploadLog) UploadLogDto {
 	error := ""
 	if log.Error != nil {
```
OrriMandarin:
```suggestion
	if log.InputError != nil {
```

??

pubapi/v1/routes.go
```diff
@@ -97,6 +97,7 @@ func BetaRoutes(conf pubapi.Config, unauthed *gin.RouterGroup, authMiddleware gi
 		root.GET("/cases/:caseId/ai_reviews/:aiReviewId", HandleGetAiCaseReviewById(uc))
 
 		root.PUT("/ingest/:objectType/upload", v1beta.HandleUploadCsv(uc))
+		root.GET("/ingest/:objectType", v1beta.HandleBatchIngestionLog(uc))
```
OrriMandarin: I’m not sure about the route naming. Should we add the `log` term somewhere to indicate that this endpoint is for retrieving logs and not ingested data?

On our internal endpoint, we have `/upload-logs`, which is more explicit and descriptive.

apognu: I think we wanted to move away from the `log` term, which is very technical and not entirely clear.

It also does not return all ingestion actions, only some of them, and not even all batches, only the CSV ones. But I would also prefer not to have "CSV" leak in the interface, if we move away from it at some point.

Even the other endpoint, ending in `/upload`, I am not entirely satisfied with, since it doesn't really describe what it does.

Happy to discuss further tomorrow.

OrriMandarin: It could be more general and high-level if we use something like `history` or `list`.

apognu: Arguably, those have the same issue of implying it's the list or history of ingested records.

apognu: Chosen to have symmetry between create and list, with `/uploads`. Not perfect, but better.

OrriMandarin (approved):

PR #1865, opened 2026-07-09
POC: Model Context Protocol (MCP) server for case/decision reporting

OrriMandarin:
## Summary
- Adds a proof-of-concept MCP server, mounted at `/mcp` in the existing Gin router (`api/routes.go`), reusing the same `auth.AuthedBy(utils.PublicApiKey, utils.ApiKeyAsBearerToken)` middleware as the public API v1 group — every tool call is org-scoped via the caller's Marble API key, exactly like `pubapi/v1`.
- Goal: let Claude (or any MCP client) answer natural-language reporting/analytics questions and perform a couple of write actions against case/decision data, without building a bespoke dashboard.
- New package `mcp/` registers 8 tools:
  - `list_users`, `list_cases` — thin wrappers on existing repos/usecases, no new SQL.
  - `case_analytics_query` — dispatches to the existing `CaseAnalyticsUsecase` (mean case-resolution time derivable from `cases_duration`, no new SQL).
  - `case_assignment_stats` — new `CasesCountByAssignee` repo method + usecase method, joined with `list_users` for name/email.
  - `decision_stats_query` — new `DecisionStatsByDayOutcomeUser` repo method + usecase method; the tool pivots results by day/outcome/assignee per the caller's requested `group_by`.
  - `entity_overview` — "customer hub" lookup composing `ScoringScoresUsecase`, `CaseUseCase` (entity-related + continuous-screening cases), `EntityAnnotationUsecase`, and `GetContinuousScreeningByObjectId`, with slim output DTOs (see note below).
  - `case_bulk_assign` — batch assign/unassign, built on `CaseUseCase.MassUpdate` with a new `unassign` action (added to `models.CaseMassUpdateAction`, `dto.CaseMassUpdateDto`, and `repositories.CaseMassUnassign`). This also enables `"unassign"` on the existing `POST /cases/mass-update` REST endpoint as a small side effect.
  - `whoami` — diagnostic tool reporting the org/role the current API key resolves to.
- Credential propagation: Gin's auth middleware attaches `models.Credentials` to the request context before the request reaches the MCP HTTP handler (`gin.WrapH`); the go-sdk's session is rooted in that same context at `initialize` time, so every tool handler can call `pubapi.UsecasesWithCreds(ctx, uc)` exactly like any other handler. Verified empirically with two separate orgs/API keys — no cross-org data leakage on read or write paths.
- One schema gotcha worth flagging for future tools: `uuid.UUID` fields serialize as `[16]byte` under the SDK's JSON-schema inference (not as strings), which breaks both input validation and output validation. Fixed by using plain `string` for UUIDs in tool I/O structs (parsed/formatted manually) instead of `uuid.UUID`.

## Test plan
- [x] `go build ./...` and `go test ./...` pass.
- [x] `gofmt -l` clean on all changed/new files.
- [x] Manual end-to-end verification against a local server + Postgres + Firebase emulator: MCP `initialize` → `tools/list` → `tools/call` for all 8 tools via curl.
- [x] Cross-org isolation verified on both read tools (`list_cases`, `case_assignment_stats`) and the write tool (`case_bulk_assign`) using two distinct API keys/orgs — org B cannot read or mutate org A's cases.
- [x] `case_bulk_assign` verified in both directions (assign/unassign) and cross-checked against the DB; also smoke-tested the new `unassign` action via the existing `POST /cases/mass-update` REST endpoint.
- [ ] Not yet tested against a real MCP client (Claude Desktop/claude.ai) — see plan notes on `mcp-remote` bridging for a follow-up POC session.

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/checkmarble/marble-backend/infra"
+	mcpserver "github.com/checkmarble/marble-backend/mcp"
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/pubapi"
 	pubapiv1 "github.com/checkmarble/marble-backend/pubapi/v1"
@@ -85,6 +86,12 @@ func addRoutes(r *gin.Engine, conf Configuration, uc usecases.Usecases, auth uti
 			auth.AuthedBy(utils.PublicApiKey, utils.ApiKeyAsBearerToken), uc)
 		pubapiv1.BetaRoutes(cfg, r.Group("/v1beta"),
 			auth.AuthedBy(utils.PublicApiKey, utils.ApiKeyAsBearerToken), uc)
+	}
+
+	// MCP server (POC)
+	{
+		mcpserver.Routes(uc, conf.AppVersion, r.Group("/mcp"),
+			auth.AuthedBy(utils.PublicApiKey, utils.ApiKeyAsBearerToken))
 	}
 
 	// Screening Indexer endpoints
--- a/dto/case_dto.go
+++ b/dto/case_dto.go
@@ -193,7 +193,7 @@ type CaseDecisionListPaginationDto struct {
 }
 
 type CaseMassUpdateDto struct {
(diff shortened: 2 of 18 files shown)


OrriMandarin:
I close it because should be a POC

PR #1866, opened 2026-07-09
feat: simplify Ai configuration and allow custom configuration

OrriMandarin:
## Summary

Simplifies how AI agent models are configured, and lets self-hosted clients override which model is used per feature — without needing to touch or understand every individual prompt file.

I wrote a documentation about the AI configuration: https://docs.checkmarble.com/docs/ai-configuration

## Before

`prompts/ai_agent_models.json` mapped every individual **prompt file path** (12+ entries) to a literal model name:

```json
{
  "default_model": "gemini-2.5-flash",
  "prompt_models": {
    "case_review/case_review.md": "gemini-2.5-pro",
    "rule/rule_generation_step1.md": "claude-opus-4-6"
  }
}
```

This didn't scale for clients plugging in their own provider/models: they'd have to pick a model for every prompt path individually, with no guidance on which ones actually need a stronger model, and a renamed/added prompt file silently fell back to `default_model` with no warning.

## After

The config now has two knobs: a `light` and a `heavy` model tier, set globally and optionally overridden per **feature** (not per prompt file):

```json
{
  "default_light_model": "gemini-2.5-flash",
  "default_heavy_model": "gemini-2.5-pro",
  "rule_builder": {
    "heavy_model": "claude-opus-4-6"
  }
}
```

- **Only two tiers**: `light` and `heavy`. Each usecase already knows whether a given prompt is a cheap/fast step or a step that needs a stronger model — it declares that once via a `promptSpec{Path, Feature, Tier}` value at the prompt's definition site (e.g. `usecases/ai_agent/ai_rule_assist.go`), instead of a model name being pinned centrally per file path.
- **Features, not paths**: `case_review`, `rule_description`, `rule_builder`, `screening_hit_suggestion`. Each is an optional, named field in the config (not a map), so the schema is self-documenting and there's no key-collision risk.
- **Resolution**: `AiAgentModelConfig.GetModel(feature, tier)` returns the feature's override for that tier if set, else the global default for that tier.

## Overriding the configuration

A new env var, `AI_AGENT_MODELS_CONFIG_OVERRIDE_FILE`, points to an optional local JSON file merged on top of the base `ai_agent_models.json` (itself resolved from the prompts filesystem — local dir or downloaded bundle). The merge is field-by-field via `mergo`: every field in the override file is optional and falls back to the base's value when empty/absent, down to individual tiers within a feature — so a client can override just `rule_builder.heavy_model` without restating anything else.

## Other notable changes

- **Resolved once at startup, not per request**: the merged config used to be reloaded and re-merged (opening/parsing JSON) on every single prompt preparation call. It's now loaded once in `cmd/server.go`/`cmd/worker.go` (via the new `configAiRessource` helper, shared by both) and injected into `AiAgentUsecase`, alongside `aiPromptsFS`.
- **Validation**: the resolved config must have both `default_light_model` and `default_heavy_model` set, otherwise it's treated as unusable (`nil`) rather than silently resolving to an empty model string. `AiAgentUsecase.preparePromptWithModel` guards against both a `nil` config and an empty resolved model, returning a clear error instead of calling the LLM with no model or panicking.
- **Removed** `AI_AGENT_MAIN_AGENT_DEFAULT_MODEL` / `MainAgentDefaultModel` — dead code, since prompts (and now the model config) can't be read at all when there's no prompts filesystem, so the fallback was never reachable.
- `rule_assist` split into two independent features/tiers: `rule_description` (describing an existing rule, light) and `rule_builder` (building a rule from natural language, heavy).

## Test plan

- `go build ./...`, `go vet ./...`
- `go test ./models/... ./usecases/...`
- Added `models/ai_agent_config_test.go` covering: missing base file, missing defaults (rejected), partial override of global defaults, partial override of a single feature/tier, override introducing a feature not in the base, and override supplying defaults when there's no prompts filesystem at all.

## Summary by CodeRabbit

* **New Features**
  * Added configurable light/heavy AI models per feature with tiered resolution.
  * Added optional local JSON override file to customize model selections on top of the base prompts configuration.
  * Updated server/worker startup to use the override-driven model configuration.
* **Bug Fixes**
  * Improved validation and fail-safe startup behavior when model config or overrides are missing/invalid.
  * Enforced required light/heavy defaults and clear errors for unconfigured feature/tier combinations.
* **Tests**
  * Added coverage for base loading, partial overrides, missing settings, and overrides for features not present in base config.
* **Documentation**
  * Updated the example environment configuration to replace the removed default model setting with the new override file option.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -185,8 +185,6 @@ KILL_IF_READ_LICENSE_ERROR=false
 # AI_AGENT_MAIN_AGENT_URL=
 # API key of the AI provider
 # AI_AGENT_MAIN_AGENT_KEY=
-# Default model to use for the AI agent (e.g. gemini-2.5-flash)
-# AI_AGENT_MAIN_AGENT_DEFAULT_MODEL=
 # Backend of the AI provider (for aistudio) gemini or vertex
 # AI_AGENT_MAIN_AGENT_BACKEND=
 # Project ID of the AI provider (for aistudio)
@@ -195,6 +193,10 @@ KILL_IF_READ_LICENSE_ERROR=false
 # AI_AGENT_MAIN_AGENT_LOCATION=
 # API key of the Perplexity API (for enrichment)
 # AI_AGENT_PERPLEXITY_API_KEY=
+# Path to a local JSON file overriding the light/heavy AI agent models (default_light_model,
+# default_heavy_model, and/or per-feature overrides). Merged field by field on top of the
+# resolved ai_agent_models.json, every field is optional and falls back to the base's value.
+# AI_AGENT_MODELS_CONFIG_OVERRIDE_FILE=
 
 MARBLE_API_INTERNAL_URL=https://api.internal.checkmarble.com
 # Token used to authenticate the screening indexer to access the dataset files. Use openssl to generate one.
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -2,9 +2,12 @@ package cmd
 
 import (
 	"context"
+	"io/fs"
 
+	"github.com/checkmarble/marble-backend/infra"
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/repositories"
+	"github.com/checkmarble/marble-backend/usecases/ai_agent"
 	"github.com/checkmarble/marble-backend/utils"
 	"github.com/cockroachdb/errors"
 )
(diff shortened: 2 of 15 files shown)


OrriMandarin:
@Pascal-Delange At least the new config file schema

cmd/common.go
```diff
@@ -53,3 +56,28 @@ func IsWebhookSystemMigrated(ctx context.Context, repositories repositories.Repo
...
+	var aiPromptsFS fs.FS
+	if license.LicenseValidationCode == models.VALID {
+		aiPromptsFS = infra.InitAiPromptsFS(ctx, aiPromptsServingDir, version, licenseConfig.LicenseKey)
+		if err := ai_agent.ValidatePromptsFS(aiPromptsFS); err != nil {
+			utils.LoggerFromContext(ctx).WarnContext(ctx, "ai prompts filesystem failed validation, ai features are unavaible", "error", err.Error())
+			aiPromptsFS = nil
+		}
+	}
+
+	aiAgentModelConfig, err := models.LoadAiAgentModelConfig(aiPromptsFS, aiAgentConfig.ModelsConfigOverridePath)
```
apognu, line 79: Shouldn't that also be gated behind the license check? Or even the feature access?

models/ai_agent_config.go
```diff
@@ -2,63 +2,165 @@ package models
...
+		return nil
+	}
+}
+
+// Validate checks that c has enough information to resolve a model for every feature: both
+// global defaults must be set, since any feature/tier without its own override falls back to
+// them. Without this, GetModel could silently return "" for an unconfigured feature/tier.
+func (c *AiAgentModelConfig) Validate() error {
+	if c.DefaultLightModel == "" || c.DefaultHeavyModel == "" {
+		return errors.New("ai agent model configuration is missing default_light_model and/or default_heavy_model")
```
apognu, line 70: We wanted to hard fail here? Not fall back to something?

.env.example
```diff
@@ -195,6 +193,10 @@ KILL_IF_READ_LICENSE_ERROR=false
 # AI_AGENT_MAIN_AGENT_LOCATION=
 # API key of the Perplexity API (for enrichment)
 # AI_AGENT_PERPLEXITY_API_KEY=
+# Path to a local JSON file overriding the light/heavy AI agent models (default_light_model,
+# default_heavy_model, and/or per-feature overrides). Merged field by field on top of the
+# resolved ai_agent_models.json, every field is optional and falls back to the base's value.
+# AI_AGENT_MODELS_CONFIG_OVERRIDE_FILE=
```
apognu, line 199: Is the format of this file documented somewhere?

apognu (changes requested):

models/ai_agent_config.go
OrriMandarin: It’s challenging to revert to a reliable option. Depending on your environment, certain models may be unavailable, or the model name might be invalid if the configured provider expects a name like `provider/model-name`. It’s better to ensure that the setting is correctly configured.

.env.example
OrriMandarin: => https://docs.checkmarble.com/docs/ai-configuration :D

cmd/common.go
OrriMandarin: Done

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	cloud.google.com/go/iam v1.11.0
 	cloud.google.com/go/profiler v0.6.0
 	cloud.google.com/go/storage v1.63.0
+	dario.cat/mergo v1.0.2
```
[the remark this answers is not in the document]
OrriMandarin: Done

OrriMandarin:
Documentation about how to write the AI configuration file: https://docs.checkmarble.com/docs/ai-configuration

apognu (approved):

PR #1873, opened 2026-07-15
rm convoy

Pascal-Delange:
Nothing big to say, it should just work.
No more need to call a callback to send webhooks in the usecases.
No more check to migrate webhooks on worker startup.
No more codegen dependency (!!!!).
Creates a view called "webhook_events" to replace "webhook_events_v2" in the next release. The old table is kept with "_deprec" suffix for now.

## Summary by CodeRabbit

* **New Features**
  * Webhook delivery now uses the updated delivery system, with webhook event creation and dispatch handled together.
  * Webhook secret rotation is now enabled via configuration.
  * Webhook availability is no longer tied to Convoy/webhook-migration setup.

* **Bug Fixes**
  * Improved webhook consistency by removing deferred delivery and legacy webhook retry/migration behavior.

* **Chores**
  * Removed Convoy webhook integration configuration and related generated/client code.


go.mod
```diff
@@ -54,7 +54,6 @@ require (
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826
 	github.com/oapi-codegen/oapi-codegen/v2 v2.7.1
```
apognu: We still depend on `oapi-codegen`?

apognu (approved):

go.mod
Pascal-Delange: no, good catch

PR #1875, opened 2026-07-17
avoid token overflow with lexis nexis data on screening review

Pascal-Delange:
Tentative fix for context overflows using AI reviews on cases with screenings. The issue mostly happens with Lexis data (for reasons of data volume).

### DIFF
--- a/dto/agent_dto/screening_payload.go
+++ b/dto/agent_dto/screening_payload.go
@@ -1,0 +1,128 @@
+package agent_dto
+
+import (
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


dto/agent_dto/screening_payload.go
OrriMandarin, line 1: To simplify the understanding of this method, we can provide an example of the input and the expected output, visually demonstrating what is retained and filtered.

OrriMandarin (approved):

PR #1884, opened 2026-07-22
Properly check if blob was not found in async upload.

apognu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of missing upload data during asynchronous uploads.
  * Ensured upload jobs continue to cancel after the timeout or retry after the scheduled interval as expected.

### DIFF
--- a/usecases/worker_jobs/async_upload_job.go
+++ b/usecases/worker_jobs/async_upload_job.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/google/uuid"
 	"github.com/riverqueue/river"
+	"gocloud.dev/gcerrors"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


usecases/worker_jobs/async_upload_job.go
```diff
@@ -63,7 +64,7 @@ func (w AsyncUploadWorker) Work(ctx context.Context, job *river.Job[models.Async
 	blobAttrs, err := w.blobRepository.GetBlobAttributes(ctx, w.ingestionBucketUrl, job.Args.Key)
 	if err != nil {
 		// If the blob was not uploaded yet
-		if errors.Is(err, models.NotFoundError) {
+		if gcerrors.Code(err) == gcerrors.NotFound {
```
OrriMandarin, line 67: I wonder if we should move this code into the `GetBlobAttributes` method to abstract the NotFound error using our `models.NotFoundError`. For instance, the `GetBlob()` method returns a `models.NotFoundError` if the object doesn’t exist.

OrriMandarin (approved):

usecases/worker_jobs/async_upload_job.go
apognu: Yeah, we'll have a set of changes to make over gocloud when we upgrade. They changed how their errors are represented, so let's table it for now.

PR #1898, opened 2026-07-27
fix: version outdated check ignore future version to be released

OrriMandarin:
Since we use the RC version before releasing the marble open-source package, the `outdated` feature on our staging environment will always open the outdated modal to inform us that we are in an unusual configuration (it cannot locate our RC version because it is after the latest open source released version of marble).

To prevent this issue, I modified the comparison of the backend versions with the current one, excluding the pre-release string value. This change ensures that the current version is always discoverable and prevents the `Outdated` field from being updated.

The only restriction is that the `LatestVersion` and `LatestUrl` will point to the last marble open-source version, rather than the latest backend pre-released version, but the frontend should not use these value since we return `outdated = false` in the body

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version update detection by using the backend release list as the source of truth for the latest available version.
  * Enhanced semantic version matching by normalizing prerelease tags and comparing versions semantically (not by raw tag text).
  * More reliably handles unusual tags and accurately flags future or unreleased versions.
  * Kept release-note generation behavior, while improving overall accuracy across a wider range of versions.
* **Tests**
  * Expanded coverage with additional historical and prerelease scenarios.

### DIFF
--- a/utils/outdated.go
+++ b/utils/outdated.go
@@ -62,12 +62,12 @@ func checkOutdated(ctx context.Context, appVersion string) (info OutdatedInfo) {
 	withoutPrerelease, _ := currentSv.SetPrerelease("")
 	currentSv = &withoutPrerelease
 
-	releases, err := fetchReleases(ctx, GithubBackendReleaseUrl)
+	backendReleases, err := fetchReleases(ctx, GithubBackendReleaseUrl)
 	// In case we cannot retrieve releases or the list is empty, bail out.
 	if err != nil {
 		return
 	}
-	if len(releases) == 0 {
+	if len(backendReleases) == 0 {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


utils/outdated.go
```diff
@@ -85,10 +85,14 @@ func checkOutdated(ctx context.Context, appVersion string) (info OutdatedInfo) {
 	// Keep the latest unique minor version found, so we can count them.
 	var seenMinor *semver.Version
 
-	info.ReleaseNotes = make([]string, 0, minorsSince)
```
OrriMandarin, line 88: Make the slice with `minorsSince`, but this variable is updated after this line. However, it doesn’t represent the number of release notes we can include in the array since we can include minor version release notes or all existing release notes if the current version is unknown.

apognu (approved):

PR #1899, opened 2026-07-27
feat: Save the rule description

OrriMandarin:
## Summary

Automatically generate an AI description for scenario rules and store it separately from the user-authored `Description` to prevent manual overwriting.

Add a new column in `scenario_iteration_rules` to store the AI-generated description. If a rule already has a user-written description, skip generating an AI description since the user description will override it.

When the rule formula changes compared to the previous one, clear the AI description column to force a new generation because the AI description is no longer valid.

The generation process occurs at the scenario iteration commitment, not before.

## Summary by CodeRabbit

* **New Features**
  * Rules now support AI-generated descriptions, which are available through rule data and preserved when duplicating iterations.
  * AI descriptions are generated automatically in the background for eligible rules.
  * Descriptions are refreshed when a rule’s formula changes.
* **Bug Fixes**
  * Scenario AST validation now runs against the correct organization context.
* **Tests**
  * Added coverage for identifying rules that require AI description generation and comparing rule formulas.


usecases/scenario_usecase.go
```diff
@@ -104,11 +104,11 @@ func (usecase *ScenarioUsecase) UpdateScenario(
 }
 
 func (usecase *ScenarioUsecase) ValidateScenarioAst(ctx context.Context,
-	scenarioId string, astNode *ast.Node, expectedReturnType ...string,
+	orgId uuid.UUID, scenarioId string, astNode *ast.Node, expectedReturnType ...string,
```
OrriMandarin, line 107: I modified the method signature because the method extracts the organization ID using the `usecase.enforceSecurity`. This is acceptable if we call this method via an HTTP request, but from a worker, the security doesn’t have the organization ID.

apognu: I think I am going to finish my concept of a "system token" where we can obtain, from within the code, a token that is emitted for the backend itself, that can be attached to any organization.

That would save us from having to modify those method signatures or having completely separate "unauthenticated" methods for when the worker performs an action. With this, we would be able to rely solely on the security providers to determine the org ID.

usecases/ai_agent/rule_ai_description.go
```diff
@@ -0,0 +1,48 @@
+package ai_agent
+
+import "github.com/checkmarble/marble-backend/models"
+
+// PreviousCommittedRulesByStableId returns, for the most recently committed
+// iteration among the given ones (highest non-nil Version), its rules
+// indexed by StableRuleId. Returns an empty map if none of the iterations
+// is committed yet.
+func PreviousCommittedRulesByStableId(iterations []models.ScenarioIteration) map[string]models.Rule {
```
apognu: This method isn't used anywhere except in tests, right? I can't find any other references.

If used, I could see a race condition by doing "latest committed iteration".

OrriMandarin: Oh right, initially, I intended to compare the AST formula to determine if the version required an AI description generation by fetching the previous committed version of the rule. 
However, I removed this section because I emptied the ai_description field to force a new generation when updating the formula (so there’s no need to use this method).

OrriMandarin: Removed

usecases/scenario_usecase.go
OrriMandarin: Yeeeeeeeeahhhhhhhhhhhhh

apognu (approved):

PR #1908, opened 2026-08-03
Case SLAs

jyrivallan:
This PR introduces Service Level Agreement (SLA) functionality to inboxes, allowing teams to configure response time expectations for cases.

## Features

- SLA Configuration: Inbox admins can now set an SLA duration (in days, >= 1) via a new bulk update endpoint PATCH /inboxes. The SLA value is nullable to allow disabling SLAs.
- Case Due Dates: Every case response now includes a computed due_at field, calculated as created_at + inbox.sla_days using calendar days arithmetic.
- Efficient Data Loading: Single-case fetches compute due_at with one additional inbox lookup; list endpoints batch-load all inbox SLAs in a single query to avoid N+1 problems.
- Atomic Bulk Updates: The bulk SLA update endpoint guarantees atomicity—all inboxes are updated successfully or none are updated.

## Technical Details

- Database: Added nullable sla integer column to inboxes table with CHECK constraint (sla IS NULL OR sla >= 1).
- Validation: SLA values are validated at both the DTO layer (binding tags) and usecase layer (manual checks).
- API: Internal API only (public API unchanged); due_at is omitted from responses when the case's inbox has no SLA configured.

## Summary by CodeRabbit

* **New Features**
  * Configure and update SLA targets for one or multiple inboxes.
  * Clear or update SLA settings with validation.
  * Cases now display calculated due dates based on inbox SLA settings.
* **Bug Fixes**
  * Improved inbox update validation and access checks.
  * Case due dates now reflect related inbox SLA settings.

### DIFF
--- a/dto/inbox_dto.go
+++ b/dto/inbox_dto.go
@@ -85,7 +85,7 @@ type UpdateInboxInput struct {
 	CaseReviewManual        *bool                      `json:"case_review_manual"`
 	CaseReviewOnCaseCreated *bool                      `json:"case_review_on_case_created"`
 	CaseReviewOnEscalate    *bool                      `json:"case_review_on_escalate"`
-	Sla                     pure_utils.Null[int]      `json:"sla" binding:"omitempty,min=1"`
+	Sla                     pure_utils.Null[int]       `json:"sla" binding:"omitempty,min=1"`
 }
 
 func AdaptUpdateInboxInput(i UpdateInboxInput) models.UpdateInboxInput {
--- a/dto/inbox_dto.go
+++ b/dto/inbox_dto.go
@@ -85,6 +85,7 @@ type UpdateInboxInput struct {
 	CaseReviewManual        *bool                      `json:"case_review_manual"`
 	CaseReviewOnCaseCreated *bool                      `json:"case_review_on_case_created"`
 	CaseReviewOnEscalate    *bool                      `json:"case_review_on_escalate"`
+	Sla                     pure_utils.Null[int]      `json:"sla" binding:"omitempty,min=1"`
 }
 
 func AdaptUpdateInboxInput(i UpdateInboxInput) models.UpdateInboxInput {
@@ -95,22 +96,22 @@ func AdaptUpdateInboxInput(i UpdateInboxInput) models.UpdateInboxInput {
 		CaseReviewManual:        i.CaseReviewManual,
 		CaseReviewOnCaseCreated: i.CaseReviewOnCaseCreated,
 		CaseReviewOnEscalate:    i.CaseReviewOnEscalate,
-		Sla:                     pure_utils.Null[int]{},
+		Sla:                     i.Sla,
 	}
 }
 
-type UpdateInboxSlaItem struct {
+type UpdateInboxItem struct {
 	Id  uuid.UUID `json:"id" binding:"required"`
 	Sla *int      `json:"sla" binding:"omitempty,min=1"`
 }
 
-type BulkUpdateInboxSlaInput struct {
-	Inboxes []UpdateInboxSlaItem `json:"inboxes" binding:"required"`
+type BulkUpdateInboxInput struct {
+	Inboxes []UpdateInboxItem `json:"inboxes" binding:"required"`
 }
 
-func AdaptBulkUpdateInboxSlaInput(i BulkUpdateInboxSlaInput) []models.UpdateInboxSlaItem {
-	return pure_utils.Map(i.Inboxes, func(item UpdateInboxSlaItem) models.UpdateInboxSlaItem {
-		return models.UpdateInboxSlaItem{
+func AdaptBulkUpdateInboxInput(i BulkUpdateInboxInput) []models.UpdateInboxItem {
+	return pure_utils.Map(i.Inboxes, func(item UpdateInboxItem) models.UpdateInboxItem {
+		return models.UpdateInboxItem{
 			Id:  item.Id,
 			Sla: item.Sla,
 		}
--- a/dto/inbox_dto.go
+++ b/dto/inbox_dto.go
@@ -17,6 +17,7 @@ type InboxDto struct {
 	EscalationInboxId *uuid.UUID     `json:"escalation_inbox_id"`
 	AutoAssignEnabled bool           `json:"auto_assign_enabled"`
 	Users             []InboxUserDto `json:"users"`
+	Sla               *int           `json:"sla"`
 
 	CaseReviewManual        bool `json:"case_review_manual"`
 	CaseReviewOnCaseCreated bool `json:"case_review_on_case_created"`
@@ -35,6 +36,7 @@ func AdaptInboxDto(i models.Inbox) InboxDto {
 		EscalationInboxId:       i.EscalationInboxId,
 		AutoAssignEnabled:       i.AutoAssignEnabled,
 		Users:                   pure_utils.Map(i.InboxUsers, AdaptInboxUserDto),
+		Sla:                     i.Sla,
 		CaseReviewManual:        i.CaseReviewManual,
 		CaseReviewOnCaseCreated: i.CaseReviewOnCaseCreated,
 		CaseReviewOnEscalate:    i.CaseReviewOnEscalate,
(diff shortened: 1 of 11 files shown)


dto/inbox_dto.go
```diff
@@ -93,5 +95,24 @@ func AdaptUpdateInboxInput(i UpdateInboxInput) models.UpdateInboxInput {
...
 		CaseReviewOnEscalate:    i.CaseReviewOnEscalate,
+		Sla:                     pure_utils.Null[int]{},
 	}
 }
+
+type UpdateInboxSlaItem struct {
+	Id  uuid.UUID `json:"id" binding:"required"`
+	Sla *int      `json:"sla" binding:"omitempty,min=1"`
+}
+
```
OrriMandarin: WDYT about using the existing `UpdateInboxInput` struct because the purpose is the same and you avoid creating a very specific struct

usecases/inboxes_usecase.go
```diff
@@ -218,6 +218,50 @@ func (usecase *InboxUsecase) DeleteInbox(ctx context.Context, inboxId uuid.UUID)
 	return nil
 }
 
+func (usecase *InboxUsecase) BulkUpdateInboxSla(ctx context.Context, orgId uuid.UUID, items []models.UpdateInboxSlaItem) ([]models.Inbox, error) {
```
OrriMandarin: Let’s see if we can utilize the `UpdateInbox` method since the two functions are quite similar

models/case.go
```diff
@@ -310,3 +311,13 @@ func CaseMassUpdateActionFromString(s string) CaseMassUpdateAction {
...
+
+// ComputeSlaDueAt computes the SLA due date as created_at + sla days.
+// Returns nil if sla is nil or zero.
+func ComputeSlaDueAt(createdAt time.Time, sla *int) *time.Time {
+	if sla == nil || *sla == 0 {
+		return nil
+	}
+	dueAt := createdAt.AddDate(0, 0, *sla)
+	return &dueAt
+}
```
OrriMandarin, line 323: I don’t have a strong opinion on my previous remark, but I’d like to suggest an alternative approach. Instead of fetching the inbox to retrieve the SLA value and then calling this method to populate the `dueAt` field, what are your thoughts on relying on the database to populate this field when fetching the case?

I’m particularly concerned about forgetting to call this method when manipulating a case in our use case and inadvertently sending a false representation of the case to our client.

OrriMandarin: I’ve encountered this issue multiple times. After refactoring, I often forget to call a method to populate a field, leading to incomplete data.

However, I’m aware that we don’t have a robust pattern to ensure that the data we manipulate is correctly populated and fetched.

cc: @apognu 👋

OrriMandarin (commented):
The public relations aspect is satisfactory. I made some comments about the code design, but the functionality appears to be good. :)

usecases/inboxes_usecase.go
william-schlegel: I prefer to send one array of inbox ids and settings instead of multiple updates  from the front end, but I know that it is not the case in other configurations. It's just a preference

OrriMandarin: I didn’t intend to replace what he did with the `patch` endpoint to update multiple inboxes simultaneously. Instead, I suggested using the same method to update an inbox in our usecase layer to avoid having different ways to update inbox data. In this method, we emit something at the end. It’s typically something we could forget to change if we need to modify it.

dto/inbox_dto.go
```diff
@@ -83,6 +85,7 @@ type UpdateInboxInput struct {
 	CaseReviewManual        *bool                      `json:"case_review_manual"`
 	CaseReviewOnCaseCreated *bool                      `json:"case_review_on_case_created"`
 	CaseReviewOnEscalate    *bool                      `json:"case_review_on_escalate"`
+	Sla                     pure_utils.Null[int]      `json:"sla" binding:"omitempty,min=1"`
```
OrriMandarin: The JSON tag is not aligned with the previous line; it is missing a whitespace before it.

usecases/inboxes_usecase.go
```diff
@@ -132,40 +132,49 @@ func (usecase *InboxUsecase) CreateInboxWithExecutor(
...
-		ctx,
-		usecase.transactionFactory,
-		func(tx repositories.Transaction) (models.Inbox, error) {
-			inbox, err := usecase.inboxRepository.GetInboxById(ctx, tx, inboxId)
-			if err != nil {
-				return models.Inbox{}, err
-			}
+func (usecase *InboxUsecase) updateSingleInboxWithValidation(
+	ctx context.Context,
+	tx repositories.Transaction,
```
OrriMandarin: You can use `exec repositories.Executor` instead forcing using a transaction in the method

usecases/inboxes_usecase.go
```diff
@@ -132,40 +132,49 @@ func (usecase *InboxUsecase) CreateInboxWithExecutor(
...
-func (usecase *InboxUsecase) UpdateInbox(ctx context.Context, inboxId uuid.UUID, input models.UpdateInboxInput) (models.Inbox, error) {
-	inbox, err := executor_factory.TransactionReturnValue(
-		ctx,
-		usecase.transactionFactory,
-		func(tx repositories.Transaction) (models.Inbox, error) {
-			inbox, err := usecase.inboxRepository.GetInboxById(ctx, tx, inboxId)
-			if err != nil {
-				return models.Inbox{}, err
-			}
+func (usecase *InboxUsecase) updateSingleInboxWithValidation(
```
OrriMandarin: Why `WithValidation` name?

OrriMandarin (approved):

dto/inbox_dto.go
jyrivallan: You’re right, this is ugly

Pascal-Delange: Regardless of ugly, I'm surprised this isn't something that gofmt handles automatically 🤔

repositories/migrations/20260803134500_add_inbox_sla.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+alter table inboxes
+    add column sla integer,
+    add constraint inboxes_sla_check check (sla is null or sla >= 1);
```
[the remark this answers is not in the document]
Pascal-Delange: not relevant for this table

usecases/inboxes_usecase.go
```diff
@@ -132,40 +132,49 @@ func (usecase *InboxUsecase) CreateInboxWithExecutor(
...
+		if err := usecase.enforceSecurity.ReadInbox(escalationInbox); err != nil {
+			return models.Inbox{}, err
+		}
+	}
 
-			if err := usecase.inboxRepository.UpdateInbox(ctx, tx, inboxId, input); err != nil {
-				return models.Inbox{}, err
-			}
+	if err := usecase.inboxRepository.UpdateInbox(ctx, tx, inboxId, input); err != nil {
+		return models.Inbox{}, err
```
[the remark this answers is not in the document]
Pascal-Delange: Well, that sounds like low criticity

usecases/inboxes_usecase.go
Pascal-Delange: Right, after which you don't need to wrap in a transaction further below.

Pascal-Delange (approved):
Nice !

usecases/inboxes_usecase.go
```diff
@@ -132,41 +132,45 @@ func (usecase *InboxUsecase) CreateInboxWithExecutor(
...
-				escalationInbox, err := usecase.inboxRepository.GetInboxById(ctx, tx, input.EscalationInboxId.Value())
-				if err != nil {
-					return models.Inbox{}, err
-				}
-				if err := usecase.enforceSecurity.ReadInbox(escalationInbox); err != nil {
-					return models.Inbox{}, err
-				}
-			}
+	if input.EscalationInboxId.Set && input.EscalationInboxId.Valid {
+		escalationInbox, err := usecase.inboxRepository.GetInboxById(ctx, exec, input.EscalationInboxId.Value())
```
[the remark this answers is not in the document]
OrriMandarin: We already test if the inbox is from this organization or not, see the usecase.enforceSecurity.ReadInbox and usecase.enforceSecurity.UpdateInbox we use in this method

PR #1915, opened 2026-08-04
fix: prompt version selection, bad comparaison

OrriMandarin:
Since we deploy the `rc` version before the official release, all versions with a prerelease part or "suffix" are considered before those without.

For example, `v1.7.0` is considered before `v1.7.0-xxxxx`.

Therefore, in production, the prompt folder is not correctly fetched and used, resulting in the use of an outdated version that may be incompatible depending on the changes made.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prompt version selection for prerelease requests by matching them with the corresponding stable release when available.
  * Added fallback handling for prerelease versions when only earlier releases exist.
  * Correctly returns no matching version when a prerelease is older than all available candidates.


jyrivallan (approved):
LGTM

PR #1916, opened 2026-08-04
fix: process large csv file

OrriMandarin:
## Problem

A CSV upload is handled by a single `csv_ingestion` job with a 1-hour timeout. Large files —
the presigned-link path allows up to 10 GB — don't finish in time: river cancels the job, and
the next attempt returns immediately because the upload log is no longer `pending`. The upload
stays stuck in `processing` and the whole file has to be re-uploaded.

## Approach

Make the job resumable, the same way `apply_delta_file` does in continuous screening: persist a
byte offset after each batch, and when the attempt is about to run out of time, save the
checkpoint and snooze instead of dying.

## What changed

- **Migration** — `upload_logs.byte_offset bigint`: offset of the first row not yet ingested.
  `bigint` and not `integer`, since files go up to 10 GB.
- **Checkpoint** — `SaveUploadLogCheckpoint` writes `byte_offset` + `num_rows_ingested` in a
  single UPDATE, once per batch of `CSV_INGESTION_BATCH_SIZE` rows.
- **Resume** — `processUploadLog` now accepts a log already in `processing` and reopens the blob
  at its stored offset through `WithBeginOffset`.
- **Snooze** — the loop compares against a deadline derived from `ctx.Deadline()` minus a
  2-minute margin. Past it, it checkpoints and returns `Incomplete`, which the worker turns into
  `river.JobSnooze`. `CsvIngestionOutcome` exists so the usecase doesn't return river sentinels.
- **Cap** — after `CSV_INGESTION_MAX_SNOOZES` (24) resumes the upload is failed, with the offset
  and row count in the error message.

## Where to look carefully

1. **The header is read on its own, every attempt** (`readCsvHeader`). The data reader is then
   always opened at an explicit offset, so it never sees the header row or the BOM — that's what
   makes an arbitrary resume point possible.
2. **The BOM is discarded before `csv.Reader` sees it, and added back into the offset.** Trimming
   it off the parsed header name doesn't work: with a quoted header, `csv.Reader` reads the BOM as
   the start of an unquoted field and rejects the file. And since discarding it hides 3 bytes from
   the reader, `bomLen` has to be added back or every persisted offset is 3 bytes short and a
   resume lands mid-field.
3. **A cancelled context is not a failure.** `failAttempt` skips `setToFailed` when
   `ctx.Err() != nil` (worker shutdown, river timeout). `setToFailed` uses `context.WithoutCancel`,
   so without that guard it would happily write a terminal `failure` and make the checkpoint
   unreachable.

## Deliberate choices

- **The last batch is at-least-once.** `upload_logs` lives in the marble DB and client objects in
  the org DB, so the checkpoint can't share a transaction with the insert. It's ordered
  checkpoint-after-insert, so a crash in that window re-ingests up to one batch instead of
  skipping rows. Cost: duplicate object versions and re-enqueued screening/scoring for that batch.
- **No lease on the upload log.** River runs one attempt at a time; the checkpoint UPDATE also
  carries a `byte_offset <= new_offset` guard so a stale attempt can't rewind it.
- **After a resume, error line numbers are relative to the current pass.** `num_rows_ingested`
  counts inserted objects (deduped by `object_id`), not rows read, so the absolute line number
  isn't knowable. The byte offset is reported next to it.

## Fixed along the way

An upload whose worker was killed mid-file used to stay in `processing` forever, because the
retry early-returned on the status check. It now resumes.

## Summary by CodeRabbit

- **New Features**
  - CSV ingestion can now resume from saved progress after interruptions or job time limits.
  - Progress is checkpointed after each completed batch, preserving processed rows and file position.
  - Improved handling for cancellations, deadlines, BOMs, quoted fields, and incomplete files.
  - Upload processing now reports whether ingestion completed or needs to continue later.
- **Bug Fixes**
  - Prevented progress checkpoints from moving backward.
  - Improved failure handling without losing previously saved progress.
- **Tests**
  - Added comprehensive coverage for CSV headers, resume behavior, end-of-file handling, and deadlines.

### DIFF
--- a/mocks/upload_log_repository.go
+++ b/mocks/upload_log_repository.go
@@ -24,7 +24,14 @@ func (r *UploadLogRepository) UpdateUploadLogStatus(ctx context.Context, exec re
 	return true, args.Error(0)
 }
 
-func (r *UploadLogRepository) UploadLogById(ctx context.Context, exec repositories.Executor, id string) (models.UploadLog, error) {
+func (r *UploadLogRepository) SaveUploadLogCheckpoint(ctx context.Context, exec repositories.Executor,
+	id uuid.UUID, byteOffset int64, rowsIngested int,
+) error {
+	args := r.Called(ctx, exec, id, byteOffset, rowsIngested)
+	return args.Error(0)
+}
+
+func (r *UploadLogRepository) UploadLogById(ctx context.Context, exec repositories.Executor, id uuid.UUID) (models.UploadLog, error) {
 	args := r.Called(ctx, exec, id)
 	return args.Get(0).(models.UploadLog), args.Error(1)
 }
--- a/mocks/upload_log_repository.go
+++ b/mocks/upload_log_repository.go
@@ -36,6 +36,18 @@ func (r *UploadLogRepository) UploadLogById(ctx context.Context, exec repositori
 	return args.Get(0).(models.UploadLog), args.Error(1)
 }
 
+func (r *UploadLogRepository) ListUploadLogs(
+	ctx context.Context,
+	exec repositories.Executor,
+	organizationId uuid.UUID,
+	tableName string,
+	filters models.UploadLogFilters,
+	pagination models.PaginationAndSorting,
+) ([]models.UploadLog, error) {
+	args := r.Called(ctx, exec, organizationId, tableName, filters, pagination)
+	return args.Get(0).([]models.UploadLog), args.Error(1)
+}
+
 func (r *UploadLogRepository) AllUploadLogsByTable(
 	ctx context.Context,
 	exec repositories.Executor,
--- a/pure_utils/clean_bom.go
+++ b/pure_utils/clean_bom.go
@@ -12,14 +12,23 @@ const (
 )
 
 func NewReaderWithoutBom(r io.Reader) io.Reader {
+	reader, _ := TrimBom(r)
+	return reader
+}
+
+// TrimBom behaves like NewReaderWithoutBom and additionally reports how many bytes it discarded
+// (0 or 3). Callers that track absolute byte offsets into the file need that count: the returned
+// reader starts after the BOM, so offsets measured on it are short by exactly this many bytes.
+func TrimBom(r io.Reader) (io.Reader, int64) {
 	buf := bufio.NewReader(r)
 	b, err := buf.Peek(3)
 	if err != nil {
 		// not enough bytes
-		return buf
+		return buf, 0
 	}
 	if b[0] == bom0 && b[1] == bom1 && b[2] == bom2 {
 		_, _ = buf.Discard(3)
+		return buf, 3
 	}
-	return buf
+	return buf, 0
 }
--- a/usecases/ingestion_usecase.go
+++ b/usecases/ingestion_usecase.go
@@ -38,9 +38,11 @@ const (
 	CSV_INGESTION_ITERATION_TIMEOUT = 10 * time.Second
 
 	// CSV_INGESTION_TIMEOUT_MARGIN is the headroom reserved before river's own Timeout, so there is
-	// time to save the checkpoint and snooze before river cancels the job. It must exceed the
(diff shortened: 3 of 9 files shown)


mocks/upload_log_repository.go
```diff
@@ -24,11 +24,30 @@ func (r *UploadLogRepository) UpdateUploadLogStatus(ctx context.Context, exec re
...
 
-func (r *UploadLogRepository) UploadLogById(ctx context.Context, exec repositories.Executor, id string) (models.UploadLog, error) {
+func (r *UploadLogRepository) SaveUploadLogCheckpoint(ctx context.Context, exec repositories.Executor,
+	id uuid.UUID, byteOffset int64, rowsIngested int,
+) error {
+	args := r.Called(ctx, exec, id, byteOffset, rowsIngested)
+	return args.Error(0)
+}
+
+func (r *UploadLogRepository) UploadLogById(ctx context.Context, exec repositories.Executor, id uuid.UUID) (models.UploadLog, error) {
```
jyrivallan, line 34: I can see you’re keeping this mock definition in line with its actual counterpart (including `ListUploadLogs` that was absent), but I can’t find where this mock is used 🤔

jyrivallan (approved):
LGTM

I would’ve been fooled into using [river resumable jobs](https://riverqueue.com/docs/resumable-jobs#durable-checkpoints-with-resumablesetsteptx). But this is clearly preferable.
I have just one comment about a possibly unused mock.

mocks/upload_log_repository.go
Pascal-Delange: I didn't check specifically, but most likely if it's not included some test does not compile because the class passed to a constructor does not match its expected interface (even if it's not really used in tests)

pure_utils/clean_bom.go
```diff
@@ -12,14 +12,23 @@ const (
...
 
 func NewReaderWithoutBom(r io.Reader) io.Reader {
+	reader, _ := TrimBom(r)
+	return reader
+}
+
+// TrimBom behaves like NewReaderWithoutBom and additionally reports how many bytes it discarded
+// (0 or 3). Callers that track absolute byte offsets into the file need that count: the returned
+// reader starts after the BOM, so offsets measured on it are short by exactly this many bytes.
+func TrimBom(r io.Reader) (io.Reader, int64) {
```
Pascal-Delange, line 22: Hmm wait, why do we need this?

usecases/ingestion_usecase.go
```diff
@@ -596,24 +696,77 @@ func (usecase *IngestionUseCase) processUploadLog(ctx context.Context, uploadLog
...
+// absolute byte offset at which the first data row begins.
+//
+// Excel and other Windows tools prefix UTF-8 CSVs with a BOM. It has to be discarded before parsing
+// rather than trimmed off the parsed header name, because csv.Reader treats it as part of the first
+// field: on a quoted header it turns `\ufeff"object_id"` into an unquoted field containing a bare quote
+// and fails the whole file. Only presigned uploads can carry one, since the synchronous path strips
+// it on upload. Discarding it hides those bytes from csv.Reader, so bomLen is added back to keep the
+// returned offset absolute \u2014 without it every persisted offset would be 3 bytes short and a resume
+// would restart mid-field.
+func (usecase *IngestionUseCase) readCsvHeader(ctx context.Context, fileName string) ([]string, int64, error) {
```
Pascal-Delange, line 714: Hmm, ok I see

Pascal-Delange (approved):
Damn, that sure makes it more complicated 😅 
Largely inspired from changes to batch scenario execution I assume?

One thing: this contains a migration, so it's not a good candidate for a rc hotfix (or we'd need to be _very_ careful with migration ordering, I've been bit by this before).

OrriMandarin:
@Pascal-Delange This time, there is no migration between the last `rc` and this change, so we can safely merge this PR and create a hotfix without breaking the migration

PR #1918, opened 2026-08-06
change CS intervals

OrriMandarin:
Reduce the default interval of our CS worker job

## Summary by CodeRabbit

* **Improvements**
  * Reduced the default dataset update scan interval from 24 hours to 4 hours.
  * Reduced the default full dataset generation interval from 24 hours to 8 hours.
  * Existing environment settings remain available for customizing both intervals.


Pascal-Delange (approved):
Thanks 👍 
A useful follow-up: 
- we should add a doc page, adjacent to where we explain how to configure cont screening, on how customers should have a regular cleanup of those files (or they'll just accumulate. Mostly the ones we generate)
- and then, we should try and do the same internally.
Limitation: ideally we'd want to keep at least one full file at all times for every org... which is not really something we can do with a lifecycle policy. So we're in a bit of a bind. Or we need to script to do the cleanup in code. To discuss (please create an issue ?)

Pascal-Delange:
And, in light of this, perhaps 8h or 12h for the delta files we generate, not 4h (delta files from provider every 4h is good)

PR #1919, opened 2026-08-07
fix redis delete prefix when redis is not configured

OrriMandarin:
This method didn't check if redis is configured


ChibiBlasphem (approved):
lgtm

PR #1921, opened 2026-08-10
feat: add case_sla_status_by_date case analytics

jyrivallan:
## Add SLA status time-series to case analytics

Adds a new case analytics query, `case_sla_status_by_date`, exposed via the
existing `POST /analytics/cases/:query` endpoint.

For each day in the requested range (bucketed by case creation date, in the
caller's timezone), it returns the count of cases in three buckets relative
to their inbox's configured SLA (`inboxes.sla`, in calendar days):

- `completed_within_sla`: closed before the SLA due date
- `sla_breached`: closed after the due date, or still open past the due date
- `still_open_within_sla`: still open and not yet past the due date

Cases whose inbox has no SLA configured never breach (fall into
`completed_within_sla`/`still_open_within_sla` depending on status).

Respects the existing filters (date range, timezone, inbox, assigned user)
and reuses the existing Redis time-series cache.

## Summary by CodeRabbit

* **New Features**
  * Added daily case SLA analytics.
  * Reports cases completed within SLA, SLA-breached cases, and cases still open within SLA.
  * Supports filtering by date range, inbox, organization, and assignee.
  * Available through the case analytics query interface.

### DIFF
--- a/repositories/case_analytics_repository.go
+++ b/repositories/case_analytics_repository.go
@@ -321,6 +321,70 @@ func (repo MarbleDbRepository) CaseStatusByDate(
 	})
 }
 
+func (repo MarbleDbRepository) CaseSlaStatusByDate(
+	ctx context.Context,
+	exec Executor,
+	filters analytics.CaseAnalyticsFilter,
+) ([]analytics.CaseSlaStatusByDate, error) {
+	// Same "closed_at" derivation as CasesDurationByTimeStats.
+	closedAtSubq := squirrel.
+		Select("ce.case_id", "max(ce.created_at) as closed_at").
+		From(dbmodels.TABLE_CASE_EVENTS + " ce").
+		Where(squirrel.Eq{
+			"ce.event_type": "status_updated",
+			"ce.new_value":  "closed",
+		}).
+		GroupBy("ce.case_id")
+
+	closedAtSQL, closedAtArgs, err := closedAtSubq.ToSql()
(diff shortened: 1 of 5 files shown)


repositories/case_analytics_repository.go
```diff
@@ -321,6 +321,70 @@ func (repo MarbleDbRepository) CaseStatusByDate(
...
+			) as completed_within_sla`, dueAt),
+			fmt.Sprintf(`count(*) filter (
+				where i.sla is not null and (
+					(c.status = 'closed' and ce_agg.closed_at > %s) or
+					(c.status != 'closed' and now() > %s)
+				)
+			) as sla_breached`, dueAt, dueAt),
+			fmt.Sprintf(`count(*) filter (
+				where c.status != 'closed' and (i.sla is null or now() <= %s)
+			) as still_open_within_sla`, dueAt),
```
apognu, line 361: Can't those be expressed as `squirrel.Expr` so we don't rely on string interpolation? Squirrel can be finicky with placeholders for nested queries, so maybe not, but let's make sure.

jyrivallan: Apparently `SelectBuilder.Select(columns ...string)` only accepts strings.
Besides, these string interpolations wouldn’t end up being bind params. `squirrel.Expr("...")` would be called with zero args anyway.

apognu (approved):

PR #1926, opened 2026-08-11
Graph exploration

apognu:
This adds the ability to explore the network of connections around a party, account, or transaction — across all of an organization's ingested data — without manually cross-referencing tables by hand.

- **Same-field relations (new configuration).** Organizations can now declare that two fields on two different tables represent the same real-world value, even when there's no formal link between them — for example, two accounts sharing an IBAN, or two logins sharing an IP address. This is pure configuration; it doesn't touch or move any data, it just tells the system which shared values are meaningful.

- **Adjacency index (background worker).** A periodic per-organization job rebuilds an internal index of which records share which values, combining the data model's existing relationships with the newly configured same-field relations. The index is rebuilt in the background and swapped in atomically, so live queries are never blocked or slowed down during a rebuild.

- **Graph walk.** Starting from a chosen record, the system explores outward for a configurable number of degrees (2 by default, up to 5, a degree usually covers multiple hops), following both existing relationships and same-field matches in either direction. Instead of dumping every intermediate record it passes through, it collapses the result down to the meaningful end points (parties, by default) and explains *why* they're connected — e.g. "these two parties share the same IBAN" or "this party's account sent money to that one." If a shared value turns out to be extremely common (e.g. a generic identifier shared by thousands of records), it's reported as a capped, approximate count instead of being expanded in full, to keep results fast and readable.

## Summary by CodeRabbit

- **New Features**
  - Added graph exploration APIs for traversing related records and viewing nodes, connections, risk levels, tags, and captions.
  - Added graph relation management, including listing, creating, and deleting relationships between data fields.
  - Added configurable graph-building schedules to keep exploration data current.
  - Added graph exploration licensing and organization-level feature access controls.
  - Added semantic caption fields to improve record labeling.
- **Bug Fixes**
  - Improved validation of graph endpoints, traversal options, relation groups, and malformed inputs.


repositories/migrations/20260827143800_graph_walking.sql
```diff
@@ -0,0 +1,35 @@
+-- +goose Up
+
+alter table licenses
+    add column if not exists graph_exploration boolean default false not null;
+
+alter table organization_feature_access
+    add column if not exists graph_exploration text default 'allowed' not null;
+
+create table graph_relations (
+    id uuid primary key default gen_random_uuid(),
```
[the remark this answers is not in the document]
apognu: It's already the case, the repository uses `pure_utils.NewId()` to generate UUIDv7s.

usecases/graph_relation_usecase.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+
+	if relation.OrgId != uc.enforceSecurity.OrgId() {
+		return errors.Wrapf(models.ForbiddenError, "graph relation %s does not belong to this organization", relationId)
+	}
+
+	return uc.graphRelationRepository.DeleteGraphRelation(ctx, exec, relationId)
+}
```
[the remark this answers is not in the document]
apognu: No, that's explicitly what we **do not** want in this PR. It is expected the graph will temporarily be stale if relations are changed until the full rebuild. It is a potentially very expensive operation that we do not want to repeat too often.

repositories/graph_builder_repository.go
```diff
@@ -0,0 +1,210 @@
...
+		// A table this young has no statistics at all, so without this the estimate would be the
+		// planner's no-stats default and every hypernode count the walk reports would be
+		// meaningless. Analyzing here rather than after the swap means the live table is never
+		// briefly without statistics.
+		fmt.Sprintf("analyze %s", buildTable),
+	}
+
+	for _, sql := range statements {
+		if _, err := exec.Exec(ctx, sql); err != nil {
+			return errors.Wrap(err, "error while indexing the graph build table")
```
Pascal-Delange, line 148: FWIW, log which statement failed?

usecases/graph_walk_usecase.go
```diff
@@ -0,0 +1,1700 @@
...
+// new turns up, so a record reaches its parent, its parent's parent and so on within a single
+// degree:
+//
+//	seed --(link)--> parent --(link)--> grandparent --(link)--> ...
+//
+// Running this to convergence is safe in a way followDownward is not: an upward link resolves
+// to exactly one record in a well-formed data model (that is what "many-to-one" means), so
+// walking all the way up from a handful of seeds only ever visits a handful of ancestors.
+// maxUpwardFanout and maxUpwardDepth exist only to bound the damage on a data model where that
+// assumption turns out to be false — a parent field that is not actually unique, or a cycle.
```
Pascal-Delange, line 714: only the latter should really be possible (and even then, not supported by today's data model creation UI)

usecases/graph_walk_usecase.go
```diff
@@ -0,0 +1,1700 @@
...
+
+	count, err := w.repo.EstimateValueCount(w.ctx, w.exec, recordType, fieldName, value)
+	w.queries++
+
+	if err != nil {
+		return 0, err
+	}
+
+	// The planner works off sampled statistics and routinely lands under the truth — it will
+	// happily estimate 21 rows for a value 48 records carry. Reporting a number below the
```
Pascal-Delange, line 1077: this english 🤯

Pascal-Delange (approved):
Ok, no major questions from reading the code.
Though I think I'll try to spend a bit more time understanding all of the SQL tonight, if time allows.
Let's see how long this takes to run on different sizes of client DB.

PR #1927, opened 2026-08-13
Omit empty strings from being elligible pivot values.

apognu:
When making a decision, providing an empty string for a column used in a Belongs to relationship makes it elligible to become the decision's pivot value, which obviously in 99.999…% of the cases is not what is expected.

It is especially egregious when using CSV to create batch decisions, since no columns can be omitted, and null cannot be expressed.

Now, values that are empty or made entirely of whitespace will be excluded from becoming pivot values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pivot values containing only whitespace are now treated as undefined and excluded from eligible pivot selection.
  * Non-empty pivot values continue to be selected as expected.


jyrivallan (approved):
LGTM

PR #1928, opened 2026-08-13
SLA cases analytics return only stats cases in SLA-enabled inboxes

jyrivallan:
## Fix: exclude inboxes without SLA from SLA analytics stats

The `case_sla_status_by_date` analytics endpoint was mixing stats from
inboxes with and without SLA configured, which was confusing for users.
Cases from no-SLA inboxes were silently counted as "completed within SLA" /
"still open within SLA", inflating those buckets.

Now the query filters on `i.sla IS NOT NULL`, so only inboxes with SLA
configured contribute to the stats.

### Changes

- `repositories/case_analytics_repository.go`: extracted query building into
  `buildCaseSlaStatusByDateQuery`, added `WHERE i.sla IS NOT NULL`, simplified
  the 3 aggregate filters (removed now-redundant `i.sla is null` fallbacks)
- `repositories/case_analytics_repository_test.go`: new unit tests covering
  the generated SQL (no DB required)

## Summary by CodeRabbit

* **Bug Fixes**
  * SLA status reporting now excludes inboxes without a configured SLA, improving the accuracy of case analytics.
  * Organization, inbox, date-range, aggregation, and assigned-user filters continue to work as expected.

* **Tests**
  * Added coverage for SLA filtering and query criteria to help ensure reliable analytics results.

### DIFF
--- a/repositories/case_analytics_repository.go
+++ b/repositories/case_analytics_repository.go
@@ -7,6 +7,7 @@ import (
 	"github.com/Masterminds/squirrel"
 	"github.com/checkmarble/marble-backend/models/analytics"
 	"github.com/checkmarble/marble-backend/repositories/dbmodels"
+	"github.com/cockroachdb/errors"
 	"github.com/jackc/pgx/v5"
 )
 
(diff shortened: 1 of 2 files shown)


apognu (approved):
Sounds good to me. Two notes though:

 - Just so you know, we have a stub executor, `executor_factory.ExecutorFactoryStub` which can be used in place of regular executors, and that records performed queries and mock their responses. Not really important here, but could be helpful in more complex cases.
 - What is the performance situation of that query? Does it require a specific index, how would it perform on larger datasets? We might only see that when deployed, but to keep in mind.

PR #1929, opened 2026-08-13
Adapt license key generator to produce a dashless UUIDv4.

apognu:
This method was unused until now, but some tools expect the license key to be 36-byte long max.

## Summary by CodeRabbit

* **Enhancements**
  * License keys are now generated in a standardized UUID-based format.
  * Hyphens are removed from generated license keys for a compact value.

### DIFF
--- a/usecases/license_usecase.go
+++ b/usecases/license_usecase.go
@@ -2,9 +2,7 @@ package usecases
 
 import (
 	"context"
-	"crypto/rand"
-	"encoding/hex"
-	"fmt"
+	"strings"
 	"time"
 
 	"github.com/checkmarble/marble-backend/models"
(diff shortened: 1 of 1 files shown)


jyrivallan (approved):
This is fine by me as-is.

If I may suggest another format, I think [Crockford’s Base 32](https://www.crockford.com/base32.html) would be a good fit here.
It’d be more compact, and is intended to avoid ambiguity.

apognu:
> This is fine by me as-is.
> 
> If I may suggest another format, I think https://en.wikipedia.org/wiki/Base32#Crockford's_Base32 would be a good fit here. It’d be more compact, and is intended to avoid ambiguity.

I don't think this is necessary, Crockford is useful for IDs that people might give each other verbally. This is purely a random string that is meant to be copied/pasted once (and hexadecimal already has no I, L, O or U).

And for compactness, it would be exactly the same number of characters (32).

PR #1930, opened 2026-08-14
Option to avoid re-creating decisions on already-scored objects

jyrivallan:
### Problem

Transaction monitoring usually requires each object to be scored **exactly once** per
scenario. In batch mode there is no good way to get that today: customers use
overlapping windows (26h for a daily scenario) to avoid missing objects, which
guarantees duplicate alerts — and can still miss objects if the batch starts late.

This PR adds a `deduplicate_batch_objects` setting on the scenario that enforces, **for
batch executions only**, at most one decision per object per scenario.

### Design decisions

- **Dedup key: `(scenario_id, object_id)`** — matches the ticket wording. Publishing a
  new iteration does not re-score history, so there is no duplicate burst on every publish.
- **Setting on `scenarios`** — consistent with a scenario-level key, no draft/publish cycle
  to toggle it, already available as `scenario` in both worker paths.
- **v2 coordinator only** (`BATCH_EXECUTION_V2`). A warning is logged when the setting is on
  without the flag, so the toggle is never silently inert.
- **No Redis.** It is optional in deployments (`NewRedisClient` returns `nil` without
  `REDIS_HOST`) so it cannot hold the guarantee, and the Postgres pre-filter measures
  0.08 ms per 50 ids as an index-only scan — the cache would save nothing. Add later if
  profiling says otherwise.

### Why a dedicated table and not an index on `decisions`

`decisions` has no `object_id` column (only `trigger_object jsonb`) and, more decisively,
**already contains duplicates** — creating them via overlapping windows is exactly the
workaround this PR replaces, so a unique index on existing data would fail to build.

### Mechanism

New table `scenario_scored_objects (scenario_id, object_id, created_at)` with a unique
index on `(scenario_id, object_id)`. Per 50-object manifest batch:

1. **Pre-filter** (`FilterAlreadyScoredObjects`, outside any transaction) — drops
   already-scored objects **before evaluation**, where the cost is. Safe without locking
   because the table is append-only: a committed claim is permanent, so this can only
   produce false negatives (one wasted evaluation the claim then rejects), never false
   positives.
2. **Claim** (`ClaimScoredObjects`, **inside** the batch transaction, before the writes) —
   `INSERT ... ON CONFLICT (scenario_id, object_id) DO NOTHING RETURNING object_id`. This is
   not a duplicate-tolerant upsert: on conflict with an in-progress transaction Postgres
   **waits** for its outcome before deciding, and `RETURNING` on `DO NOTHING` yields only
   rows actually inserted — so the returned set *is* the arbitration result. Ids are sorted
   so two concurrent runs take the conflicting keys in the same order and cannot deadlock.

Every evaluated object is claimed, trigger passed or not — the goal is also to stop
re-evaluating trigger conditions. Objects not found in the table are not claimed, so they
stay eligible if ingested later.

## Per-run override

The dedup setting has two layers now:
- `scenarios.deduplicate_batch_objects` — persistent default, used by scheduled (cron) runs.
- `scheduled_executions.deduplicate_objects` — the *effective* value, resolved once at
  creation and snapshotted on the row. `POST /scenario-iterations/:id/schedule-execution`
  accepts an optional `{"deduplicate_objects": bool}` body to override the default for a
  single manual run; an absent/empty body keeps prior behavior (backward compatible).

This isn't just the requested per-run option — snapshotting also fixes a real bug: the v2
coordinator re-enters `Run()` on every River slice (`JobSnooze` every ~9min), and a
full-table manifest spans several slices. Reading the scenario's setting live on each slice
meant a mid-run toggle could flip behavior partway through a single execution (stop
consulting *and* stop writing claims on the remainder → duplicate decisions on that portion,
and those objects left unclaimed for the next run). The value is now fixed at creation time.

The manual override is also the escape hatch for the limitation below — **removed**:
~~A faulty trigger condition can exclude objects permanently. Support recovery: `delete
from scenario_scored_objects where scenario_id = '…' and created_at >= '…'`~~
→ now: re-run manually with `{"deduplicate_objects": false}`.

## Unconditional recording

`scenario_scored_objects` is now populated for **every v2 batch execution**, regardless of
whether that specific run enforces dedup (`scheduled_executions.deduplicate_objects`).
Only *enforcement* — pre-filtering before evaluation and gating decision storage on the
claim outcome — stays conditional on that flag.

This makes enabling dedup on an existing scenario **immediately effective**: any object
already processed by v2 since this change shipped is excluded from the very first
enforcing run, no separate backfill pass required. The insert is idempotent
(`ON CONFLICT DO NOTHING`), so recording an object already known from a prior run (dedup
on or off) is a harmless no-op — and disabling enforcement never silently suppresses a
decision because of a claim recorded during a dedup-enabled run in the past, since the
claimed set is only consulted when the run enforces dedup.

Trade-offs accepted knowingly:
- Storage growth for `scenario_scored_objects` is no longer bounded by dedup adoption, but
  by every object ever evaluated by any v2 scenario. No purge exists yet, same as `decisions`.
- A failure recording an object now fails the whole batch even for scenarios that never use
  dedup (same fatal-error handling as before, just no longer gated) — accepted because the
  write is cheap and reliable (same transaction, same database), and best-effort handling
  would let the table silently drift out of sync with what was actually evaluated.
- Only retroactive back to whenever v2 started processing an object under this code change;
  objects only ever touched by v1, or by v2 before this patch, are not covered.

### Unique index rather than isolation level

The ticket floated "unique index or transaction isolation level". The index won:

- Under Repeatable Read / Serializable, `ON CONFLICT DO NOTHING` stops skipping silently and
  raises `40001`, which would cost us `RETURNING` as the arbitration mechanism.
- The existing retry is a single attempt (`db_executor_getter.go`) — not enough under
  contention, and each retry would replay the whole batch write.
- The batch transaction reads the data model, scenario, pivots and analytics settings and
  writes 6 tables; under SSI it would produce false positives well beyond the claim itself.
- `Transaction()` is a single shared entry point with no `TxOptions`.

Above all, the unique index is **unconditional**: it holds against any writer, at any
isolation level, including manual SQL.

Side effect worth noting: the claim **closes a pre-existing v2 duplication window**. On an
ambiguous commit (commit sent, connection lost before the ack) the retry previously
re-stored the whole batch → duplicate decisions and duplicate `DecisionWorkflowArgs` jobs,
hence duplicate alerts. The retry's claim now returns the empty set.

### Performance

Measured on a 1,575,203-row table: pre-filter is an **Index Only Scan, 0.08 ms, 3 buffers**
for 50 ids, so its cost is entirely network round-trip, not execution. Measured evaluation
throughput is **0.99 ms/object**, giving a **~49:1** ratio (~1 ms of pre-filter for ~49 ms
of evaluation saved). Net effect: run #1 is ~4% slower, repeat runs are **~10–25× faster**.

Storage: ~134 B/row, so ~210 MB per scenario at 1.57M objects. No purge, same as `decisions`.

### Progress reporting

`manifest_rows_processed` is added to the `GET /scheduled-executions` DTO.
`number_of_evaluated_decisions` can stay frozen at 0 for a whole deduplicated run — objects
are dropped before evaluation — making an active run indistinguishable from a stuck one.
`manifest_rows_processed` advances every batch regardless of dedup.

## New Features

- Added optional batch-object deduplication for scenarios.
- Prevents previously scored objects from being evaluated again in later batch runs.
- Scheduled executions inherit the scenario setting by default.
- Individual scheduled runs can override deduplication when creating an execution.
- Scheduled execution details now include deduplication status and processed manifest row counts.

## Bug Fixes

- Added a warning when deduplication is requested but unavailable for the execution mode.
- Invalid scheduled-execution request bodies now return a 400 error.


jyrivallan:
@coderabbitai review

dto/scenarios.go
```diff
@@ -9,26 +9,28 @@ import (
...
-	Archived          bool      `json:"archived"`
+	Id                      string    `json:"id"`
+	CreatedAt               time.Time `json:"created_at"`
+	Description             string    `json:"description"`
+	LiveVersionID           *string   `json:"live_version_id,omitempty"`
+	Name                    string    `json:"name"`
+	OrganizationId          uuid.UUID `json:"organization_id"`
+	TriggerObjectType       string    `json:"trigger_object_type"`
+	Archived                bool      `json:"archived"`
+	DeduplicateBatchObjects bool      `json:"deduplicate_batch_objects"`
```
apognu, line 20: Didn't we determine that this would be an option on each batch run instead of global on the scenario? Cascades to all files where this is used.

usecases/worker_jobs/batch_execution_coordinator.go
```diff
@@ -315,13 +356,50 @@ func (c *BatchExecutionCoordinator) Run(ctx context.Context, scheduledExecutionI
...
 				}
+				if inv.deduplicate {
+					if _, ok := claimed[r.objectId]; !ok {
+						// Lost the claim to a concurrent run scoring the same object. Rare
+						// (would require two runs of the same scenario overlapping on the
+						// same object within one batch window); not stored, not counted,
+						// no phantom-decision callback since the outcome is not this run's.
+						continue
+					}
+				}
```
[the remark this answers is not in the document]
apognu: I don't especially agree with the solution. If we could not claim the decision execution, why would we still execute the phantom decision if `r.skipped`?

dto/scenarios.go
jyrivallan: You’re right! I just did

repositories/migrations/20260817150000_scheduled_execution_deduplicate.sql
apognu, line 1: Can you merge this migration with the previous one? No point in having them separated that I can see.

dto/scenarios.go
apognu: The field is still present on `ScenarioDto`, `UpdateScenarioBody`, `DBScenario`, and in some comments, I think it's now unused?

apognu (changes requested):

repositories/migrations/20260817150000_scheduled_execution_deduplicate.sql
jyrivallan: Done

repositories/migrations/20260814090000_scenario_scored_objects.sql
```diff
@@ -0,0 +1,46 @@
...
+
+-- No org_id: reachable through scenario_id, which is globally unique. No table_name: a
+-- scenario has exactly one trigger_object_type.
+create unique index scenario_scored_objects_unique_idx
+    on scenario_scored_objects (scenario_id, object_id);
+
+-- +goose Down
+drop table scenario_scored_objects;
+alter table scheduled_executions drop column deduplicate_objects;
+alter table scenarios drop column deduplicate_batch_objects;
```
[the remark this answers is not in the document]
apognu: This is fine. If we roll back, we roll back.

apognu (approved):

PR #1931, opened 2026-08-17
Accept partner as a valid target for belongs to relationships.

apognu:
A generic person table is sent by the frontend as a `partner` semantic type, which was not allowed to become the target of a belongs to relationship, which is not expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Partner records are now correctly recognized as party types, alongside people and companies.


jyrivallan (approved):
LGTM

PR #1935, opened 2026-08-18
Fix race condition in public API tests.

apognu:
On macOS, Docker running on a virtual machine runs in the possibility to have both clock skewed by a few seconds. Some public API tests relies on the fact that a decision that was just made is visible through the API.

When `now()` was used in the database, a decision could be created in the future, making is invisible to `created_at <= now()`.

Also, use a native ARM64 PostgreSQL image when on ARM64.

## Summary by CodeRabbit

- **Tests**
  - Updated test data to use a decision timestamp one minute in the past.
  - Improved test environment compatibility across ARM64 and x86_64 systems by selecting the appropriate PostGIS container image automatically.


jyrivallan (approved):
LGTM

PR #1936, opened 2026-08-18
Maintain graph adjacency table incrementally during ingestion.

apognu:
`_graph` was previously refreshed only by a periodic full rebuild (default 24h), so a record ingested right after a build was invisible to graph walks until the next cycle.

Now we upsert into the table inside the existing ingestion transaction, behind a subtransaction so a failure never fails ingestion itself. Retraction covers a field an ingested version leaves null/empty, which the upsert cannot express since the table has no valid_until.

Rework the periodic build's swap to rename the live table aside instead of dropping it, so records ingested while a build is running are reconciled from the previous generation after the swap rather than replayed under a lock that graph read queries would otherwise block on.

## Summary by CodeRabbit

- **New Features**
  - Graph exploration now updates incrementally as records are ingested.
  - Changes made during graph rebuilding are preserved and incorporated after the rebuild completes.
  - Updated records automatically add current relationships and remove outdated or missing relationships.

- **Bug Fixes**
  - Improved graph rebuild reconciliation prevents recently ingested data from being lost during table swaps.
  - Graph relationships now remain consistent when field values change, disappear, or become unavailable.


usecases/ingestion_usecase.go
```diff
@@ -1251,7 +1263,17 @@ func (usecase *IngestionUseCase) insertEnumValuesAndIngest(
...
 		ingestionResults, err = usecase.ingestionRepository.IngestObjects(ctx, tx, payloads, table)
-		return err
+		if err != nil {
+			return err
+		}
+
+		if infra.HasFeatureFlag(infra.GRAPH_EXPLORATION_FEATURE_FLAG, organizationId) {
+			// In the transaction so the adjacency table cannot disagree with the version flip it derives
+			// from, and so the existing retries carry it along. Best-effort: it does not return an error.
+			usecase.maintainGraphRows(ctx, tx, organizationId, table, ingestionResults)
```
Pascal-Delange, line 1273: Thinking out loud: if it's best effort, maybe this can happen out of band after the load-bearing part completes, in a goroutine. The daily job re-synchronizes the full data as a failsafe.

usecases/worker_jobs/graph_build_job.go
```diff
@@ -112,6 +112,16 @@ func (w GraphBuilder) Build(ctx context.Context, organizationId uuid.UUID) error
 		return err
 	}
 
+	// Read before the first populate, so that everything the incremental ingestion writer commits
+	// from here on is either seen by a populate below or replayed before the swap. Erring early is
```
Pascal-Delange, line 116: Herring 

Image

or earring 

Image
?

usecases/ingestion_graph.go
```diff
@@ -0,0 +1,154 @@
...
+// on every batch. A relation registered during the TTL is picked up seconds late; measured against
+// the daily rebuild that used to be the only path, that is not a regression.
+var graphIndexedFieldsCache = expirable.NewLRU[string, map[string][]models.Field](
+	50, nil, utils.GlobalCacheDuration())
+
+// maintainGraphRows brings the graph adjacency table in line with the records just ingested, so a
+// graph walk can reach them without waiting for the next periodic rebuild.
+//
+// It reports no error by design. The graph is derived data with a rebuild job behind it, so the worst
+// outcome of giving up here is the staleness that was the status quo — whereas propagating would
```
Pascal-Delange, line 32: What is the status quo though :D

Pascal-Delange (approved):
Ok, I must confess I didn't go super deep into all SQL sequences, but looks correct from a safe distance.

usecases/worker_jobs/graph_build_job.go
apognu: [`to err`](https://dictionary.cambridge.org/dictionary/english-french/err).

Pascal-Delange: Yes I got it, but the -ing form is really uncommon and awkward :D

PR #1960, opened 2026-09-03
Add configurable lifecycle settings to client data tables.

apognu:
For each table, cleanup can be enabled and configured to:

- Delete stale records after a chosen retention period.
- Delete active records after a chosen retention period.

A scheduled cleanup runs hourly for each organization and removes records that exceed the configured retention periods.

## Summary by CodeRabbit

- **New Features**
  - Added configurable table lifecycle settings for automatically removing stale and active data.
  - Added scheduled client data purging based on each table’s retention settings.
  - Added graph-building jobs to organization task queues with configurable intervals.
  - Added support for viewing and updating lifecycle settings through data model APIs.

- **Bug Fixes**
  - Lifecycle durations are now validated, rejecting negative values and periods shorter than one month.
  - Improved enum-value loading efficiency when retrieving data models.


repositories/client_data_purge_repository.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	tableName := sanitizedTableName(exec, table)
+
+	cte := WithCtes("objects", func(b squirrel.StatementBuilderType) squirrel.SelectBuilder {
+		return b.
+			Select("id").
+			From(tableName).
+			Where("valid_from < ? and valid_until = 'infinity'", before.UTC()).
+			OrderBy("valid_until").
```
Pascal-Delange, line 61:
```suggestion
			OrderBy("valid_from").
```

Pascal-Delange (changes requested):
Other than the comment in a file below: Without an index (or a pair of scoped indexes) to identify the rows, this will probably have a pretty high DB cpu impact.
Typically, we could have an index on valid_from (where valid_until=infinity) and an index on valid_from (or valid_until, should be kinda equivalent) where valid_until!=infinity.

I think I would suggest we first try it our in prod conditions by creating the relevant indexes manually, then if it's all working add the useful indexing jobs?

Pascal-Delange (approved):
Thanks!

PR #1961, opened 2026-09-03
Per-org auth token minting

jyrivallan:
Token minting per organization or platform

### Summary

- **Grant-backed token minting**: `POST /token` now mints tokens from the union of the
  legacy role (`users.role` / `api_keys.role`) and the active grants. The legacy role
  only enters a token of its own scope (org-scoped role → org token, `MARBLE_ADMIN` →
  platform token), so `TENANT_ADMIN` grants backfilled in 1a appear alongside legacy
  roles. Authority switches to grants-only together with the backfill (separate change).
- **Organization selection**: `POST /token` accepts an optional `organization_id`;
  auto-selects the org when there is exactly one; otherwise issues an org-less
  platform token. New `GET /me/organizations` returns `id`, `name`, `roles` of the
  authenticated user's organizations.
- **JWT payload**: emits `tenant_id` and `roles[]`, keeps reading the legacy `role`
  field during rollout (dual-read window = configured token lifetime). Permissions
  are recomputed per request from the role-permission map, never carried in the token.
- **IP allowlist policy**: org-less session → union of the allowlists of the user's
  organizations (an org without allowlist makes the union unrestricted); org token →
  that org's allowlist only. Enforcement happens before minting, no response buffering.
- **Organization deletion**: `DeleteUsersOfOrganization` now removes the org's grants
  instead of deleting `users` rows — a multi-org user is never deleted across orgs.
- **API keys**: created keys get an org-scoped grant; auth resolves roles from grants
  union with the legacy `api_keys.role`.

### Notes

- Users and API keys created after this change get their grant written at creation
  time, so the future grants-only switch needs no extra backfill for new records.
- No schema change in this PR (relies on 1a's `tenants`/`grants` migration).

### Tests

- Unit: minting outcomes (legacy+grants union, `TENANT_ADMIN` inclusion, platform
  token for `MARBLE_ADMIN`), org selection, API-key grant union, JWT dual-read fields.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated endpoint to list the organizations a user belongs to, including organization names and assigned roles.
  * Added support for selecting an organization when requesting a token.
  * Credentials can now include multiple roles and associated tenant or organization context.
  * API keys and users now receive permissions based on their applicable grants.
  * Organization network access now supports user-based memberships and unrestricted access.


jyrivallan:
@coderabbit review

repositories/organization_repository.go
```diff
@@ -314,6 +314,37 @@ func (repo *MarbleDbRepository) GetOrganizationAllowedNetworks(ctx context.Conte
 	return SqlToModel(ctx, exec, sql, dbmodels.AdaptOrganizationWhitelistedSubnets)
 }
 
+func (repo *MarbleDbRepository) GetUserOrganizationAllowedNetworks(ctx context.Context, exec Executor, userID string) ([]net.IPNet, bool, error) {
+	if err := validateMarbleDbExecutor(exec); err != nil {
+		return nil, false, err
+	}
+	rows, err := exec.Query(ctx, `
+		SELECT o.allowed_networks
+		FROM active_grants g
```
Pascal-Delange, line 323: Thinking out loud - this looks like a recurring pattern, maybe a view could be useful

usecases/auth/generator.go
```diff
@@ -55,26 +57,104 @@ func NewGenerator(repository marbleRepository, encoder encoder, lifetime time.Du
...
+	switch creds.Type {
+	case CredentialsBearer:
+		if credentials.OrganizationId != uuid.Nil {
 			tracking.Identify(ctx, credentials.ActorIdentity.UserId, map[string]any{
 				"email": credentials.ActorIdentity.Email,
 			})
 			tracking.Group(ctx, credentials.ActorIdentity.UserId,
 				credentials.OrganizationId, map[string]any{
-					"name": organization.Name,
+					"name": credentials.OrganizationId.String(),
```
Pascal-Delange: can we no longer have name here ?

usecases/auth/generator.go
```diff
@@ -55,26 +57,104 @@ func NewGenerator(repository marbleRepository, encoder encoder, lifetime time.Du
...
+		}
+		for _, grant := range grants {
+			if grant.TenantId == uuid.Nil && grant.OrganizationId == uuid.Nil {
+				addRole(grant.Role)
 			}
-			credentials.TenantId = organization.TenantId
+		}
+	} else {
+		if legacyRole != models.NO_ROLE && legacyRole != models.MARBLE_ADMIN &&
+			legacyOrganizationId == credentials.OrganizationId {
```
Pascal-Delange: Not sure if I'm missing something, but you're setting legacyOrganizationId to be credentials.OrganizationId on line 67, and it's not mutated later, so this is trivially true?

usecases/auth/generator.go
```diff
@@ -55,26 +57,104 @@ func NewGenerator(repository marbleRepository, encoder encoder, lifetime time.Du
...
+			addRole(legacyRole)
+		}
+		for _, grant := range grants {
+			if grant.OrganizationId == credentials.OrganizationId || grant.TenantId == credentials.TenantId {
+				addRole(grant.Role)
+			}
+		}
+	}
+	credentials.Roles = roles
+	credentials.Role = legacyRole
```
Pascal-Delange: Ditto here: this should be no-op always?

usecases/auth/generator.go
```diff
@@ -55,26 +57,104 @@ func NewGenerator(repository marbleRepository, encoder encoder, lifetime time.Du
...
+		if legacyOrganizationId != uuid.Nil {
+			organizationIDs[legacyOrganizationId] = struct{}{}
+		}
+		for _, grant := range grants {
+			if grant.OrganizationId != uuid.Nil {
+				organizationIDs[grant.OrganizationId] = struct{}{}
+			}
+		}
+		if len(organizationIDs) == 1 {
+			for organizationID = range organizationIDs {
```
Pascal-Delange: empty loop?

usecases/auth/generator.go
Pascal-Delange: Ok I get it, it's credentials.OrganizationId that is mutated, sometimes.

usecases/auth/generator.go
Pascal-Delange: OOOOooooh wait, don't tell me it builds a set (well, a map of struct{}) to then iterate on its keys  (nb: `if len(...) == 1` seems redundant?) to update `organizationID`? This is hard to follow

Pascal-Delange (changes requested):
Ok, sorry I'm not going to lie, there is dark magic going on in `generator.go`.
For what I know the logic may be correct, but the coexistence of `organizationID`, `legacyOrganizationId` and `credentials.OrganizationId`, containing something similar in nature, where two input params (`credentials.OrganizationId` and `organizationID`) are mutated throughout the function, are a serious code smell. I looked at it for 10min and didn't understand the actual flow, even if I could guess the general idea.
Please rework that part 🙏

usecases/auth/generator.go
jyrivallan: We could. I thought it would be easier to have something technical here. But you’re right, it’s misleading with the `name` key. I reverted this change

usecases/auth/generator.go
jyrivallan: I simplified this part. It should be better now

usecases/auth/generator.go
jyrivallan: I don’t think it was. But I simplified it anyway

Pascal-Delange (approved):
ok, go (cf comment in another PR about squirrel - but we can add that to the stack rather than do it here)

PR #1964, opened 2026-09-04
feat: implement numeric switch functionality with case evaluation

william-schlegel:
## What was added

Rules can now express:

```go
Switch(
  Case(predicate, number),
  Case(predicate, number),
  ...,
  fallback: number
)
```

Evaluation is first-match: the first true case returns its number; if none match, the required fallback is returned. The result is always a `float64`, so it can sit in comparisons like Greater(payload.amount, Switch(...)).

## How it works

New Case AST node. `FUNC_CASE` takes exactly two children: a boolean-or-nil predicate, and a numeric value. False or nil means “unmatched” (so missing data falls through to fallback). Non-boolean predicates and non-numeric values are rejected with argument-indexed errors.

Switch now has two families. It looks at the first child:

- `ScoreComputationResult` → existing scoring switch (optional fallback, default metadata)
- `SwitchCaseResult` → new numeric switch (required numeric fallback)
- 
Mixing the two is rejected during full validation.

Short-circuiting. The old scoring-only circuit breaker now also stops after the first matching numeric case, so later cases are not evaluated at runtime. Validation still evaluates all children so malformed later branches are reported.

## What did not change

Scoring switches still validate and evaluate the same way, including Branch, Fallback, and Default metadata. Persistence is still the generic `AST DTO` (name, children, namedChildren), so there is no migration or OpenAPI change. Case is picked up automatically by the existing function-name map.

## Summary by CodeRabbit

* **New Features**
  * Added numeric switch evaluation with ordered case matching and fallback values.
  * Added conditional case branches supporting boolean conditions and numeric results.
  * Numeric switches can be used in formulas and reference payload fields.
  * Added validation for missing or invalid case and fallback values.
  * Existing scoring switch behavior now reports selected branch and fallback metadata more consistently.

* **Tests**
  * Added coverage for numeric switch evaluation, validation, type conversion, and scoring branch selection.

### DIFF
--- a/dto/dto_ast_node_test.go
+++ b/dto/dto_ast_node_test.go
@@ -57,3 +57,34 @@ func TestAstToJson(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEmpty(t, serialized)
 }
+
+func TestNumericSwitchDtoRoundTrip(t *testing.T) {
+	input := NodeDto{
+		Name: "Switch",
+		Children: []NodeDto{
+			{
+				Name: "Case",
+				Children: []NodeDto{
+					{Constant: true},
+					{Constant: 42.0},
+				},
+			},
+		},
+		NamedChildren: map[string]NodeDto{
+			"fallback": {Constant: 10.0},
+		},
+	}
+
+	node, err := AdaptASTNode(input)
+	assert.NoError(t, err)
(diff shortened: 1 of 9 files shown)


apognu (approved):
LGTM.

Tested circuit breaking and cache usage locally, all seems good.

PR #1965, opened 2026-09-07
Fix timestamp not being accepted in screening birth date field.

apognu:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved string concatenation handling for string, integer, and decimal values, including optional separators.
  - Empty values are skipped during concatenation.
  - Time values are returned immediately when encountered.
  - Unsupported value types now produce a clear error instead of being formatted unexpectedly.

### DIFF
--- a/usecases/ast_eval/evaluate/eval_string_concat.go
+++ b/usecases/ast_eval/evaluate/eval_string_concat.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/checkmarble/marble-backend/models/ast"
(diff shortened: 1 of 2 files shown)


Pascal-Delange (approved):

apognu:
Just as a note, something else feels "weird" in the way this is set up.

All field use string concatenation, which is okay for a name, but feels off for countries or dates. The UI makes it believable adding "fr" and "us" as countries will search both, whereas it will send "fr us", which will be silently ignored by motiva.

Even in this PR, we stop at the first timestamp, which means subsequent timestamp are ignored, which may not be what we ultimately want.

cc @Pascal-Delange

PR #1982, opened 2026-09-11
manage numeric switch with specific validation rules

william-schlegel:
- thresholds in ascending order
- same field for comparison
- <= opertor

## Summary by CodeRabbit

- **New Features**
  - Added validation for numeric-first-dimension switch formulas.
  - Reports a specific error for invalid switch configurations, including decreasing thresholds, inconsistent case structures, mismatched fields, and non-constant thresholds.
  - Applies validation across scenario formulas, rules, screening queries, counterparty expressions, and related scenario expressions.

- **Bug Fixes**
  - Correctly recognizes valid matrix cases with supported condition arrangements.
  - Screening query fields now accept non-empty lists of strings or timestamps as valid results.


usecases/scenarios/validate_numeric_switch.go
```diff
@@ -0,0 +1,224 @@
+package scenarios
+
+import (
+	"github.com/checkmarble/marble-backend/models"
+	"github.com/checkmarble/marble-backend/models/ast"
+	"github.com/checkmarble/marble-backend/usecases/ast_eval/evaluate"
+	"github.com/cockroachdb/errors"
+)
+
+// Structural validation for numeric-first-dimension Switches stored in rule formulas.
```
Pascal-Delange, line 10: I don't think I have anything to say about the implementation.
If I may nitpick a bit, however: the doc at the top is HARD to understand (and typical of AI generated text). 
IMO this would benefit of spending another 20min just giving examples of what AST structure this expects and what values it is supposed to generate and a high level overview of how it validates it. The whole file is hard to get if you don't already have the full mental model loaded.

Pascal-Delange: (Now, to be fair the test file helps)

usecases/scenarios/validate_numeric_switch_test.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+			if tt.wantMessage != "" {
+				assert.ErrorContains(t, errs[0].Error, tt.wantMessage)
+			}
+		})
+	}
+}
+
+func TestValidateNumericSwitchFromStoredJson(t *testing.T) {
+	const formula = `{
```
Pascal-Delange, line 157: Thinking out loud about the structure we use for this AST function.
I believe this works like this, and it's not necessary to change it now, but it does produce very verbose AST json because of how the input value is repeated in every cell.
Thinking forward a bit, if we at some point start to do this with computed aggregates as input, it would get even more verbose (because aggregates are verbose by themselves). I can't help but think that a representation where the input values are present once each, and the conditions are stored with just their boundaries, could also make sense. This is not critical, but it will have implications in particular when LLM rule generation or AST piece interpretation (e.g. by the case review agent) are concerned. 
Not a blocker but it would be interesting to discuss this later, cc @apognu

Pascal-Delange (approved):
Approved, with reservations on the doc (if it's there, worth to take 10min to improve it) and a more medium term subject to discuss later

usecases/scenarios/validate_numeric_switch.go
william-schlegel: ok, I changed the comment section to something more readable

PR #1983, opened 2026-09-14
Pin audited queries to their database connection

apognu:
Acquire dedicated connections for audited `Exec`, `Query`, and `QueryRow` calls so session audit settings apply to the statement they belong to.

Return release callbacks for row-based queries, update callers to release connections safely, and avoid retaining connections across subsequent or long-running work.

---

Add a `QueryGroup` helper to easily execute multiple queries in a pinned connection.

---

Add a new unaudited executor, to bypass prequeries used for auditing.


apognu:
@coderabbitai review

PR #1993, opened 2026-09-16
Add a SYSTEM role for internal workers

jyrivallan:
## `1d`: System principal for workers

- Added an internal SYSTEM role used exclusively by River workers.
- Migrated production worker wiring and the integration-test River harness from synthetic MARBLE_ADMIN credentials to SYSTEM.
- Kept SYSTEM internal: it cannot be selected as a user role or parsed from API input.
- Granted only worker-required permissions:
  - cross-organization access
  - decisions and phantom decisions
  - ingestion
  - workflow webhook events
  - case and inbox access
- Documented the temporary cross-organization bypass: derive organization-scoped credentials from each job payload in a future iteration.
- Replaced a fixed delay in the end-to-end workflow test with an assertion that waits for the actual case-creation outcome.

## Deferred to `1e`

- Remove ANY_ORGANIZATION_ID_IN_CONTEXT from MARBLE_ADMIN.
- Update integration-test fixtures that use synthetic Marble-admin credentials without an organization.
- Scope these fixture credentials to the organization they create.
- Verify that staff/admin actions require explicit organization scope, while SYSTEM workers continue to process cross-organization jobs.

## Summary by CodeRabbit

- **New Features**
  - Added an internal system role for background processing with narrowly scoped permissions.
  - Worker activity now records whether execution originated from an automated job or a manual job, including relevant job details.
  - System-level operations can access organization-scoped resources when authorized.

- **Bug Fixes**
  - Improved credential handling for worker and administrative setup flows.

- **Tests**
  - Added coverage for system permissions, organization access, and execution-source tracking.


jyrivallan:
@coderabbit review

jobs/generate_usecases_with_creds.go
```diff
@@ -1,16 +1,14 @@
...
 import (
-	"context"
-
 	"github.com/checkmarble/marble-backend/models"
 	"github.com/checkmarble/marble-backend/usecases"
 	"github.com/google/uuid"
 )
 
-func GenerateUsecaseWithCredForMarbleAdmin(ctx context.Context, jobUsecases usecases.Usecases) usecases.UsecasesWithCreds {
+func GenerateUsecaseWithCredForSystem(jobUsecases usecases.Usecases) usecases.UsecasesWithCreds {
```
apognu, line 9: Please let's try and add something that stores the source of the generation, to be included in logs. The fact that an action was performed through a system token generated by job 123 has value for auditability.

apognu (changes requested):

jobs/generate_usecases_with_creds.go
jyrivallan: You can have a second look @apognu

apognu (approved):
Looks good to me.

Something to keep in mind, but probably for another pull request, is that with this, we might finally be able to get rid of a pattern we use across the codebase, where some methods have two versions: one checking the active credentials, and one that does not (for example to be executed by a worker).

Some examples of this are:

- [InternalComputeScore](https://github.com/checkmarble/marble-backend/blob/main/usecases/scoring/scoring_scores_usecase.go#L111)
- [EnrichMatchWithoutAuthorization](https://github.com/checkmarble/marble-backend/blob/main/usecases/screening_usecase.go#L976C28-L976C59)

models/role_permission.go
```diff
@@ -101,4 +101,16 @@ var ROLES_PERMISSIONS = map[Role][]Permission{
 		LICENSE_CREATE,
 		LICENSE_UPDATE,
 	),
+	SYSTEM: {
+		ANY_ORGANIZATION_ID_IN_CONTEXT,
```
Pascal-Delange, line 105: I don't know how much effort went into verifying this list, but this may be the tricky part of this PR.
Today the "system" jobs effectively work with the largest permission scope, let's keep a strong eye out for errors related to this on the worker

PR #1995, opened 2026-09-16
CI: Handle gracefully flaky tests

jyrivallan:


apognu:
Let's actually debug and see why the tests are flaky, and why the Postgres instance dies instead of looping, making an actual failing test come back in twice the time.

jyrivallan:
> Let's actually debug and see why the tests are flaky, and why the Postgres instance dies instead of looping, making an actual failing test come back in twice the time.

I didn’t mean to not do that. I just wanted to make flaky tests detectable while not randomly blocking unrelated PRs.
As a bonus, this also makes the test output much more readable.

